PR #3, opened 2023-01-30
SMA core + add SMA to prototype test

ianspektor:
Extend core to be able to handle:
- `Sampling`s as inputs, instead of only `Event`s . I did this by adding all samplings of input features and created features to the data available to each operator, but we should make a distinction between `Sampling` input nodes and `Event` input nodes. Left a bunch of TODOs in some key lines to do so. Can integrate it into this same PR if preferred.
- Static `Attribute`s for an operator set during graph definition (vs. `EventArgument`s which are passed during eval)

Unified logic of splitting index into index columns and timestamp column in a new utils file.

Also added usage of `sma` operator to `prototype_test`.


temporian/core/evaluator.py
```diff
@@ -68,14 +65,15 @@ def evaluate(
...
+  for input_event, input_event_spec in input_data.items():
+    if not isinstance(input_event_spec, event):
+      input_event_spec = read_csv_fn(input_event_spec, input_event.sampling())
+
+    materialized_input_data[input_event.sampling(
+    )] = input_event_spec.index  # TODO: handle samplings as different type of nodes instead of doing this
+    materialized_input_data[input_event] = input_event_spec
+
   # evaluate schedule
   outputs = evaluate_schedule_fn(materialized_input_data, schedule)
```
achoum, line 78: Nit: What about having comments on a separate line (e.g., above) so as to limit the line length?

temporian/core/operators/simple_moving_average.py
```diff
@@ -43,19 +43,15 @@ def __init__(
...
     self.add_input("data", data)
 
-    features = [  # pylint: disable=g-complex-comprehension
-        Feature(
-            name=f.name(),
-            dtype=f.dtype(),
-            sampling=sampling,
-        )
+    output_features = [  # pylint: disable=g-complex-comprehension
+        Feature(name=f.name(), dtype=f.dtype(), sampling=sampling, creator=self)
```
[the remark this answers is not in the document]
achoum: For later: During the e2e execution, with this current code, there should be failure since the SMA operator definition contains a prefix.

For later: What about having a method that automatically generates the prefix using the attribute values.

temporian/core/operators/simple_moving_average.py
```diff
@@ -66,6 +62,13 @@ def __init__(
   def build_op_definition(cls) -> pb.OperatorDef:
     return pb.OperatorDef(
         key="SIMPLE_MOVING_AVERAGE",
+        attributes=[
+            pb.OperatorDef.Attribute(
+                key="window_length",
+                type=pb.OperatorDef.Attribute.Type.STRING,
+                is_optional=False,
```
achoum, line 69: s/False/True ?

temporian/core/operators/base.py
```diff
@@ -61,6 +61,7 @@ class Operator(ABC):
   def __init__(self):
     self._inputs = {}
     self._outputs = {}
+    self._attributes = {}
 
   def __str__(self):
     return f'Operator<key:{self.definition().key},id:{id(self)}>'
```
achoum, line 67: What about adding the attributes in the __str__?

temporian/implementation/pandas/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""Utility functions for the Pandas implementation."""
+from typing import Union
+from temporian.implementation.pandas.data import event
+from temporian.implementation.pandas.data import sampling
+
+
+def get_index_and_timestamp_column_names(
+    obj: Union[event.PandasEvent, sampling.PandasSampling]
+) -> tuple[list[str], str]:
```
achoum, line 23: Does this act similarly as "Tuple" from the typing package? (same for list).

temporian/implementation/pandas/utils.py
```diff
@@ -0,0 +1,40 @@
...
+  """Get the names of the index columns and the timestamp column from a
+  PandasEvent or PandasSampling.
+
+  Args:
+      obj (Union[event.PandasEvent, sampling.PandasSampling]): the event or
```
achoum, line 28: No need for typing information in the documentation.

temporian/core/operators/base.py
```diff
@@ -61,6 +61,7 @@ class Operator(ABC):
   def __init__(self):
     self._inputs = {}
     self._outputs = {}
+    self._attributes = {}
```
achoum, line 64: Can the attributes be typed to the allowed types (str and int) ? (https://github.com/google/temporian/blob/main/temporian/proto/core.proto#L60)

temporian/core/operators/simple_moving_average.py
```diff
@@ -29,12 +29,12 @@ class SimpleMovingAverage(Operator):
...
       self,
       data: Event,
-      window_length: int,
+      window_length: str,
       sampling: Optional[Event] = None,
   ):
     super().__init__()
 
-    self.window_length = window_length
+    self.add_attribute("window_length", window_length)
```
achoum, line 37: For later: It would be good to parse the window_length before the graph execution as to catch early any issues.

achoum (approved):
Looks good :).
Let a few comments.

ianspektor:
Discussed on 31/1: let's change to only having inputs of type Event, and if an op only needs to use its sampling have it take it from whichever event it came from.

temporian/core/operators/simple_moving_average.py
ianspektor: If this means substituting False for True: isn't window length mandatory? Do we want it to have a default value?

temporian/implementation/pandas/utils.py
ianspektor: Yep! Python docs for [typing.List](https://docs.python.org/3.10/library/typing.html#typing.List) and [typing.Tuple](https://docs.python.org/3.10/library/typing.html#typing.Tuple) declare them as deprecated since 3.9, replaced by the built-in `list` and `tuple` directives :) Same goes for `dict`, `set` and `type`. See [PEP 585](https://peps.python.org/pep-0585/).

rstz (approved):

temporian/core/operators/base.py
ianspektor: Adding type hint `self._attributes: dict[str, Union[str, int]] = {}`, if you referred to something else lmk 🙌🏼

temporian/implementation/pandas/utils.py
achoum: Python 3.7 might be a stretch, but we should be compatible with python 3.8 (https://w3techs.com/technologies/details/pl-python/3). We need to check, but maybe in this case we need  a `from __future__ import annotations`.

temporian/core/operators/simple_moving_average.py
achoum: My bad, I was looking at sampling. This looks good.

temporian/core/operators/simple_moving_average.py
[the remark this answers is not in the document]
achoum: >could you expand on this?

I was referring to the line `Feature(name=f"sma_{f.name()}",` in the SML definition.

temporian/core/operators/base.py
achoum: Looks good.

PR #6, opened 2023-02-01
Improve graph schedule

getchebarne:
This PR improves the graph scheduling algorithm by:

Removing previous never ending loop bug
- Adding operator depth calculation (for parallelization purposes)
- I've extended the prototype_test.py with a more complex example in which there's 3 different depths (2, 1, & 0). The depth for each operator is:
```
{
    <temporian.core.operators.assign.AssignOperator object at 0x12b108820>: 2,      # assign_output_1
    <temporian.core.operators.assign.AssignOperator object at 0x12b1086d0>: 2,     # assign_output_2
    <temporian.core.operators.assign.AssignOperator object at 0x11fd87670>: 1,      # final_assign_output
    <temporian.core.operators.simple_moving_average.SimpleMovingAverage object at 0x12b13ff40>: 1,    # sma_assigned_event
    <temporian.core.operators.assign.AssignOperator object at 0x11e757520>: 0  # output_event
}
```
Operators with the same depth can be parallelized. **The depth calculation isn't finished though; it can be improved**. In the `prototype_test.py` example, the sma operator should have a depth of 2 (same as the first two assigns). This could be fixed by doing a second pass of the graph, from top to bottom. I'm still investigating the best way of implementing this.

All tests pass.

### DIFF
--- a/temporian/core/data/feature.py
+++ b/temporian/core/data/feature.py
@@ -14,7 +14,7 @@
 
 """A feature."""
 
-from typing import Any, Optional, Set
+from typing import Any, Optional
 
 from temporian.core.data import sampling as sampling_lib
 
@@ -49,13 +49,3 @@ class Feature(object):
 
     def set_sampling(self, sampling: sampling_lib.Sampling):
         self._sampling = sampling
-
-    def parent_features(self) -> Set["Feature"]:
-        if self.creator() is None:
-            return {}
-
-        return {
-            input_feature
-            for input_event in self.creator().inputs().values()
-            for input_feature in input_event.features()
-        }
--- a/temporian/core/data/feature.py
+++ b/temporian/core/data/feature.py
@@ -14,7 +14,7 @@
 
 """A feature."""
 
-from typing import Any, Optional
+from typing import Any, Optional, Set
 
 from temporian.core.data import sampling as sampling_lib
 
@@ -49,3 +49,13 @@ class Feature(object):
 
     def set_sampling(self, sampling: sampling_lib.Sampling):
         self._sampling = sampling
+
+    def parent_features(self) -> Set["Feature"]:
+        if self.creator() is None:
+            return {}
+
+        return {
+            input_feature
+            for input_event in self.creator().inputs().values()
+            for input_feature in input_event.features()
+        }
--- a/temporian/core/evaluator.py
+++ b/temporian/core/evaluator.py
@@ -19,7 +19,6 @@ from typing import Any, Dict, List, Set, Union
 
 from temporian.core import backends
 from temporian.core.data.event import Event
-from temporian.core.data.feature import Feature
 from temporian.core.operators import base
 from temporian.implementation.pandas.data import event as pandas_event
 
@@ -58,15 +57,8 @@ def evaluate(
     evaluate_schedule_fn = selected_backend["evaluate_schedule_fn"]
     read_csv_fn = selected_backend["read_csv_fn"]
 
-    # get features from events
-    features_to_compute = {
-        feature
-        for event in events_to_compute
-        for feature in event.features()  # pytype: disable=attribute-error
-    }
-    # calculate opeartor schedule. Only using keys (operators) for now, discarding
-    # depth
-    schedule = get_operator_schedule(features_to_compute).keys()
+    # calculate opeartor schedule
+    schedule = get_operator_schedule(events_to_compute)
 
     # materialize input data. TODO: separate this logic
     materialized_input_data = {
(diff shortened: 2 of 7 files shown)


ianspektor (approved):
LGTM! Left some minor non blocking ideas

temporian/core/data/feature.py
```diff
@@ -49,3 +49,13 @@ def creator(self):
 
     def set_sampling(self, sampling: sampling_lib.Sampling):
         self._sampling = sampling
+
+    def parent_features(self) -> Set["Feature"]:
```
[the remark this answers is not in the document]
achoum: +1

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
-        if not isinstance(input_event_spec, event):
-            input_event_spec = read_csv_fn(
-                input_event_spec, input_event.sampling()
-            )
-        materialized_input_data[input_event] = input_event_spec
-
+    materialized_input_data = {
+        input_event: (
+            input_event_spec
+            if isinstance(input_event_spec, event)
```
achoum, line 75: Is this "pandas_event.PandasEvent"?

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
+            else read_csv_fn(input_event_spec, input_event.sampling())
```
achoum, line 76: For later / add todo: Add a "isinstance" check in the csv case, and raise an exception if input_event_spec's type is not supported.

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
-
-    return operators_to_compute
+def get_operator_schedule(query: Set[Feature]) -> Dict[base.Operator, int]:
+    """Calculates which operators need to be executed in which order to
+    compute a set of query features.
+
+    Args:
+        query: set of query features to be computed.
+
+    Returns:
```
achoum, line 93: Output type is not needed here.

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
+        Dict[base.Operator, int]: dictionary mapping operators to their respective
+        depths in the compute graph. Depth is measured from bottom to top, i.e.
+        0 depth corresponds to the output features (query). Operators with the
+        same depth can be computed in parallel.
```
achoum, line 97: If there are multiple paths from the output to a given event, is the depth guarantied to be the length of the shortest path?

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
+    """
+    # start features. Depth initialized as 0
+    feature_depth = [(feature, 0) for feature in query]
+
+    # get all features and depths required to compute the query. One feature can
```
achoum, line 102: Super Nit; Here and in other place: It could help the reading if sentence comments (i.e. comments that ends with a ".") has a first letter upper-case.

temporian/core/evaluator.py
```diff
@@ -52,72 +52,76 @@ def evaluate(
...
+    # have multiple depths in this set (if it appears in more than one feature's
+    # compute path at different depths)
+    for feature, depth in feature_depth:
+        this_depth = depth + 1
+        for parent_feature in feature.parent_features():
+            feature_depth.append((parent_feature, this_depth))
```
achoum, line 108: Nit: For the readability sake, I would avoid to add items to the array "feature_depth" while iterating over it.

If I read this code correctly, in the worst case scenario (with a strongly connected graph), you can end up with 2^p elements in "parent_feature" where p is the number of operators. This could very quickly becoming an issue.

Formally, this problem is called topological sorting (currently, we don't care/handleabout parallel execution). I would look to the Kahn's algorithm, or possibly the Depth-first search one (https://en.wikipedia.org/wiki/Topological_sorting).

achoum (approved):
As discussed during meeting, let' go with the proposed implementation of the scheduler, and work on improvements in a subsequent work.

The other comments would be great to have a followup.

temporian/core/evaluator.py
getchebarne: it is if the user selects "pandas" as the backend

getchebarne: `input_event_spec` can also be a string or path pointing to the data's path in disk

ianspektor (approved):
LGTM! Thanks for the clear comments on the step by step of the DFS impl in evaluate :)

achoum (approved):

PR #8, opened 2023-02-01
Select Operator

getchebarne:
This PR implements the Select Operator. 

The operator can be called:

- using ```temporian.core.operators.select.select``` (as with any other operator)
- using the `__getitem__` method in the core `Event` class, i.e. ```selected_event = event[["f_1", "f_2"]]```

The implementation is hacky, but it's the only way I found to be able to create a view of multiple columns of a `pandas` `DataFrame`. Using `df[["f_1", "f_2"]]` always returns a copy, which we don't want. The current implementation does not duplicate any data.

I've also updated the `prototype_test.py` script to include an instance of the select operator. 

Other unrelated, small changes:
- fixed `PandasAssignOperator` return type
- removed unnecessary print in core `Operator` class
- updated core `Event` __repr__ method to be more readable

All tests pass.

### DIFF
--- a/temporian/core/data/event.py
+++ b/temporian/core/data/event.py
@@ -16,7 +16,6 @@
 
 from typing import Any, List, Optional
 
-import temporian
 from temporian.core.data.feature import Feature
 from temporian.core.data.sampling import Sampling
 
@@ -35,11 +34,13 @@ class Event(object):
         self._creator = creator
 
     def __getitem__(self, feature_names: List[str]) -> "Event":
+        # import select operator
+        from temporian.core.operators.select import SelectOperator
+
         # instance select operator. Import from base temporian package
         # to avoid circular import
-        select_operator = temporian.core.operators.select.SelectOperator(
-            self, feature_names
-        )
+        select_operator = SelectOperator(self, feature_names)
+
         # return Event
         return select_operator.outputs()["event"]
 
--- a/temporian/core/data/event.py
+++ b/temporian/core/data/event.py
@@ -41,7 +41,7 @@ class Event(object):
             self, feature_names
         )
         # return Event
-        return select_operator.outputs()["output_event"]
+        return select_operator.outputs()["event"]
 
     def __repr__(self) -> str:
         features_print = "\n\t\t".join(
--- a/temporian/core/data/event.py
+++ b/temporian/core/data/event.py
@@ -16,6 +16,7 @@
 
 from typing import Any, List, Optional
 
+import temporian
 from temporian.core.data.feature import Feature
 from temporian.core.data.sampling import Sampling
 
@@ -32,6 +33,15 @@ class Event(object):
         self._features = features
         self._sampling = sampling
         self._creator = creator
+
+    def __getitem__(self, feature_names: List[str]) -> "Event":
+        # instance select operator. Import from base temporian package
+        # to avoid circular import
+        select_operator = temporian.core.operators.select.SelectOperator(
+            self, feature_names
+        )
+        # return Event
+        return select_operator.outputs()["output_event"]
 
     def __repr__(self) -> str:
         features_print = "\n\t\t".join(
--- a/temporian/core/data/event.py
+++ b/temporian/core/data/event.py
@@ -33,8 +33,18 @@ class Event(object):
         self._sampling = sampling
         self._creator = creator
 
-    def __repr__(self):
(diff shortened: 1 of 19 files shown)


temporian/core/data/event.py
```diff
@@ -16,6 +16,7 @@
 
 from typing import List
 
+import temporian
```
achoum, line 19: It would be best for internal code not to use the top level user api import

temporian/core/operators/select.py
```diff
@@ -0,0 +1,84 @@
...
+from temporian.core.data.feature import Feature
+from temporian.core.operators.base import Operator
+from temporian.proto import core_pb2 as pb
+
+
+class SelectOperator(Operator):
+    """Select operator."""
+
+    def __init__(
+        self, input_event: Event, feature_names: Union[str, List[str]]
```
achoum, line 29: s/input_event/event 
I think we agreed on this a few weeks ago.

temporian/core/operators/select.py
```diff
@@ -0,0 +1,84 @@
...
+    ):
+        super().__init__()
+
+        # inputs
+        self.add_input("input_event", input_event)
+
+        # outputs
+        output_features = [
```
achoum, line 37: This is effectively changing the orders of the features.
What about having the output features in the same order as "feature_names"?

In addition, it would be interesting for this function to fails is "feature_names" does not exist in "input_event".

temporian/core/operators/select.py
```diff
@@ -0,0 +1,84 @@
...
+            Feature(
```
achoum, line 38: No need to re-create the Feature object.

temporian/core/operators/select.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+            for feature in input_event.features()
+            if feature.name() in feature_names
+        ]
+        output_sampling = input_event.sampling()
+        self.add_output(
+            "output_event",
+            Event(features=output_features, sampling=output_sampling),
+        )
+        # feature names
```
achoum, line 52: Move before the "output_features  = [" loop?

temporian/core/operators/select.py
```diff
@@ -0,0 +1,84 @@
...
+        if isinstance(feature_names, str):
+            feature_names = [feature_names]
+        self.add_attribute("feature_names", feature_names)
```
achoum, line 55: Nit: It might not be needed to add this as an attribute.
I'll explain why at the next meeting.

temporian/implementation/pandas/operators/select.py
```diff
@@ -0,0 +1,52 @@
...
+        # get index
+        event_index = input_event.index
+
+        # get column positions
+        column_positions = {
+            col: pos for pos, col in enumerate(input_event.columns)
+        }
+        # get numpy array from DataFrame. This creates a view (doesn't
+        # duplicate memory)
+        input_event_arr = input_event.to_numpy()
```
achoum, line 38: I think this will not work if the input_event features have different dtypes.

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_numpy.html

temporian/test/prototype_test.py
```diff
@@ -59,15 +59,17 @@ def test_prototoype(self) -> None:
         )
 
         # call assign operator
-        output_event = assign(assignee_event, assigned_event)
+        assign_output_event = assign(assignee_event, assigned_event)
 
         # call sma operator
-        sma_assigned_event = sma(
-            assigned_event, window_length="7d", sampling=assigned_event
+        sma_assigned_selected_event = sma(
```
achoum, line 65: That is awesome.

temporian/core/data/event.py
getchebarne: Haven't found a way to circumvent this yet.

achoum: I see two solutions:

- [Simple] The select op file included in the same build rule as the abstract operator.
- [Scallable] Create a registration mechanism (like this one https://github.com/google/temporian/blob/main/temporian/core/operator_lib.py) for ops to register to special callings on op (like [] and [[]]) in this case.

ianspektor (approved):
LGTM 💯

ianspektor (approved):

achoum (approved):

temporian/core/data/event.py
achoum: Let's keep this for a next pr.

PR #9, opened 2023-02-02
Serialization / unserialization of computation graph to / from file.

achoum:
This PR introduce the methods `t.save(...)` and `t.load` to save and load a computation graph to / from file.


temporian/core/processor.py
```diff
@@ -139,15 +175,24 @@ def infer_processor(inputs: List[Event], outputs: List[Event]) -> Preprocessor:
 
         else:
             p.operators().add(feature.creator())
-            # Add the parent features
+            p.samplings().add(feature.sampling())
```
ianspektor, line 178: Nit: would change these to `add_operator` and `add_sampling` methods in `Processor` for consistency with methods in operator class

ianspektor: same with `add_event` below

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Serialization / unserialization of a processor."""
+
+from google.protobuf import text_format
```
ianspektor, line 17: Is this a google internal tool? Doesn't seem to be available in our environment 👀

temporian/test/api_test.py
```diff
@@ -55,9 +48,31 @@ def disabled_test_create_processor(self):
                 a: input_signal_data,
             },
         )
-
         logging.info("results: %s", results)
 
+    def test_serialization(self):
+        a = t.place_holder(
```
ianspektor, line 54: Is the placeholder op still necessary? In [this test](https://github.com/google/temporian/blob/main/temporian/test/prototype_test.py#L52) we create an `Event` from scratch that is then passed to some operators. Is making this an operator necessary? I recall also talking about the pro of allowing the user to input any intermediate event as `input` when calling `evaluate`if the input events are not "special" in any way, rather just normal events without a `creator`.

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+            _identifier(src.creator) if src not in input_samplings else None
+        ),
+    )
+
+
+def _unserialize_sampling(src: pb.Sampling) -> Sampling:
+    return Sampling(index=src.index, creator=None)
+
+
+def _type_to_proto(dtype) -> pb.Feature.DType:
```
ianspektor, line 300: nit: naming of `_type_to_proto`, ..., `_attribute_from_proto` could follow the `_serialize`/`_unserialize` convention the other objects follow?

temporian/test/api_test.py
```diff
@@ -55,9 +48,31 @@ def disabled_test_create_processor(self):
...
+                t.Feature(name="f2", dtype=t.dtype.FLOAT),
+            ],
+            index=[],
+        )
+        b = t.sma(data=a, window_length=7)
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            path = os.path.join(tempdir, "my_processor.tem")
+            t.save(
+                inputs={"a": a},
```
ianspektor, line 66: Gave some thought to what we discussed today about needing to specify inputs 💭 

We think this could be optional if the user wants to specify some intermediate events as a cutoff for creating the graph, but if not default we should default to creating the graph from the `outputs` all the way up to events that don't have a `creator` (raw input events). Same goes for the `inputs` param in the `infer_processor` method - by default we should create and save the whole graph, crawling up from the `output` ones until we reach events with no creator (similar as the evaluator's logic)

See related comment about the place holder op not being strictly necessary if we treat events without a creator as input events implicitly.

Let's dedicate some time to this in tuesday's meeting if necessary!

ianspektor (commented):
👏🏼👏🏼👏🏼👏🏼👏🏼

Main comments are about allowing the user not to specify the `inputs` to `save`, serialization logic looks solid! Great work :)

temporian/core/processor.py
achoum: Good point.

*Added and used add_operator, add_sampling, add_feature and add_event.

temporian/core/serialize.py
achoum: The name is weird, but this is the classical protobuf python package (https://pypi.org/project/protobuf/#description).

temporian/test/api_test.py
achoum: You right, the placeholder op is not necessary. We can replace it with a helper function with an equivalent signature or let the users create the events manually. Let's do some test / chat if an helper function make sense.

*Added todo in place_holder.py.

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+
+    return p.inputs(), p.outputs()
+
+
+def serialize(src: processor.Preprocessor) -> pb.Processor:
+    """Serializes a processor into an equivalent protobuffer."""
+
+    return pb.Processor(
+        operators=[_serialize_operator(o) for o in src.operators()],
+        events=[_serialize_event(o) for o in src.events()],
```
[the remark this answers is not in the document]
achoum: Good idea.

*Changed accordingly.

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+            _serialize_io_signature(k, e) for k, e in src.outputs().items()
+        ],
+    )
+
+
+def unserialize(src: pb.Processor) -> processor.Preprocessor:
+    """Unserializes a protobuffer into a processor."""
+
+    # Decode the components.
+    # All the fieds except for the "creator" ones are set.
```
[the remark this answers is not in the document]
achoum: Done
*Applied

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+    samplings = {s.id: _unserialize_sampling(s) for s in src.samplings}
```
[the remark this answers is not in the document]
achoum: *Done

temporian/core/serialize.py
```diff
@@ -0,0 +1,340 @@
...
+        ],
+    )
+
+
+def _unserialize_operator(
+    src: pb.Operator, events: Dict[str, Event]
+) -> base.Operator:
+    operator_class = operator_lib.get_operator_class(src.operator_def_key)
+
+    def gen_event(key):
```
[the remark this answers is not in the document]
achoum: Good catch.

*Fixed

temporian/core/serialize.py
achoum: *Done

temporian/test/api_test.py
achoum: I like the idea of optional simplification. Let's discuss tuesday about the logic we want.

Question: Imagine a situation of a graph with two input events. How is the user specifying the input of each specific event? 

Bonus: I am wondering if the use of the processor directly would be a good idea. Here is how it would look like.

p = t.infer_processor(....)
result = p(event_1=..., event_2=...) # Equivalent to t.evaluate(inputs={p.inputs()["event_1"]:..., p.inputs()["event_2"]:...}, outputs=p.outputs()
p.save(path=...)
p2 = t.load(path=...)

temporian/core/serialize.py
ianspektor: Import was failing but I reinstalled the env and it worked. My bad!

temporian/test/api_test.py
ianspektor: > Imagine a situation of a graph with two input events. How is the user specifying the input of each specific event?

Currently in `evaluate` the user specifies the input events [using an input `Event`'s ref](https://github.com/google/temporian/blob/main/temporian/test/prototype_test.py#L77), but we might need to change that to using the event's unique name or id since we won't have that ref after serializing and deserializing.

I think this is independent from the serialization and deserialization logic though, which should just walk back from the specified output event until source events (i.e., events with no incoming edges in the graph, i.e. events without a creator, i.e. "input events") unless the user specifies otherwise with your proposed syntax.

achoum:
The thing is that currently, events are not identified by name (hence the
need for an input key in the serialization api). What about making the
input argument / key optionnal when there is only one input?

Le ven. 3 févr. 2023 à 15:30, Ian Spektor ***@***.***> a
écrit :

ianspektor:
As discussed today: decision is to extend the API with both alternatives.

A. The user can override event names during save
```
t.save(
    inputs={"a": a},
    outputs={"b": b},
)
```

B. The user can just specify an output event and have the graph go up to input (leaf) events automatically with the names specified when instantiating each event
```
t.save(
    outputs=[b]
)
```

Option A will be merged now, @achoum will implement option B in a follow up PR.

ianspektor (approved):

achoum:
Thanks

PR #10, opened 2023-02-03
Sum Operator

diegomarvid:
>__Warning__: This PR is rebased from [Numpy Backend PR](https://github.com/google/temporian/pull/11) not main

In this PR I create a simple Sum Operator with Pandas and NumPy as backends. Only the sum per feature index is supported.

The sum operator takes two events as inputs and outputs an event.  The output is an event with the same shape as the inputs. Each new feature is the sum of the two input features. 

If the argument `resolution` of the sum is `PER_FEATURE_IDX`, the sum will be feature index-wise. On the other hand, if the argument is `PER_FEATURE_NAME`, the sum will be feature name-wise.

Both events must have the same sampling, index, and number of features. In the case of `PER_FEATURE_NAME`, both event features must be equal.

- [x] Created Sum Operator Core
- [x] Created Sum Operator Pandas Implementation
- [x] Created Sum Operator Numpy Implementation
- [x] Updated Tests
- [x] Created NumpySampling Equals


temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,67 @@
...
+            raise ValueError(
+                f"Resolution must be one of {RESOLUTIONS}, got {resolution}"
+            )
+
+        # raise value error if data_1 and data_2 have different shape
+        if data_1.shape != data_2.shape:
+            raise ValueError("data_1 and data_2 must have same shape.")
+
+        # sum each feautre index wise
+        if resolution is None or resolution == "PER_FEATURE_IDX":
```
getchebarne, line 54: How about converting the DataFrames to numpy and summing those instead? That way we can avoid a for loop on the columns. NumPy will always sum columns by position.

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,67 @@
...
+            output = data_1.copy()
+            for i, column in enumerate(data_1.columns):
+                output.iloc[:, i] = data_1.iloc[:, i] + data_2.iloc[:, i]
+
+        if resolution == "PER_FEATURE_NAME":
+            raise NotImplementedError(
+                "PER_FEATURE_NAME resolution not implemented yet."
+            )
+
+        output_feature_names = "sum_" + data_1.columns + "_" + data_2.columns
```
getchebarne, line 64: Can we use f-strings instead of additions for creating format strings?

temporian/implementation/pandas/operators/tests/sum/test_data/different_shape.py
```diff
@@ -0,0 +1,51 @@
...
+        "timestamp": [
+            pd.Timestamp("2020-11-09"),
+            pd.Timestamp("2020-11-10"),
+        ],
+        "B": [
+            3,
+            4,
+        ],
+    }
+).set_index(["timestamp"])
```
getchebarne, line 38: I think this formatting can be improved - we don't need to split lists into multiple lines if they only have 2 elements.

getchebarne (changes requested):

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,97 @@
...
+from temporian.proto import core_pb2 as pb
+
+
+class SumOperator(Operator):
+    """Sum operator."""
+
+    def __init__(
+        self,
+        data_1: Event,
+        data_2: Event,
```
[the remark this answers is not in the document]
diegomarvid: Yes. I used those names because they were the ones chosen in the notes for the operator implementation. I agree that `event_1` and `event_2` are better names!

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__()
+
+        # inputs
+        self.add_input("data_1_event", data_1)
+        self.add_input("data_2_event", data_2)
+
+        if resolution is not None:
+            self.add_attribute("resolution", resolution)
+
+        sampling = data_1.sampling()
```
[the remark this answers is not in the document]
diegomarvid: Agree, should this be implemented on the core implementation?

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # outputs
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"sum_{data_1.features()[f].name()}_{data_2.features()[f].name()}",
+                dtype=f.dtype(),
+                sampling=sampling,
+                creator=self,
+            )
+            for f in range(
```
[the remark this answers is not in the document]
diegomarvid: I did this in the pandas implementation by ensuring that both DataFrames have the same shape. A better option is to check the number of features in the core, as you said, right?

temporian/implementation/pandas/operators/sum.py
diegomarvid: Yes, that was my first solution. However, I changed to this one because I don't know how reliable is changing to a numpy matrix if you have different types. Could the numpy implementation introduce errors in the feature data types?

temporian/implementation/pandas/operators/core_mapping.py
```diff
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from temporian.implementation.pandas.operators import assign
+from temporian.implementation.pandas.operators import assign, sum
```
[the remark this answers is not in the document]
ianspektor: Yep! You can disable it for a specific workspace from the VSCode extension's page

image

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,97 @@
...
+        if resolution is not None:
+            self.add_attribute("resolution", resolution)
+
+        sampling = event_1.sampling()
+
+        # outputs
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"sum_{event_1.features()[f].name()}_{event_2.features()[f].name()}",
+                dtype=f.dtype(),
```
ianspektor: Isn't `f` an `int` here coming from the `range` you're looping on?

temporian/core/operators/sum.py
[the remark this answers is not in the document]
ianspektor: Yep. I'd rather have sampling + features length checked in core than in each impl.

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,67 @@
...
+from temporian.implementation.pandas.operators.base import PandasOperator
+
+RESOLUTIONS = ["PER_FEATURE_IDX", "PER_FEATURE_NAME"]
+
+
+class PandasSumOperator(PandasOperator):
+    def __call__(
+        self, event_1: PandasEvent, event_2: PandasEvent, resolution: str = None
+    ) -> PandasEvent:
+        """Sum two EventSequences.
```
ianspektor, line 27: `EventSequence` is not a term we're using anymore, would change for just `"events"`. Same for the same word appearing below.

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Args:
+            event_1: First EventSequence.
+            event_2: Second EventSequence.
+            resolution: PER_FEATURE_IDX -> Sum is done to each feature index wise.
+                        PER_FEATURE_NAME (Not implemented yet) -> Sum is done to each feature name wise.
```
ianspektor, line 33: Comments and docstrings shouldn't go over the character limit for each line if possible

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Implementation for the sum operator."""
+
+from temporian.implementation.pandas.data.event import PandasEvent
+from temporian.implementation.pandas.operators.base import PandasOperator
+
+RESOLUTIONS = ["PER_FEATURE_IDX", "PER_FEATURE_NAME"]
```
ianspektor, line 20: Making this an enum could make it easier for documentation purposes and for the user to see their available choices when calling the operator. Defining it in core would be best to also use it there.

ianspektor: Also, I think making the default `PER_FEATURE_IDX` instead of `None` more accurately represents the operator's behavior

temporian/implementation/pandas/operators/tests/sum/sum_test.py
```diff
@@ -0,0 +1,38 @@
...
+from temporian.implementation.pandas.operators.tests.sum.test_data import (
+    different_shape,
+)
+
+
+class SumOperatorTest(absltest.TestCase):
+    def setUp(self) -> None:
+        self.operator = sum.PandasSumOperator()
+
+    def test_different_shape(self) -> None:
```
ianspektor, line 27: A test for the happy path would be useful too :) Maybe one using columns with different number types and checking they are preserved in output too.

ianspektor (commented):
Great work 😄 left some more comments!

temporian/core/operators/sum.py
diegomarvid: Yes it was! The loop changed though so I will resolve this conversation!

temporian/implementation/pandas/operators/sum.py
diegomarvid: Yeah, I thought about making an enum but didn't know well where to place it. I will try to create one in core/sum

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class SumOperator(Operator):
+    """Sum operator."""
+
+    def __init__(
+        self,
+        event_1: Event,
+        event_2: Event,
+        resolution: Optional[Resolution] = Resolution.PER_FEATURE_IDX,
```
ianspektor, line 41: This isn't `Optional` anymore, since it shouldn't be `None`

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,105 @@
...
+    ):
+        super().__init__()
+
+        # inputs
+        self.add_input("event_1", event_1)
+        self.add_input("event_2", event_2)
+
+        if resolution is not None:
```
ianspektor, line 49: This check isn't needed either, resolution will now always have a value

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,63 @@
...
+from temporian.implementation.pandas.data.event import PandasEvent
+from temporian.implementation.pandas.operators.base import PandasOperator
+
+
+class PandasSumOperator(PandasOperator):
+    def __call__(
+        self,
+        event_1: PandasEvent,
+        event_2: PandasEvent,
+        resolution: Optional[Resolution] = Resolution.PER_FEATURE_IDX,
```
ianspektor, line 29: Same, `Optional` not needed

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> PandasEvent:
+        """Sum two Events.
+
+        Args:
+            event_1: First Event.
+            event_2: Second Event.
+            resolution: Resolution of the output Event.
```
ianspektor, line 36: I liked the explanation there was on the behavior for each resolution option :)

temporian/implementation/pandas/operators/sum.py
```diff
@@ -0,0 +1,63 @@
...
+
+        Raises:
+            ValueError: If event_1 and event_2 have different shape.
+            NotImplementedError: If resolution is PER_FEATURE_NAME.
+        """
+        # raise value error if event_1 and event_2 have different shape
+        if event_1.shape != event_2.shape:
+            raise ValueError("event_1 and event_2 must have same shape.")
+
+        # sum each feautre index wise
```
ianspektor, line 49:
```suggestion
        # sum each feature index wise
```

ianspektor (commented):
One extra request: could we add the new operator to `prototype_test` to ensure the core <-> pandas impl link works as expected?

ianspektor (approved):

temporian/core/operators/sum.py
```diff
@@ -0,0 +1,110 @@
...
+
+class Resolution(str, Enum):
+    PER_FEATURE_IDX = "PER_FEATURE_IDX"
+    PER_FEATURE_NAME = "PER_FEATURE_NAME"
+
+
+class SumOperator(Operator):
+    """Sum operator."""
+
+    def __init__(
```
achoum, line 35: This method does not implements the PER_FEATURE_NAME resolution.
Note: It is fine to keep it for another PR

achoum (approved):
That's great!

temporian/implementation/numpy/operators/sum.py
```diff
@@ -1,8 +1,48 @@
...
+            resolution: Resolution of the output Event. PER_FEATURE_IDX sum is done feature index wise. PER_FEATURE_NAME sum is done feature name wise.
+
+        Returns:
+            Sum of the two Events according to resolution.
+
+        Raises:
+            IndexError: If index of both events is not equal.
+            NotImplementedError: If resolution is PER_FEATURE_NAME.
+        """
+        if event_1.data.keys() != event_2.data.keys():
```
ianspektor, line 29: Think it would be better to check the event's samplings are identical? Since that would also make sure that the timestamps present in each are the same

temporian/implementation/numpy/operators/sum.py
```diff
@@ -1,8 +1,48 @@
...
+            for i in range(len(event_index_array)):
+                event_1_feature = event_1.data[event_index][i]
+                event_2_feature = event_2.data[event_index][i]
+
+                output.data[event_index].append(
+                    (
+                        f"sum_{event_1_feature.name}_{event_2_feature.name}",
+                        event_1_feature.data + event_2_feature.data,
+                    )
+                )
```
ianspektor, line 46: I think we could skip this for loop by summing `event_1.data[event_index]` and `event_2.data[event_index]` directly and then setting the correct columns names?

ianspektor: Nope, event_1.data[event_index] is a list, not an array

PR #11, opened 2023-02-07
Initial NumPy backend

getchebarne:
This PR implements the initial NumPy backend. 

The main changes are in `temporian/implementation/numpy/data/event.py` - that script contains an initial implementation of the `NumpyEvent` and `NumpyFeature` classes. I also implemented the `NumpySelectOperator` and a simple `temporian/test/prototype_test_numpy.py` script to verify the backend is working as expected. The rest of the scripts were copied from the `pandas` backend and emptied.

I took the commits from the Select Operator PR (#8) since I needed the core SelectOperator to test the NumPy implementation. We should probably merge that PR one before this one.

Improvements to this code I can think of right now:
- Have `NumpyFeature` inherit from `numpy.ndarray` ([link](https://numpy.org/doc/stable/user/basics.subclassing.html))
- Have more user-friendly ways to create a `NumpyEvent`, i.e. with a friendlier API (current one is kinda verbose), parsing directly from `.csv` files, etc.
- Improve `NumpyEvent` and `NumpyFeature` `__repr__` methods

Disclaimer: I'm having trouble getting `assertEquals` to work with the materialized and expected output events, that's why I'm printing them instead (they are equal).


temporian/core/operators/select.py
```diff
@@ -0,0 +1,83 @@
...
+        if isinstance(feature_names, str):
+            feature_names = [feature_names]
+        self.add_attribute("feature_names", feature_names)
+
+        # inputs
+        self.add_input("event", event)
+
+        # outputs
+        output_features = []
+        for feature_name in feature_names:
```
achoum, line 40: I would fail if a requested feature does not exist.

temporian/core/data/event.py
```diff
@@ -25,15 +25,43 @@ def __init__(
...
 
-    def __repr__(self):
-        return f"Event<features:{self._features},sampling:{self._sampling},id:{id(self)}>"
+    def __getitem__(self, feature_names: List[str]) -> "Event":
+        # import select operator
+        from temporian.core.operators.select import SelectOperator
+
+        # instance select operator. Import from base temporian package
+        # to avoid circular import
+        select_operator = SelectOperator(self, feature_names)
```
achoum, line 42: Nit: Can you call the "select(...)" method directly?

temporian/core/operators/simple_moving_average.py
```diff
@@ -39,9 +39,9 @@ def __init__(
         if sampling is not None:
             self.add_input("sampling", sampling)
         else:
-            sampling = data.sampling()
+            sampling = event.sampling()
 
-        self.add_input("data", data)
+        self.add_input("event", event)
```
achoum, line 44: Nit: It is a bit hard to read the method because both the input and output are called "event" :)

temporian/implementation/numpy/data/event.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+
+        self.name = name
+        self.index = index
+        self.data = data
+        self.dtype = self.data.dtype
+
+    def schema(self) -> Feature:
+        return Feature(
+            name=self.name, dtype=self.dtype, sampling=self.index.names
```
achoum, line 25: We have our own dtype. There need to be some conversion from/to the numpy dtype.

temporian/implementation/numpy/data/sampling.py
```diff
@@ -0,0 +1,9 @@
+from typing import Dict, List, Tuple
+
+import numpy as np
+
+
+class NumpySampling:
+    def __init__(self, names: List[str], data: Dict[Tuple, np.ndarray]) -> None:
```
achoum, line 7: Is this like the "index" in NumpyFeature?

temporian/implementation/numpy/operators/tests/assign/test_data/different_index.py
```diff
@@ -0,0 +1,59 @@
...
+
+"""PandasAssignOperator - different index test.
+
+Tests that the assignee and assigned events cannot have different indexes.
+"""
+import pandas as pd
+
+from temporian.implementation.pandas.data.event import PandasEvent
+
+INPUT_1 = PandasEvent(
```
achoum, line 23: Should it be Numpy event?

achoum (approved):
LGTM;
Left a few comments.

temporian/core/operators/simple_moving_average.py
getchebarne: Good point - I made the change because we agreed that using `input_event` and `output_event` repeatedly was a bit too verbose. What other name could we use for these attributes?

temporian/implementation/numpy/data/sampling.py
getchebarne: It is - I'll update `NumpyFeature` to have this instead of `index`.

temporian/implementation/numpy/operators/tests/assign/test_data/different_index.py
getchebarne: Yeah, this script should be empty 😅 good catch!

diegomarvid (approved):

ianspektor (approved):

temporian/implementation/numpy/data/event.py
ianspektor: Will address in upcoming PR

PR #12, opened 2023-02-07
Remove the "place_holder" op

achoum:
The place_holder op was replaced by an optional utility method "input_event" which is similar (but with some extra check) as to a manually created Event object.


achoum:
Thanks :)

PR #14, opened 2023-02-08
Numpy assign operator

diegomarvid:
Adding NumPy backend for Assign operator. The current backend is in pandas.

- [x] Created `AssignNumpyOperator`.
- [x] Created Tests for `AssignNumpyOperator`.
- [x] Created `equals` for `NumpyEvent`, `NumpyFeature`, and `NumpySampling`.
- [x] Fixed minor error in `NumpySumOperator`.
- [x] Add tests for `NumpySumOperator`.
- [x] Removed **Sampling** attribute in `NumpyFeature`.


temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+from temporian.implementation.numpy.data.event import NumpyEvent
+
+
+class NumpyAssignOperator:
+    def __call__(
+        self, assignee_event: NumpyEvent, assigned_event: NumpyEvent
+    ) -> Dict[str, NumpyEvent]:
+        """Assign features to an event.
+
+        Input event and features must have same index. Features cannot have more
```
ianspektor, line 28: In this docstring, do `input event` and `features` refer to `assignee_event` and `assigned_event` respectively? Think we could make it more precise

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+            assigned_event (NumpyEvent): features to assign to the event.
+
+        Returns:
+            NumpyEvent: a new event with the features assigned.
+        """
+        output = NumpyEvent(data=assignee_event.data.copy(), sampling=assignee_event.sampling)
+
+        for index in assignee_event.data.keys():
+            if index in assigned_event.data.keys():
+                different_sampling = np.where(np.array(assignee_event.sampling.data[index]) != np.array(assigned_event.sampling.data[index]))[0]
```
ianspektor, line 44: Some more inline comments in this code could help with readability

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+                for i, _ in enumerate(assigned_event.data[index]):
+                    event_2_feature = assigned_event.data[index][i]
```
ianspektor, line 46: isn't this value the `_` you're discarding in the line above?

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+                    # get values with same timestamp as event 1
+                    event_2_feature = (event_2_feature[0], np.delete(event_2_feature[1], different_sampling))
+                    output.data[index].append(event_2_feature)
+            else:
+                for i, _ in enumerate(event_1[index]):
+                    output.data[index][i] += (None, [None for _ in range(len(sampling_1[index]))])
```
ianspektor, line 52: Not sure I understand this line - are we appending a tuple to the feature in the `i`th place of the `index` index value? Let's discuss live.

Also, `np.full(len(sampling_1[index]), np.nan)` might be of use here?

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+        understood as a left join on the index and timestamp columns.
+
+        Args:
+            assignee_event (NumpyEvent): event to assign the feature to.
+            assigned_event (NumpyEvent): features to assign to the event.
+
+        Returns:
+            NumpyEvent: a new event with the features assigned.
+        """
+        output = NumpyEvent(data=assignee_event.data.copy(), sampling=assignee_event.sampling)
```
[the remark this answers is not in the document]
ianspektor: Nope, only copying first level of `data` which is the (tuple, list[np.ndarray]) structure. np arrays not being copied. @diegomarvid will confirm this.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,55 @@
...
+
+        for index in assignee_event.data.keys():
+            if index in assigned_event.data.keys():
+                different_sampling = np.where(np.array(assignee_event.sampling.data[index]) != np.array(assigned_event.sampling.data[index]))[0]
+                for i, _ in enumerate(assigned_event.data[index]):
+                    event_2_feature = assigned_event.data[index][i]
+                    # get values with same timestamp as event 1
+                    event_2_feature = (event_2_feature[0], np.delete(event_2_feature[1], different_sampling))
```
ianspektor, line 48: Might be easier to understand if calculating `same_sampling` instead of `different_sampling` and then keeping only those instead of deleting the different ones?

temporian/implementation/numpy/operators/assign.py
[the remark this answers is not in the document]
diegomarvid: Yes, `copy` is a shallow copy so everything inside the array in the keys is a reference. See https://docs.python.org/3/library/copy.html#:~:text=A%20shallow%20copy%20constructs%20a,objects%20found%20in%20the%20original. 

> A shallow copy constructs a new compound object and then (to the extent possible) inserts references into it to the objects found in the original.

> A deep copy constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original.

temporian/implementation/numpy/data/event.py
```diff
@@ -62,6 +93,27 @@ def schema(self) -> Event:
...
+        # check same features
+        if self.feature_names != __o.feature_names:
+            return False
+
+        # Check each feature is equal in each index
+        for index in self.data.keys():
+            for feature in self.data[index]:
+                if feature not in __o.data[index]:
+                    print(f"Feature {feature} not in {__o.data[index]}")
+                    return False
```
ianspektor, line 113: Here we are checking that all of `self`'s features are in `__o`, but not other way round?

If we want to make equal events necessarily have features in the same order (which I think could make sense), we could just check that `all(f1 == f2 for f1, f2 in zip(self.data[index], __o.data[index]))` for example, which is also faster (with the `in` operator we are getting O(n^2) since we are traversing the second list for each element in the first one).

Also, would remove this print.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+    return np.where(sampling_1.data[index] == sampling_2.data[index])[0]
+
+
+class NumpyAssignOperator:
+    def __call__(
+        self, assignee_event: NumpyEvent, assigned_event: NumpyEvent
+    ) -> Dict[str, NumpyEvent]:
+        """Assign features to an event.
+
+        Assignee and assigned must have same index names. Features cannot have more
```
ianspektor, line 46:
```suggestion
        Assignee and assigned must have same index names. Assigned cannot have more
```

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+        than one row for a single index + timestamp occurence. Output event will
+        have same exact index and timestamps (sampling) as the assignee event.
+
+        Assignment is done by matching the timestamps and index of the assignee and assigned.
+        The assigned features will be appended to the assignee features in the matching
+        indexes. If assignee event has more indexes values than assigned, the assigned features
+        will be broadcasted to the assignee indexes with np.nan values according
+        to the assignee sampling. If the assigned event has more timestamps in a matching
+        index, then those values will not be included in the output event.
```
ianspektor, line 55: "Index values in the assigned event missing in the assignee event will be filled with NaNs. Index values present in the assigned event missing in the assignee event will be ignored." might be simpler to understand?

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+
+
```
ianspektor, line 57: Duplicated newline

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+        Args:
+            assignee_event (NumpyEvent): event to assign the feature to.
```
ianspektor, line 59:
```suggestion
            assignee_event (NumpyEvent): event to assign features to.
```

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns:
+            NumpyEvent: a new event with the features assigned.
+
+        Raises:
+            ValueError: if assignee and assigned events have different indexes names.
+
+        """
+        # check both keys are the same
+        if assignee_event.sampling.names != assigned_event.sampling.names:
+            raise ValueError("Assign sequences must have the same index names.")
```
ianspektor, line 71:
```suggestion
            raise ValueError("Assigned and assignee events must have the same index names.")
```

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        sampling_1 (NumpySampling): sampling of assignee event.
+        sampling_2 (NumpySampling): sampling of assigned event.
+        index (tuple): index to check.
+
+    Returns:
+        np.ndarray: indexes of timestamps that are equal to sampling_1 & sampling_2.
+    """
+    return np.where(sampling_1.data[index] == sampling_2.data[index])[0]
```
ianspektor, line 37: We could move this utility function to below the main class def

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+"""Implementation for the Assign operator."""
+from typing import Dict, List
+
+import numpy as np
+from temporian.implementation.numpy.data.event import NumpyEvent
+from temporian.implementation.numpy.data.event import NumpyFeature
+
+
+def _get_equal_sampling_indexes(
+    sampling_1: "NumpySampling", sampling_2: "NumpySampling", index: tuple
```
ianspektor, line 24: Why are we using string type hints?

temporian/implementation/numpy/operators/assign.py
ianspektor, line 37: If I understand correctly, this is checking that for each position in the timestamps list, the timestamp is equal between both arrays. What happens if the two samplings have timestamps in common, but not in the exact same positions? E.g.:
```
s1.data[index] = [1, 2, 3]
s2.data[index] = [2, 3, 4]
```
in this case we would want the first and second positions in `s2` to be assigned to the second and third positions in `s1`. Can we write a test for this case?

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class NumpyAssignOperator:
+    def __call__(
+        self, assignee_event: NumpyEvent, assigned_event: NumpyEvent
+    ) -> Dict[str, NumpyEvent]:
+        """Assign features to an event.
+
+        Assignee and assigned must have same index names. Features cannot have more
+        than one row for a single index + timestamp occurence. Output event will
```
ianspektor, line 47: Are we actually checking that there are no duplicated sampling values in the assigned event?

temporian/implementation/numpy/operators/assign.py
ianspektor: (or remove it and just drop an in-line comment, since it's a one liner and used in a single place)

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,110 @@
...
+                # get indexes of timestamps that are equal to sampling_1 & sampling_2
+                equal_sampling_indexes = _get_equal_sampling_indexes(
+                    sampling_1=assignee_event.sampling,
+                    sampling_2=assigned_event.sampling,
+                    index=index,
+                )
+                # loop over assigned features
+                for assigned_feature in assigned_event.data[index]:
+                    # filter indexes of timestamps that are equal to sampling_1 & sampling_2
+                    assigned_feature.data = np.take(
```
ianspektor, line 89: Isn't this assigment modifying the content of the `assigned_event` we received?

temporian/implementation/numpy/operators/sum.py
```diff
@@ -39,9 +40,12 @@ def __call__(self, event_1: NumpyEvent, event_2: NumpyEvent) -> NumpyEvent:
...
                 output.data[event_index].append(
-                    (
-                        f"sum_{event_1_feature.name}_{event_2_feature.name}",
-                        event_1_feature.data + event_2_feature.data,
+                    NumpyFeature(
+                        name=(
+                            f"sum_{event_1_feature.name}_{event_2_feature.name}"
+                        ),
+                        data=event_1_feature.data + event_2_feature.data,
+                        sampling=event_1_feature.sampling,
```
ianspektor, line 48: Good catch 💯

temporian/implementation/numpy/operators/tests/assign/assign_test.py
```diff
@@ -0,0 +1,83 @@
...
+        operator_output = self.operator(
+            with_idx_same_timestamps.INPUT_1,
+            with_idx_same_timestamps.INPUT_2,
+        )
+        self.assertEqual(
+            True,
+            with_idx_same_timestamps.OUTPUT == operator_output["event"],
+        )
+
+    def test_broadcast_assigned(self) -> None:
```
ianspektor, line 67: The word `broadcast` here might be confusing later on when we implement the broadcast operator. Maybe "test_less_index_values"?

temporian/test/prototype_test_numpy.py
```diff
@@ -151,41 +172,6 @@ def setUp(self) -> None:
...
-        #             NumpyFeature(
-        #                 name="costs",
-        #                 sampling=sampling,
-        #                 data=np.array([1, 2]),
-        #             ),
-        #         ],
-        #     },
-        #     sampling=sampling,
-        # )
-
```
ianspektor, line 188: Remove

temporian/test/prototype_test_numpy.py
```diff
@@ -151,41 +172,6 @@ def setUp(self) -> None:
...
     def test_prototoype(self) -> None:
```
ianspektor, line 175:
```suggestion
    def test_prototype(self) -> None:
```

temporian/implementation/numpy/data/event.py
diegomarvid: Practically, features in a different event should still be the same event because they have the same information. However, I agree that is easier and faster to just check that every feature is in the same order

temporian/implementation/numpy/operators/assign.py
diegomarvid: good catch, fixing it with extra border cases

temporian/implementation/numpy/operators/assign.py
diegomarvid: not yet, waiting for our conversation with @getchebarne to talk about data duplication in assignee and assigned

temporian/implementation/numpy/operators/tests/assign/assign_test.py
diegomarvid: sure, i will change it

temporian/implementation/numpy/operators/assign.py
```diff
@@ -18,23 +18,62 @@
...
 
-def _get_equal_sampling_indexes(
-    sampling_1: "NumpySampling", sampling_2: "NumpySampling", index: tuple
+def _get_common_timestamps(
+    sampling_1: NumpySampling, sampling_2: NumpySampling, index: tuple
 ) -> np.ndarray:
-    """Get indexes of timestamps that are equal to sampling_1 & sampling_2
-      in a specific sampling index.
+    """It returns a np.ndarray with same shape as sampling_1.data[index] with True
+    values where sampling_1.data[index] == sampling_2.data[index], and False otherwise.
```
ianspektor, line 28: This description doesn't seem correct - "It returns a np.ndarray with same shape as sampling_1.data[index] indicating if each value in it is present in sampling_2.data[index]" maybe? `==` would check positionally which isn't the case

temporian/implementation/numpy/operators/assign.py
diegomarvid: agree

temporian/test/prototype_test_numpy.py
diegomarvid: it's in read, it means it already has been removed. Almost got me there

temporian/implementation/numpy/operators/core_mapping.py
```diff
@@ -1,8 +1,10 @@
 from temporian.implementation.numpy.operators.select import NumpySelectOperator
 from temporian.implementation.numpy.operators.sum import NumpySumOperator
+from temporian.implementation.numpy.operators.assign import NumpyAssignOperator
```
getchebarne, line 3: Wrong import order - `from temporian.implementation.numpy.operators.assign import NumpyAssignOperator` should go first.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    return np.isin(sampling_1.data[index], sampling_2.data[index])
+
+
+def _convert_feature_to_new_sampling(
+    feature: NumpyFeature,
+    common_timestamps: List[bool],
+    new_sampling: NumpySampling,
+    index: tuple,
+):
```
getchebarne, line 69: Missing return type.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,193 @@
...
+    """Convert feature to new sampling. It requires a boolean list with same length
+    as the new sampling, where True values indicate that the feature has a value
+    for that timestamp. And it needs the new sampling to assign to the new Feature object.
+
+    Args:
+        feature (NumpyFeature): feature to convert.
+        common_timestamps (List[bool]): list of booleans indicating if sampling_2 has same timestamp in sampling_1
```
getchebarne, line 76: We haven't named `sampling_2` and `sampling_1` in this docstring yet - I'm assuming they are `new_sampling` and the feature's sampling respectively, correct? In that case, we should reflect that in the documentation.

temporian/implementation/numpy/operators/tests/sum/sum_test.py
```diff
@@ -0,0 +1,73 @@
...
+)
+from temporian.implementation.numpy.operators.tests.sum.test_data import (
+    different_timestamps,
+)
+from temporian.implementation.numpy.operators.tests.sum.test_data import (
+    sum_with_nan,
+)
+
+
+class AssignOperatorTest(absltest.TestCase):
```
getchebarne, line 29: Shouldn't this be `SumOperatorTest`?

temporian/test/prototype_test_numpy.py
```diff
@@ -21,6 +21,7 @@
 from temporian.core.data.sampling import Sampling
 from temporian.core.operators.select import select
 from temporian.core.operators.sum import sum
+from temporian.core.operators.assign import assign
```
getchebarne, line 24: Fix import order.

temporian/implementation/numpy/data/event.py
```diff
@@ -51,6 +66,22 @@ def __init__(
...
+        if len(self.data.keys()) == 0:
+            return 0
+
+        first_index = next(iter(self.data))
+        return len(self.data[first_index])
+
+    @property
+    def feature_names(self) -> List[str]:
+        if len(self.data.keys()) == 0:
+            return 0
```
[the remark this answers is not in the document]
diegomarvid: Agree, I think it got overlooked by mistake after doing feature_counts

temporian/test/prototype_test_numpy.py
ianspektor: lol. oops.

diegomarvid:
Another option is to further encapsulate `NumpySampling` and convert it from a flat dictionary with `numpy.ndarray` to a List of objects such as `SamplingTimestamps`. Each object has an **index** and a **list of timestamps**. With this, we can have the `NumpyFeature` be able to just refer to the specific index, and the event to refer to the whole `NumpySampling`.

temporian/implementation/numpy/data/event.py
```diff
@@ -62,6 +93,27 @@ def schema(self) -> Event:
...
+        for index in self.data.keys():
+            if not all(
+                f1 == f2 for f1, f2 in zip(self.data[index], __o.data[index])
+            ):
+                return False
+
+        return True
+
 
 DTYPE_MAPPING = {
```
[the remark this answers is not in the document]
getchebarne: yes

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,200 @@
...
+    # Assuming that the feature data is sorted by timestamp, we can use binary search
+    # to find the index of the common timestamp in the feature data, instead of
+    # numpy where search. This would make the algorithm O(n * (1 + log(m)))
+    # (n: len(common_timestamps), m: len(feature.data). Adding extra n for
+    # np.full at the beginning.
+    # We can make this more efficient by trimming the feature data to the
+    # last found array index because as it sorted by timestamps, the next one
+    # will be after that one.
+
+    for i, is_common in enumerate(common_timestamps):
```
getchebarne, line 102: I think we can circumvent the `for` loop here and replace it with a vectorized implementation.
```
mask = (
        assignee_sampling.data[index][:, np.newaxis]
        == assigned_sampling.data[index][np.newaxis, :]
)
mask_i, mask_j = mask.nonzero()
new_feature.data[mask_i] = feature.data[mask_j]
```
This implementation also avoids using the `np.in1d` in `_get_common_timestamps`. Note: this only works if the assignee event does not contain repeated timestamps. We need to make some changes to the code to support that case.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,97 @@
...
+            raise ValueError(
+                "Assigned sequence cannot have repeated timestamps in the same"
+                " index."
+            )
+
+        output = NumpyEvent(
+            data=assignee_event.data.copy(), sampling=assignee_event.sampling
+        )
+
+        for index in assignee_event.data.keys():
```
achoum, line 65: Use ".items()" to avoid the query "assignee_event.sampling.data[index]" on line 77.

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,97 @@
...
+                mask_i, mask_j = mask.nonzero()
+                for feature in assigned_event.data[index]:
+                    new_feature = NumpyFeature(
+                        name=feature.name,
+                        data=np.full(
+                            len(assignee_event.sampling.data[index]), np.nan
+                        ),
+                    )
+                    new_feature.data[mask_i] = feature.data[mask_j]
+                    output.data[index].append(new_feature)
```
achoum, line 81: Nit: Can you cache the result of the query to "output.data[index]" over the "for feature in assigned_event.data[index]:" loop?

temporian/implementation/numpy/operators/tests/assign/BUILD
```diff
@@ -12,8 +12,11 @@ py_test(
     deps = [
         # absl/testing:absltest dep,
         "//temporian/implementation/numpy/operators:assign",
+        "//temporian/implementation/numpy/operators/tests/assign/test_data:assigned_repeated_timestamps",
```
achoum, line 15: Nit: Make sure to fix formatting.

temporian/implementation/numpy/operators/tests/assign/assign_test.py
```diff
@@ -0,0 +1,132 @@
...
+            assigned_repeated_timestamps.INPUT_2,
+        )
+
+    def test_assignee_repeated_timestamps(self) -> None:
+        operator_output = self.operator(
+            assignee_repeated_timestamps.INPUT_1,
+            assignee_repeated_timestamps.INPUT_2,
+        )
+
+        print("Actual:")
```
achoum, line 68: Note that "print" don't work well in tests.
You can either use logging.info (from the absl package; best solution) or do "print(..., flush=True)" (for quick debugging).

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,97 @@
...
+                mask = (
+                    assignee_event.sampling.data[index][:, np.newaxis]
+                    == assigned_event.sampling.data[index][np.newaxis, :]
+                )
+                mask_i, mask_j = mask.nonzero()
+                for feature in assigned_event.data[index]:
+                    new_feature = NumpyFeature(
+                        name=feature.name,
+                        data=np.full(
+                            len(assignee_event.sampling.data[index]), np.nan
```
achoum, line 77: Nit; For readability: What about saving "len(assignee_event.sampling.data[index])" in a variable called "num_timestamps", and use "num_timestamps" instead (in both places)?

achoum (commented):
This looks good :)

temporian/implementation/numpy/operators/assign.py
```diff
@@ -0,0 +1,97 @@
...
+
+import numpy as np
+from temporian.implementation.numpy.data.event import NumpyEvent
+from temporian.implementation.numpy.data.event import NumpyFeature
+from temporian.implementation.numpy.data.sampling import NumpySampling
+
+
+class NumpyAssignOperator:
+    def __call__(
+        self, assignee_event: NumpyEvent, assigned_event: NumpyEvent
```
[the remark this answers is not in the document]
getchebarne: I like it.

temporian/implementation/numpy/operators/assign.py
diegomarvid: We are looping over the event so the query `assignee_event.sampling.data[index]` has the value for the sampling data, not the event data.

temporian/implementation/numpy/operators/tests/assign/BUILD
diegomarvid: Can you explain what should be fixed? I don't seem to realize. Is the order of the tests?

temporian/implementation/numpy/operators/assign.py
achoum: My bad.

temporian/implementation/numpy/operators/tests/assign/BUILD
achoum: (In github) I see the lines 15 and 16 as not aligned the same way as the other lines (e.g. 17-21).

achoum:
Once the remaining comments addressed, this looks good to me.

achoum (approved):

getchebarne (approved):

PR #15, opened 2023-02-08
Map to core dtypes in NumpyFeature

ianspektor:
Make `NumpyFeature`'s `schema()` return one of core's dtype instead of numpy one.
Addresses [this comment](https://github.com/google/temporian/pull/11#discussion_r1099117911) in np backend implementation.

Also added 32 bit int and float to available types in proto def.


achoum (approved):
Thanks

PR #16, opened 2023-02-09
add bazel tests to github action

ianspektor:
Added GitHub actions to check that tests pass and code is formatted correctly.
Added badges of both actions to readme.

⚠️ Using python 3.9 in both actions. We may want to downgrade it if we aim to support older versions.

# Tests action

Runs bazel tests.
Used --test_output=errors for easy debugging of errors that occur in action.
Bazelisk comes preinstalled in linux runners so no need to install it.
Used `actions/cache` to store bazel cache for reduced action runtime.
Poetry is installed and environment is also cached thanks to the `cache: 'poetry'` arg of `setup-python` action.

# Formatting action

Very similar to tests one, but drops the bazel caching step, and runs `black --check` instead of `bazel test`.

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -3,7 +3,7 @@ name: formatting
 on: push
 
 jobs:
-  test:
+  formatting:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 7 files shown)


diegomarvid (approved):

achoum:
That is awesome. Thanks!

achoum (approved):

PR #19, opened 2023-02-12
Implement part of the SMA numpy operator

achoum:
In addition to this partial implementation,  I've tried some changes in the code organization & style (https://docs.google.com/document/d/1cgh_1lawKu2xD_QIZVv7JM9h3cIBkyFhO-EIIfR-IO8/edit#heading=h.yikhuc3t00vk).


temporian/implementation/numpy/data/event.py
```diff
@@ -1,44 +1,37 @@
 from typing import Dict, List, Tuple, Any
 
 import numpy as np
-
+import pandas as pd
```
getchebarne, line 4: Unused import.

temporian/implementation/numpy/data/event.py
```diff
@@ -1,44 +1,37 @@
...
             raise ValueError(
                 "NumpyFeatures can only be created from flat arrays. Passed"
                 f" input's shape: {len(data.shape)}"
             )
-        if data.dtype.type not in DTYPE_MAPPING:
-            raise ValueError(
-                f"Unsupported dtype {data.dtype} for NumpyFeature. Supported"
-                f" dtypes: {DTYPE_MAPPING.keys()}"
-            )
+        if data.dtype.type is np.string_:
```
getchebarne, line 18: What if the dtype is not string and it's also not supported by `DTYPE_MAPPING`?

temporian/implementation/numpy/data/event.py
```diff
@@ -1,44 +1,37 @@
...
-            dtype=self.core_dtype(),
-            sampling=self.sampling.names,
-        )
-
     def __repr__(self) -> str:
         return f"{self.name}: {self.data.__repr__()}"
 
     def core_dtype(self) -> Any:
+        if self.dtype.type is np.string_:
+            return dtype.STRING
```
getchebarne, line 34: Why not add add `dtype.STRING` to `DTYPE_MAPPING`?

temporian/implementation/numpy/data/event.py
```diff
@@ -51,16 +44,8 @@ def __init__(
...
         self.sampling = sampling
 
-    def schema(self) -> Event:
-        return Event(
-            features=[
-                feature.schema() for feature in list(self.data.values())[0]
-            ],
-            sampling=self.sampling.names,
-        )
-
```
getchebarne, line 61: What's the motivation behind removing the ability to create schemas from materialized data?

temporian/implementation/numpy/data/sampling.py
```diff
@@ -4,15 +4,18 @@
...
-    def __init__(self, names: List[str], data: Dict[Tuple, np.ndarray]) -> None:
-        self.names = names
+    def __init__(self, index: List[str], data: Dict[Tuple, np.ndarray]) -> None:
+        self.index = index
         self.data = data
 
+    def __repr__(self) -> str:
+        return f"index:{self.index} data: {self.data.__repr__()}"
+
     def __eq__(self, other):
```
getchebarne, line 14: Nit: add `__eq__` input and output return types, i.e. `other : Any` and `-> bool`.

temporian/implementation/numpy/operators/window/simple_moving_average.py
```diff
@@ -0,0 +1,60 @@
...
+        # Note: mts => multi time series
+        # Note: ts => time series
+        for index, src_mts in event.data.items():
+            dst_mts = []
+            dst_event.data[index] = dst_mts
+            src_time = event.sampling.data[index]
+
+            # For each feature
+            for src_ts in src_mts:
+                dst_feature_name = f"sum_{src_ts.name}"
```
getchebarne, line 40: Nit: `sma` prefix instead of `sum`

temporian/implementation/numpy/operators/window/simple_moving_average.py
```diff
@@ -0,0 +1,60 @@
...
+def _impl(src: np.array, time: np.array, win: Duration) -> np.array:
+    # TODO: Implement in c++.
+    # TODO: Add support for missing (NaN) input values.
+
+    # This implementation is simple but expensive. It will create multiple
+    # O(n^2) arrays, where n is the number of time samples.
+    r = time[:, np.newaxis] >= time[np.newaxis, :]  # Right side
+    l = time[:, np.newaxis] <= (time[np.newaxis, :] + win)  # Left side
+    m = r & l
+    dst = np.sum(src * m, axis=1) / np.sum(m, axis=1)
```
getchebarne, line 58: Neat 💯

temporian/implementation/numpy/operators/window/simple_moving_average.py
```diff
@@ -0,0 +1,60 @@
...
+        # For each index
+        # Note: mts => multi time series
+        # Note: ts => time series
+        for index, src_mts in event.data.items():
+            dst_mts = []
+            dst_event.data[index] = dst_mts
+            src_time = event.sampling.data[index]
+
+            # For each feature
+            for src_ts in src_mts:
```
getchebarne, line 39: We could also concatenate `src_mts` generating an array of shape `m x n`, where `m` is the number of time series and `n` the number of samples of each series. We could then vectorize the entire implementation instead of looping over each time series independently. 

I'm not exactly sure about the overhead concatenating the time series introduces vs. the performance we gain from the vectorization (and how it scales with the number and length of the series) though. We can research about it.

temporian/implementation/numpy/operators/window/simple_moving_average_test.py
```diff
@@ -0,0 +1,92 @@
...
+from temporian.core.data import dtype as dtype_lib
+
+
+class SimpleMovingAverageOperatorTest(absltest.TestCase):
+    def setUp(self):
+        pass
+
+    # TODO: Add more tests (index, sampling, nan, etc).
+
+    def test_no_index(self):
```
ianspektor, line 37: Let's try to reuse `PandasSimpleMovingAverageOperator`'s tests which cover several cases (though I agree with trying to make them as small as possible to avoid having data in separate files)

temporian/implementation/numpy/data/event.py
achoum: From experimenting, it seems a string can be different "data.dtype" values (e.g. depending on the encoding, or maximum length). "data.dtype.type" is the most consistent solution I found.

Note: Ultimately, we probably want to normalize the exactly used string type e.g. either only allowing one type (with automatic conversion), or be explicit with the exact sub-type.

temporian/implementation/numpy/data/event.py
achoum: See comment above.
Short story: It would not work.

temporian/implementation/numpy/data/event.py
achoum: From a design perspective, the semantic of a schema created by the data is not clear to me. Let brainstorm about it in our next meeting:

In the mean time, here are some thoughts:

- The graph defines the computation (which include schema i.e. input and output type of each step).
- The data is something that can be feed into the graph.
- The data feed by the user might be different from the data received by the graph. For example, the current code automatically read csv files when the data looks like a path. I suppose we will also implement automatic numerical conversions (e.g. the schema is a int8 but the user feed a int32; => The data is automatically converted to int8).

temporian/implementation/numpy/data/sampling.py
achoum: Done

temporian/implementation/numpy/operators/window/simple_moving_average.py
achoum: Good catch.

The op is only tested manually. Ultimately, the execution code should  check the input/output of implementation execution.

temporian/implementation/numpy/operators/window/simple_moving_average.py
achoum: Some thoughts:
- Different time series might have different number of observations.
- When using np, it is easy to write memory inefficient algorithms. For example, the current algorithm memory is O(n^2) with n the maximum time series length. This is too large (ultimately, we don't want to exceed O(mn); with m the number of observations in a window). I think the memory complexity of your proposed algorithm is O(n^2p) with p the number of time series.

temporian/implementation/numpy/data/event.py
```diff
@@ -1,44 +1,37 @@
...
                 "NumpyFeatures can only be created from flat arrays. Passed"
                 f" input's shape: {len(data.shape)}"
             )
-        if data.dtype.type not in DTYPE_MAPPING:
-            raise ValueError(
-                f"Unsupported dtype {data.dtype} for NumpyFeature. Supported"
-                f" dtypes: {DTYPE_MAPPING.keys()}"
-            )
+        if data.dtype.type is np.string_:
+            if data.dtype.type not in DTYPE_MAPPING:
```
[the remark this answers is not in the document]
achoum: Good catch.
*Fixed

temporian/core/operators/simple_moving_average.py
```diff
@@ -21,6 +21,8 @@
 from temporian.core.data.feature import Feature
 from temporian.core.operators.base import Operator
 from temporian.proto import core_pb2 as pb
+from temporian.core.data import duration
```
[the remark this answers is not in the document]
achoum: Done

temporian/implementation/numpy/data/event.py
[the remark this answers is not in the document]
achoum: See discussion above.
Short story, DTYPE_MAPPING index self.dtype, not self.dtype.type.
A string can have many different self.dtype values (notably, because there is a separate self.dtype for each maximum string length).

temporian/implementation/numpy/operators/window/simple_moving_average.py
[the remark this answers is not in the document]
achoum: Fixed

temporian/implementation/numpy/operators/window/simple_moving_average.py
```diff
@@ -0,0 +1,60 @@
...
+
+        dst_event = NumpyEvent(data={}, sampling=event.sampling)
+
+        # For each index
+        # Note: mts => multi time series
+        # Note: ts => time series
+        for index, src_mts in event.data.items():
+            dst_mts = []
+            dst_event.data[index] = dst_mts
+            src_time = event.sampling.data[index]
```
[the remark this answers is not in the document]
achoum: Agreed. 
s/time/timestamps

temporian/implementation/numpy/operators/window/simple_moving_average_test.py
achoum: I agree we want to reuse them (added details in todo in test file).

As you can see, the test is still too verbosity. We need to create the "pd_to_event" function and use it in the tests (see toso). With "pd_to_event", the pandas tests will be usable directly.
However, creating this "pd_to_event" function should probably be a PR on its own.

Do you you think about creating "pd_to_event" and moving the tests in two separate following PRs?

temporian/core/operators/simple_moving_average.py
```diff
@@ -64,6 +66,9 @@ def __init__(
 
         self.check()
 
+    def window_length(self):
+        return self._window_length
+
     @classmethod
     def build_op_definition(cls) -> pb.OperatorDef:
         return pb.OperatorDef(
```
diegomarvid, line 74: Should the `window_length` attribute be a float now?

`type=pb.OperatorDef.Attribute.Type.STRING,`

temporian/core/data/BUILD
```diff
@@ -43,3 +43,10 @@ py_library(
         ":sampling",
     ],
 )
+
```
diegomarvid, line 46: Should you add duration dependency on sma core ?

```
py_library(
    name = "simple_moving_average",
    srcs = ["simple_moving_average.py"],
    srcs_version = "PY3",
    deps = [
        ":base",
        "//temporian/core:operator_lib",
        "//temporian/core/data:duration", --> NEW
        "//temporian/core/data:event",
        "//temporian/core/data:feature",
        "//temporian/proto:core_py_proto",
    ],
)
```

temporian/core/operators/simple_moving_average.py
```diff
@@ -29,11 +30,12 @@ class SimpleMovingAverage(Operator):
     def __init__(
         self,
         event: Event,
-        window_length: str,
+        window_length: Duration,
         sampling: Optional[Event] = None,
     ):
         super().__init__()
 
+        self._window_length = window_length
```
getchebarne, line 38: How does the `_window_length` attribute integrate with the proto definition? Can attributes be defined outside of the `attributes` attribute? (little bit of a tongue twister, I know 😂 )

temporian/core/operators/simple_moving_average.py
achoum: Good catch.
*Done

temporian/core/data/BUILD
achoum: Good catch.
*Solved.

Note: In practice, python+bazel seems to allow for indirect dependencies i.e. this code compiled because on the of dependencies of :simple_moving_average was depending on :duration.

temporian/core/operators/simple_moving_average.py
achoum: Any field can be created in the operator.

The current logic is that during serialization, an op object is created with the constructor argument registered with the "add_{attribute,output,inputs}" function. While we don't have such op anymore (placeholder op was one of them), we can have operators where the attributes don't need to be serialized.

Later on, would could imagine improvement to this logic. For example, maybe having all the constructor argument passed to "super().__init__()" and registered there would work.

achoum:
One this pr is approved, I can continue working the SMA.

getchebarne (approved):

achoum:
Thanks

PR #21, opened 2023-02-14
Lag operator

diegomarvid:
In this PR I will implement the lag operator

- [ ] Core Lag Operator
- [ ] Numpy Lag Operator
- [ ] Tests for Numpy Lag Operator
- [ ] Add test in `prototype_test_numpy.py`


temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+# limitations under the License.
+
+"""Lag operator."""
+
+from temporian.core import operator_lib
+from temporian.core.data.event import Event
+from temporian.core.data.feature import Feature
+from temporian.core.operators.base import Operator
+from temporian.proto import core_pb2 as pb
+from temporian.core.data.duration import Duration
```
getchebarne, line 22: Fix import order

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+        self.add_input("event", event)
+
+        self.add_attribute("duration", duration)
+
+        # outputs
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"lag_{f.name()}",
+                dtype=f.dtype(),
+                sampling=event.sampling,
```
getchebarne, line 45: To think about: should core `Feature`s keep having a reference to a core `Sampling`? We removed that in their NumPy counterparts. What are we gaining from core features having a core sampling, since they are attached to a core event that already has one? cc: @ianspektor @achoum

temporian/core/operators/simple_moving_average.py
```diff
@@ -21,6 +21,7 @@
 from temporian.core.data.feature import Feature
 from temporian.core.operators.base import Operator
 from temporian.proto import core_pb2 as pb
+from temporian.core.data.duration import Duration
```
getchebarne, line 24: Import order.

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+                creator=self,
+            )
+            for f in event.features()
+        ]
+
+        self.add_output(
+            "event",
+            Event(
+                features=output_features,
+                sampling=event.sampling,
```
getchebarne, line 55: We're modifying the sampling right? In that case, the `sampling` argument of the output event shouldn't be the same as the input's - we need to create a new sampling.

temporian/implementation/numpy/operators/lag.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def __call__(self, event: NumpyEvent) -> NumpyEvent:
+        new_sampling = NumpySampling(
+            data={},
+            index=event.sampling.index,
+        )
+        output = NumpyEvent(data=event.data, sampling=new_sampling)
+
+        for index, timestamps in event.sampling.data.items():
+            new_sampling.data[index] = timestamps - self.duration
```
getchebarne, line 19: I think this should be a `+`.

temporian/implementation/numpy/operators/tests/lag/test_data/correct_lag.py
```diff
@@ -0,0 +1,58 @@
...
+                data=np.array([10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0]),
+            ),
+        ],
+    },
+    sampling=sampling,
+)
+
+output_sampling = NumpySampling(
+    index=["store_id"],
+    data={("A",): np.array([-1, -0.5, 1, 1.5, 2, 8, 18])},
```
getchebarne, line 45: Should be `np.array([3, 3.5, 5, 5.5, 6, 12, 22])`.

getchebarne (dismissed):

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+                creator=self,
+            )
+            for f in event.features()
+        ]
+
+        self.add_output(
+            "event",
+            Event(
+                features=output_features,
+                sampling=Sampling(index=event.sampling().index, creator=self),
```
achoum, line 55: Reuse the same Sampling object as for the features.

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+
+from temporian.core import operator_lib
+from temporian.core.data.event import Event
+from temporian.core.data.feature import Feature
+from temporian.core.data.sampling import Sampling
+from temporian.core.operators.base import Operator
+from temporian.proto import core_pb2 as pb
+
+
+class LagOperator(Operator):
```
achoum, line 25: I would use the same class for lag and leak, and expose them to the user with a different function i.e. t.lag and t.leak

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # inputs
+        self.add_input("event", event)
+
+        self.add_attribute("duration", duration)
+
+        # outputs
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"lag_{f.name()}",
```
achoum, line 43: Make the prefix dependent on the sign of the duration.

temporian/implementation/numpy/operators/lag.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__()
+        self.operator = operator
+
+    def __call__(self, event: NumpyEvent) -> NumpyEvent:
+        duration = self.operator.attributes()["duration"]
+
+        if duration <= 0:
+            raise ValueError("Duration must be positive and non-zero.")
+
+        if isinstance(self.operator, LeakOperator):
```
achoum, line 20: No need if there is only one operator class

temporian/implementation/numpy/operators/lag.py
```diff
@@ -0,0 +1,35 @@
...
+            data={},
+            names=event.sampling.names,
+        )
+        output = NumpyEvent(data=event.data, sampling=new_sampling)
+
+        for index, timestamps in event.sampling.data.items():
+            new_sampling.data[index] = timestamps + duration
+            output_data = output.data[index]
+            for feature in output_data:
+                feature.name = f"lag_{feature.name}"
```
achoum, line 33: This is modifying the data of the input argument. Let's avoid this.

temporian/core/operators/lag.py
diegomarvid: Sounds better in terms of reusability 💯

temporian/core/operators/lag.py
diegomarvid: Yes, it's part of the goal. But I don't have the `Duration` class and any `Duration` helper function. I could wait for you to merge the PR of SMA and then rebase to main in order to have the functions. If you think that the SMA PR will take much more to merge, you can open a PR only for `Duration` with its functions and merge that PR to main.

temporian/implementation/numpy/operators/lag.py
```diff
@@ -0,0 +1,32 @@
...
+        self.operator = operator
+
+    def __call__(self, event: NumpyEvent) -> NumpyEvent:
+        duration = self.operator.attributes()["duration"]
+
+        new_sampling = NumpySampling(
+            data={},
+            names=event.sampling.names.copy(),
+        )
+        output = NumpyEvent(data={}, sampling=new_sampling)
```
getchebarne, line 19: also: let's try to use `output_event` instead of `output`

temporian/implementation/numpy/operators/tests/lag/lag_test.py
```diff
@@ -0,0 +1,95 @@
...
+                        data=np.array(
+                            [10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0]
+                        ),
+                    ),
+                ],
+            },
+            sampling=numpy_output_sampling,
+        )
+
+        sampling = Sampling(["store_id"])
```
getchebarne, line 73: nit: define this on line 76, i.e. `sampling=Sampling(["store_id"])`. Makes the code shorter

temporian/implementation/numpy/operators/tests/lag/lag_test.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2021 Google LLC.
```
getchebarne, line 1: I think we agreed on dumping test scripts directly into the `tests` directory instead of creating a subdirectory (`/lag` in this case)

temporian/implementation/numpy/operators/tests/lag/lag_test.py
```diff
@@ -0,0 +1,95 @@
...
+            duration=2,
+            event=event,
+        )
+
+        lag_implementation = lag.LagNumpyImplementation(operator)
+        operator_output = lag_implementation(event=numpy_input_event)
+
+        self.assertEqual(
+            True,
+            numpy_output_event == operator_output["event"],
```
getchebarne, line 90: aren't we missing an `.all()` here? i.e. `numpy_output_event == operator_output["event"]` -> `(numpy_output_event == operator_output["event"]).all()`

temporian/implementation/numpy/operators/lag.py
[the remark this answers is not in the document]
diegomarvid: had that way, changed it to empty dict to avoid copying data. It was addressed due to a @achoum 's comment

temporian/core/operators/lag.py
diegomarvid: I still have the question of why we define each feature name twice. We define them in core and then in the implementation. Any comments on this? @getchebarne @ianspektor

temporian/implementation/numpy/operators/tests/lag/lag_test.py
diegomarvid: The `equals` is defined in the class NumpyEvent. So in this case we are comparing two NumpyEvents. The expected and the actual. Don't know why we should check them in list fashion if we are only comparing two objects.

temporian/core/data/duration.py
```diff
@@ -36,3 +36,74 @@ def hours(value: float) -> Duration:
...
+# Could be 29, 30 or 31 days
+def months(value: float) -> Duration:
+    return value * 60 * 60 * 24 * 30
+
+
+def years(value: float) -> Duration:
+    return value * 60 * 60 * 24 * 365
+
+
+def duration_abbreviation(duration: Duration, cutoff: str = None) -> str:
```
achoum, line 54: Add check that cutoff is a valid value.

temporian/core/data/duration.py
```diff
@@ -36,3 +36,74 @@ def hours(value: float) -> Duration:
 
 def days(value: float) -> Duration:
     return value * 60 * 60 * 24
+
+
+def weeks(value: float) -> Duration:
+    return value * 60 * 60 * 24 * 7
+
+
+# Could be 29, 30 or 31 days
```
achoum, line 45: Since a month duration is not accurate, I would not expose such method.
Same for years.

temporian/core/data/duration.py
```diff
@@ -36,3 +36,74 @@ def hours(value: float) -> Duration:
...
+        duration_str += f"{int(duration / minutes(1))}min"
+        if cutoff == "minute":
+            return duration_str
+        duration = duration % minutes(1)
+
+    if duration >= seconds(1):
+        duration_str += f"{int(duration / seconds(1))}s"
+        duration = duration % seconds(1)
+
+    return duration_str
```
achoum, line 109: This is missing sub seconds.

temporian/core/data/duration.py
```diff
@@ -36,3 +36,74 @@ def hours(value: float) -> Duration:
...
+
+    duration_str = ""
+
+    if duration >= years(1):
+        duration_str += f"{int(duration / years(1))}y"
+        if cutoff == "year":
+            return duration_str
+        duration = duration % years(1)
+
+    if duration >= months(1):
```
achoum, line 75: I wonder if this block make sense.

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,105 @@
...
+            key="LAG",
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="duration",
+                    type=pb.OperatorDef.Attribute.Type.FLOAT_64,
+                    is_optional=False,
+                ),
+            ],
+            inputs=[
+                pb.OperatorDef.Input(key="event"),
```
achoum, line 79: Remove comma

achoum (commented):
Nice. Left a few comments

temporian/core/data/duration.py
diegomarvid: What block? only months?

temporian/core/data/duration.py
diegomarvid: Do you think it's necessary? Would any user want to use sub-seconds in sma or lag in order to develop it? Do we want to support sub-seconds in those scenarios?

achoum:
LGTM

temporian/core/data/duration.py
```diff
@@ -27,6 +27,10 @@
 Duration = float
 
 
+def subseconds(value: float) -> Duration:
```
ianspektor, line 30: `milliseconds`?

temporian/core/data/duration.py
```diff
@@ -103,3 +111,61 @@ def is_a_date(value: any) -> bool:
         if value is date_type:
             return True
     return False
+
+
+def duration_abbreviation(
+    duration: Duration, cutoff: str = "subseconds"
```
ianspektor, line 117: Would be nice for this `cutoff` to be an enum if it only has a set of possible values

temporian/core/data/duration.py
```diff
@@ -103,3 +111,61 @@ def is_a_date(value: any) -> bool:
...
+        duration: Duration in seconds.
+        cutoff: Cutoff for the abbreviation. For example, if cutoff is "day", the
+        smallest unit will be days. Possible options are "week",
+        "day", "hour" and "minute", "seconds" and "subseconds". Default is
+        "subseconds".
+
+    Returns:
+        Abbreviation for the duration.
+    """
+    # check cutt off is valid
```
ianspektor, line 131:
```suggestion
    # check cutoff is valid

```

temporian/core/data/duration.py
```diff
@@ -103,3 +111,61 @@ def is_a_date(value: any) -> bool:
...
+    duration: Duration, cutoff: str = "subseconds"
+) -> str:
+    """Returns the abbreviation for a duration.
+
+    Args:
+        duration: Duration in seconds.
+        cutoff: Cutoff for the abbreviation. For example, if cutoff is "day", the
+        smallest unit will be days. Possible options are "week",
+        "day", "hour" and "minute", "seconds" and "subseconds". Default is
+        "subseconds".
```
ianspektor, line 126: missing tabs in the second and third line

temporian/core/data/duration.py
ianspektor: I see subseconds was chosen because of Mathieu's comment below, but milliseconds seems like the correct word for it. Be careful to change it where it appears below too 👀

temporian/core/operators/lag.py
```diff
@@ -0,0 +1,103 @@
...
+            outputs=[pb.OperatorDef.Output(key="event")],
+        )
+
+
+operator_lib.register_operator(LagOperator)
+
+
+def lag(event: Event, duration: Duration) -> Event:
+    if duration < 0:
+        raise ValueError("duration must be non-negative")
```
ianspektor, line 88: is duration = 0 something we want to allow? I would make both lag and leak exclude 0, and change message to `duration must be positive`

temporian/implementation/numpy/operators/lag.py
```diff
@@ -0,0 +1,35 @@
...
+
+        duration_str = duration_abbreviation(duration)
+
+        for index, timestamps in event.sampling.data.items():
+            new_sampling.data[index] = timestamps + duration
+            output_event.data[index] = []
+            output_data = output_event.data[index]
+            for feature in event.data[index]:
+                new_feature = NumpyFeature(
+                    data=feature.data.copy(),
```
ianspektor, line 30: why do we need a copy here?

temporian/implementation/numpy/operators/tests/lag_test.py
```diff
@@ -0,0 +1,94 @@
...
+from temporian.implementation.numpy.data.event import NumpyEvent
+from temporian.implementation.numpy.data.event import NumpyFeature
+from temporian.implementation.numpy.data.sampling import NumpySampling
+from temporian.implementation.numpy.operators import lag
+
+
+class LagOperatorTest(absltest.TestCase):
+    """Lag operator test."""
+
+    def test_correct_lag(self) -> None:
```
ianspektor, line 32: A test for `leak` would be nice too (even knowing they're testing the same implementation)

temporian/core/data/duration.py
```diff
@@ -103,3 +111,61 @@ def is_a_date(value: any) -> bool:
         if value is date_type:
             return True
     return False
+
+
+def duration_abbreviation(
```
ianspektor, line 116: Would be good to have a minimal test directly ensuring this works as expected

diegomarvid: We test this implicitly in lag due to the name of the new feature

temporian/implementation/numpy/operators/assign.py
```diff
@@ -58,11 +63,11 @@ def __call__(
...
-        output = NumpyEvent(
-            data=left_event.data.copy(), sampling=left_event.sampling
-        )
+        output = NumpyEvent(data={}, sampling=left_event.sampling)
 
-        for index in left_event.data.keys():
+        for index, left_features in left_event.data.items():
+            # Copy the features of left to the output event
+            output.data[index] = [feature for feature in left_features]
             output_data = output.data[index]
```
ianspektor, line 71: this assignment was a bit hard to understand - what do you think about defining `data = {}` before the for loop without creating the `output` event, and then creating the event when `data` is already created after the for loop?

diegomarvid: sounds pretty

ianspektor (approved):

PR #23, opened 2023-02-17
DataFrame <-> NumpyEvent

diegomarvid:
- [x] DataFrame To Numpy Event
- [x] Numpy Event To DataFrame
- [x] Tests
- [x] Ensure it works with index with only timestamps

### DIFF
--- a/temporian/core/data/duration.py
+++ b/temporian/core/data/duration.py
@@ -17,6 +17,11 @@
 Timestamples and durations are expressed with a double (noted float) in python.
 By convension, all calendar functions represent dates as Unix epoch in UTC.
 """
+from typing import Union
+
+import datetime
+import time
+import numpy as np
 
 # Unit for durations
 Duration = float
@@ -36,3 +41,65 @@ def hours(value: float) -> Duration:
 
 def days(value: float) -> Duration:
     return value * 60 * 60 * 24
+
+
+supported_date_types = (np.datetime64, time.struct_time, datetime.datetime)
+
+
+def convert_date_to_duration(
+    date: Union[np.datetime64, time.struct_time, datetime.datetime]
+) -> Duration:
+    """Convert date to duration epoch UTC
+
+    Args:
+        date (Union[np.datetime64, time.struct_time, datetime.datetime]):
+            Date to convert
+
+    Returns:
+        int: Duration epoch UTC
+
+    Raises:
+        TypeError: Unsupported type. Supported types are:
+        - np.datetime64
+        - time.struct_time
+        - datetime.datetime
+    """
+    if isinstance(date, np.datetime64):
+        return convert_numpy_datetime64_to_duration(date)
+    if isinstance(date, time.struct_time):
+        return convert_struct_time_to_duration(date)
+    if isinstance(date, datetime.datetime):
+        return convert_datetime_to_duration(date)
+
+    raise TypeError(f"Unsupported type: {type(date)}")
+
+
+def convert_numpy_datetime64_to_duration(date: np.datetime64) -> Duration:
+    """Convert numpy datetime64 to duration epoch UTC"""
+    return date.astype("datetime64[s]").astype("float64")
+
+
+def convert_struct_time_to_duration(date: time.struct_time) -> Duration:
+    """Convert struct_time to duration epoch UTC"""
+    return float(time.mktime(date) - time.timezone)
+
+
+def convert_datetime_to_duration(date: datetime.datetime) -> Duration:
+    """Convert datetime to duration epoch UTC"""
+    date = date.replace(tzinfo=datetime.timezone.utc)
+    return float(date.timestamp())
+
+
+def is_a_date(value: any) -> bool:
+    """Check if the value is a supported date type.
+
+    Args:
+        value (any): Value to check
+
+    Returns:
+        bool: True if the value is a date, False otherwise
+    """
+    for date_type in supported_date_types:
+        if value is date_type:
+            return True
+    return False
--- a/temporian/core/data/duration.py
+++ b/temporian/core/data/duration.py
@@ -16,11 +16,11 @@
 
 Timestamples and durations are expressed with a double (noted float) in python.
 By convension, all calendar functions represent dates as Unix epoch in UTC.
+This datatype is equivalent to a double in C.
 """
 from typing import Union
 
 import datetime
-import time
 import numpy as np
 
 # Unit for durations
@@ -43,16 +43,16 @@ def days(value: float) -> Duration:
     return value * 60 * 60 * 24
 
 
-supported_date_types = (np.datetime64, time.struct_time, datetime.datetime)
+supported_date_types = (np.datetime64, datetime.datetime)
 
 
 def convert_date_to_duration(
-    date: Union[np.datetime64, time.struct_time, datetime.datetime]
+    date: Union[np.datetime64, datetime.datetime]
 ) -> Duration:
     """Convert date to duration epoch UTC
 
     Args:
-        date (Union[np.datetime64, time.struct_time, datetime.datetime]):
+        date (Union[np.datetime64, datetime.datetime]):
             Date to convert
 
     Returns:
@@ -61,13 +61,10 @@ def convert_date_to_duration(
     Raises:
         TypeError: Unsupported type. Supported types are:
         - np.datetime64
-        - time.struct_time
         - datetime.datetime
     """
     if isinstance(date, np.datetime64):
         return convert_numpy_datetime64_to_duration(date)
-    if isinstance(date, time.struct_time):
-        return convert_struct_time_to_duration(date)
     if isinstance(date, datetime.datetime):
         return convert_datetime_to_duration(date)
 
@@ -79,15 +76,9 @@ def convert_numpy_datetime64_to_duration(date: np.datetime64) -> Duration:
     return date.astype("datetime64[s]").astype("float64")
 
 
-def convert_struct_time_to_duration(date: time.struct_time) -> Duration:
-    """Convert struct_time to duration epoch UTC"""
-    return float(time.mktime(date) - time.timezone)
-
-
 def convert_datetime_to_duration(date: datetime.datetime) -> Duration:
     """Convert datetime to duration epoch UTC"""
-    date = date.replace(tzinfo=datetime.timezone.utc)
-    return float(date.timestamp())
+    return date.replace(tzinfo=datetime.timezone.utc).timestamp()
 
 
 def is_a_date(value: any) -> bool:
--- a/temporian/core/data/duration.py
+++ b/temporian/core/data/duration.py
@@ -43,9 +43,6 @@ def days(value: float) -> Duration:
     return value * 60 * 60 * 24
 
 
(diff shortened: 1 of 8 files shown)


temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +103,105 @@ def schema(self) -> Event:
...
+        """
+        df_index = self.sampling.names + [timestamp_index_name]
+        df_features = self.feature_names
+        columns = df_index + df_features
+
+        df = pd.DataFrame(data=[], columns=columns).set_index(df_index)
+
+        for index, features in self.data.items():
+            timestamps_index = self.sampling.data[index]
+            for i, timestamp in enumerate(timestamps_index):
```
getchebarne, line 189: I think we should be able to circumvent this `for` loop. Pandas allows for multiple multi-index assignments. Should be more efficient.

temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +103,105 @@ def schema(self) -> Event:
...
+                # add timestamp to index
+                new_index = index + (timestamp,)
+                df.loc[new_index, df_features] = [
+                    feature.data[i] for feature in features
+                ]
+
+        # Convert to original dtypes, can be more efficient
+        first_index = self.first_index
+        first_features = self.data[first_index]
```
getchebarne, line 198: how about a `.head` method?

temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +103,105 @@ def schema(self) -> Event:
...
+        df = df.astype(
+            {feature.name: type(feature.data[0]) for feature in first_features}
```
getchebarne, line 200: Is there any benefit compared to using `feature.data.dtype`?

temporian/test/event_conversion_test.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        expected_df = pd.DataFrame(
+            [
+                [666964, 1.0, 740.0],
+                [666964, 2.0, 508.0],
+                [574016, 3.0, 573.0],
+            ],
+            columns=["product_id", "timestamp", "costs"],
+        ).set_index(["product_id", "timestamp"])
```
getchebarne, line 87: we shouldn't use floats as indexes, we should parse to datetimes first.

getchebarne (changes requested):

temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +103,105 @@ def schema(self) -> Event:
             sampling=self.sampling.names,
         )
 
+    @staticmethod
+    def from_dataframe(
```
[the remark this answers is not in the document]
diegomarvid: Yeah, just implemented this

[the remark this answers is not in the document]
diegomarvid: Looks better btw!

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +99,110 @@ def schema(self) -> Event:
...
+                NumpyFeature(feature, df[feature].to_numpy())
+                for feature in feature_columns
+            ]
+
+        numpy_sampling = NumpySampling(index=index_names, data=sampling)
+
+        return NumpyEvent(data=data, sampling=numpy_sampling)
+
+    def to_dataframe(
+        self, timestamp_index_name: str = "timestamp"
```
achoum, line 166: This is a very long parameter name.
What about "timestamp_column"?

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +99,110 @@ def schema(self) -> Event:
...
+    ) -> pd.DataFrame:
+        """Function to convert a NumpyEvent to a pandas DataFrame
+
+        Args:
+            timestamp_index_name: Name for timestamp index. Defaults to "timestamp".
+
+        Returns:
+            pd.DataFrame: DataFrame created from NumpyEvent
+        """
+        df_index = self.sampling.index + [timestamp_index_name]
```
achoum, line 176: Can you make sure the timestamp name does not overlap with an index name?
In addition, can you also make sure that no index name overlap with a feature name.

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +99,110 @@ def schema(self) -> Event:
...
+        df_features = self.feature_names
+        columns = df_index + df_features
+
+        df = pd.DataFrame(data=[], columns=columns).set_index(df_index)
```
achoum, line 180: I would make the output dataframe have exactly the same format as the expected format in from_dataframe. Like for from_dataframe, I think having the dataframe be without index might be better.

temporian/test/event_conversion_test.py
```diff
@@ -0,0 +1,183 @@
...
+                        data=np.array([740.0, 508.0, 573.0]), name="costs"
+                    ),
+                ],
+            },
+            sampling=numpy_sampling,
+        )
+
+        expected_df = pd.DataFrame(
+            [
+                [666964, 1.0, 740.0],
```
achoum, line 166: Idea of things to test:
- when the timestamp is a Date.
- when one of the feature is a string.
- when one of the feature contains missing values .

achoum (commented):
This is great.
Left a few comments.

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +99,110 @@ def schema(self) -> Event:
             sampling=self.sampling.names,
         )
 
+    @staticmethod
+    def from_dataframe(
+        df: pd.DataFrame,
+        index_names: List[str],
+        timestamp_name: str = "timestamp",
```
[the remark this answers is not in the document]
diegomarvid: I think every pandas dtype should be allowed right? I don't include any constraints in the code.  What do you think about?

temporian/test/event_conversion_test.py
diegomarvid: Should we support Date format in timestamp even if it's not in `DTYPE_MAPPING`?

achoum:
LGTM once what we discussed (e.g. timestamps is improvented).

achoum (approved):

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +101,166 @@ def schema(self) -> Event:
...
+                NumpyFeature(feature, df[feature].to_numpy())
+                for feature in feature_columns
+            ]
+
+        numpy_sampling = NumpySampling(index=index_names, data=sampling)
+
+        return NumpyEvent(data=data, sampling=numpy_sampling)
+
+    def to_dataframe(self) -> pd.DataFrame:
+        """Function to convert a NumpyEvent to a pandas DataFrame
```
ianspektor, line 217: nit: missing dots at the end of most sentences in docstrings

temporian/test/event_conversion_test.py
```diff
@@ -0,0 +1,316 @@
...
+            sampling=numpy_sampling,
+        )
+
+        numpy_event = NumpyEvent.from_dataframe(
+            df, index_names=["product_id"], timestamp_column="timestamp"
+        )
+
+        # validate
+        self.assertEqual(
+            True,
```
ianspektor, line 59: `self.assertTrue`

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +105,165 @@ def schema(self) -> Event:
...
+        timestamp_column: str = "timestamp",
+    ) -> "NumpyEvent":
+        """Convert a pandas DataFrame to a NumpyEvent. Supported
+        dtypes are: np.float64, np.float32, np.int64, np.int32, np.datetime64
+        (for timestamp column).
+
+        Args:
+            df: DataFrame to convert to NumpyEvent.
+            index_names: names of the DataFrame columns to be used as index for the event.
+            timestamp_column: Name for timestamp index. Defaults to "timestamp".
```
achoum, line 121: Column containing timestamps. Supported types: {...}. Timestamps of the types [...] are converted implicitly to in UTC epoch.

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +105,165 @@ def schema(self) -> Event:
             sampling=self.sampling.names,
         )
 
+    @staticmethod
+    def from_dataframe(
+        df: pd.DataFrame,
+        index_names: List[str],
```
achoum, line 111: Having the indexing logic is great. However, would it make sense to move this logic to the "addIndex" (or something similar) operator instead?

To speed-up the work, you can submit this PR as it is, and maybe make the move in another PR.

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +105,165 @@ def schema(self) -> Event:
...
+
+        # check timestamp_column is not on index_names
+        if timestamp_column in index_names:
+            raise ValueError(
+                f"Timestamp column {timestamp_column} cannot be on index_names"
+            )
+
+        # check column dtypes, every dtype should be a key of DTYPE_MAPPING
+        for column in df.columns:
+            # if it's a date, convert it to duration
```
achoum, line 164: Nit: s/duration/UTC epoch

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +105,165 @@ def schema(self) -> Event:
...
+            if is_a_date(df[column].dtype.type):
```
achoum, line 165: I would only do an automated conversion for the timestamp column.

temporian/implementation/numpy/data/tests/test_df_to_event.py
```diff
@@ -0,0 +1,189 @@
...
+            df, index_names=["product_id"], timestamp_column="timestamp"
+        )
+
+        # validate
+        self.assertTrue(numpy_event == expected_numpy_event)
+
+    def test_df_to_numpy_event_datetime_index(self) -> None:
+        df = pd.DataFrame(
+            [
+                [666964, np.datetime64("2022-01-01"), 740.0],
```
achoum, line 115: Can you test the other supported datetime object, and ensure that the timestamp is as expected?

achoum (approved):

temporian/implementation/numpy/data/event.py
diegomarvid: And what would happen for other columns if the DTYPE is a date? Should we raise an error?

temporian/core/data/duration.py
```diff
@@ -36,3 +41,65 @@ def hours(value: float) -> Duration:
...
+
+def convert_struct_time_to_duration(date: time.struct_time) -> Duration:
+    """Convert struct_time to duration epoch UTC"""
+    return float(time.mktime(date) - time.timezone)
+
+
+def convert_datetime_to_duration(date: datetime.datetime) -> Duration:
+    """Convert datetime to duration epoch UTC"""
+    date = date.replace(tzinfo=datetime.timezone.utc)
+    return float(date.timestamp())
```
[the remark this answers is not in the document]
diegomarvid: If it was a float, it would return false in the `is_a_date` function and therefore no conversion will be needed.  This only converts datetime objects.

PR #24, opened 2023-02-21
Arithmetic operator

diegomarvid:
As every arithmetic operation requires to loop in the same way as the others. I implemented a common Arithmetic Operator that receives an Enum for the specific Arithmetic Operation.

Basic Arithmetic Operations supported at the moment are:

- [ ] Addition
- [ ] Substraction
- [ ] Multiplication
- [ ] Division


temporian/core/operators/arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @staticmethod
+    def prefix(operation: "ArithmeticOperation") -> str:
+        if operation == ArithmeticOperation.ADDITION:
+            return "sum"
+        if operation == ArithmeticOperation.SUBTRACTION:
+            return "diff"
+        if operation == ArithmeticOperation.MULTIPLICATION:
+            return "prod"
+        if operation == ArithmeticOperation.DIVISION:
```
getchebarne, line 40: nit: make it `"div"`?

temporian/core/operators/arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+    SUBTRACTION = "SUBTRACTION"
+    MULTIPLICATION = "MULTIPLICATION"
+    DIVISION = "DIVISION"
+
+    @staticmethod
+    def prefix(operation: "ArithmeticOperation") -> str:
+        if operation == ArithmeticOperation.ADDITION:
+            return "sum"
+        if operation == ArithmeticOperation.SUBTRACTION:
+            return "diff"
```
getchebarne, line 37: nit: make it `"sub"`?

temporian/core/operators/arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # outputs
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"{prefix}_{event_1_f.name()}_{event_2_f.name()}",
+                dtype=event_1_f.dtype(),
+                sampling=sampling,
+                creator=self,
+            )
+            for event_1_f, event_2_f in zip(
```
getchebarne, line 95: nit: can we rename these to `feature_1` and `feature_2`? I think it makes it clearer that these are features and not events

temporian/core/operators/arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+                pb.OperatorDef.Input(key="event_1"),
+                pb.OperatorDef.Input(key="event_2"),
+            ],
+            outputs=[pb.OperatorDef.Output(key="event")],
+        )
+
+
+operator_lib.register_operator(ArithmeticOperator)
+
+
```
getchebarne, line 136: nit: rename to `sum`, `substract`, `multiply` `divide`?

temporian/core/operators/arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+) -> Event:
+    return ArithmeticOperator(
+        event_1=event_1,
+        event_2=event_2,
+        operation=ArithmeticOperation.MULTIPLICATION,
+        resolution=resolution,
+    ).outputs()["event"]
+
+
+def division(
```
getchebarne, line 176: I think we should make explicit which event is the numerator and which is the denominator.

getchebarne (commented):
Great job 💯  how about overwriting the `Event`'s class `__add__`,`__sub__`, `__truediv__` and `__mul__` magic methods so that the user can use `+`, `-`, `\` and `*` with `Event`s?

temporian/core/operators/arithmetic.py
diegomarvid: 💯

temporian/core/operators/arithmetic.py
diegomarvid: 💯

temporian/core/operators/arithmetic.py
diegomarvid: Sub could be more misleading than diff, it could mean more things. However, the short version of subtraction (sub) is also a good option. What do you all think? @ianspektor @achoum

temporian/implementation/numpy/data/event.py
```diff
@@ -89,6 +99,110 @@ def schema(self) -> Event:
             sampling=self.sampling.names,
         )
 
+    @staticmethod
+    def from_dataframe(
```
achoum, line 103: Reviewed in other PR.

temporian/implementation/numpy/operators/arithmetic.py
```diff
@@ -0,0 +1,70 @@
...
+                "PER_FEATURE_NAME is not implemented yet."
+            )
+
+        prefix = ArithmeticOperation.prefix(operation)
+
+        for event_index, event_index_array in event_1.data.items():
+            output.data[event_index] = []
+
+            for i in range(len(event_index_array)):
+                event_1_feature = event_1.data[event_index][i]
```
achoum, line 49: Don't execute this "[event_index]" query for each iteration.

temporian/implementation/numpy/operators/arithmetic.py
```diff
@@ -0,0 +1,70 @@
...
+                event_2_feature = event_2.data[event_index][i]
+
+                data = None
+
+                if operation == ArithmeticOperation.ADDITION:
+                    data = event_1_feature.data + event_2_feature.data
+                elif operation == ArithmeticOperation.SUBTRACTION:
+                    data = event_1_feature.data - event_2_feature.data
+                elif operation == ArithmeticOperation.MULTIPLICATION:
+                    data = event_1_feature.data * event_2_feature.data
```
achoum, line 59: Fail if the operation is unknown

temporian/implementation/numpy/operators/arithmetic.py
```diff
@@ -0,0 +1,70 @@
...
+            raise NotImplementedError(
+                "PER_FEATURE_NAME is not implemented yet."
+            )
+
+        prefix = ArithmeticOperation.prefix(operation)
+
+        for event_index, event_index_array in event_1.data.items():
+            output.data[event_index] = []
+
+            for i in range(len(event_index_array)):
```
achoum, line 48: What is the semantic of "i"? Could there be a better name?

temporian/core/operators/arithmetic.py
```diff
@@ -98,14 +131,56 @@ def build_op_definition(cls) -> pb.OperatorDef:
         )
 
 
-operator_lib.register_operator(SumOperator)
+operator_lib.register_operator(ArithmeticOperator)
 
 
 def sum(
```
achoum, line 137: Since those functions are part of the API, they should be fully documented.

temporian/core/operators/arithmetic.py
```diff
@@ -21,21 +23,38 @@
...
-    """Sum operator."""
+class ArithmeticOperator(Operator):
+    """Arithmetic operator."""
 
     def __init__(
         self,
         event_1: Event,
         event_2: Event,
+        operation: ArithmeticOperation,
         resolution: Resolution = Resolution.PER_FEATURE_IDX,
```
achoum, line 58: No need to double define the default values. Currently, the default values should be only defined in the API.

achoum (commented):
Look great.

achoum:
LGTM once what we discussed (e.g. prefix names is done).

achoum (approved):

temporian/implementation/numpy/operators/core_mapping.py
```diff
@@ -1,9 +1,13 @@
...
+    AssignNumpyImplementation,
+)
 from temporian.implementation.numpy.operators.select import NumpySelectOperator
-from temporian.implementation.numpy.operators.sum import NumpySumOperator
+from temporian.implementation.numpy.operators.arithmetic import (
+    ArithmeticNumpyImplementation,
+)
 
 OPERATOR_IMPLEMENTATIONS = {
     "SELECT": NumpySelectOperator,
```
ianspektor, line 10: We should rename this one to `SelectNumpyImplementation` to keep consistency with the others

temporian/implementation/numpy/operators/tests/arithmetic_test.py
```diff
@@ -0,0 +1,210 @@
...
+            resolution=Resolution.PER_FEATURE_IDX,
+        )
+
+        sum_implementation = arithmetic.ArithmeticNumpyImplementation(operator)
+
+        operator_output = sum_implementation(
+            event_1=self.numpy_event_1, event_2=self.numpy_event_2
+        )
+
+        self.assertEqual(
```
ianspektor, line 104: `self.assertTrue` everywhere

temporian/test/prototype_test_numpy.py
```diff
@@ -20,163 +21,101 @@
...
     def setUp(self) -> None:
-        # index_names
-        index_names = ["store_id", "product_id"]
+        # store ids
+        TRYOLABS_SHOP = 42
+        GOOGLE_SHOP = 101
+        # product ids
+        MATE_ID = 1
+        BOOK_ID = 2
+        PIXEL_ID = 3
```
ianspektor, line 35: ❤️

ianspektor (approved):

temporian/implementation/numpy/operators/core_mapping.py
diegomarvid: I think I should do this on another PR and change `SelectNumpyImpelemntation` accordingly. It could be made on #28 or a new PR

PR #26, opened 2023-03-12
Finish Numpy SMA backend

achoum:
Details:

- Implement logic for `sampling` argument for SMA operator.
- Add support for missing values (represented as NaN) for SMA operator.
- Improve coverage of unit tests for SMA operators and API.

Also:
- Modify the evaluation logic so that operator implementation classes expect the corresponding operator object as the only constructor argument. Prior to this change, operator implementation classes were expecting the unfolded serialized attributes of the operator class.
- The `query` argument of the `evaluate` method can be an event, list of events, or dictionary of string-to-events. The returned values of the `evaluate` method is of the same type as the  `query` argument. For example, if the query is a dictionary of events, the returns is a dictionary (with the same keys) of event values.
- The `evaluate` method cannot return anymore a dictionary  indexed by event.


temporian/core/backends.py
```diff
@@ -22,6 +22,7 @@ def raise_(exception: Exception):
     raise exception
 
 
+# TODO: Use a registration mechanism instead of listing the backends.
```
ianspektor, line 25: 💯

temporian/core/evaluator.py
```diff
@@ -26,26 +26,50 @@
...
+        will be a dictionary of event results. If "event" is a list of events,
+        the returned value will be a list of event values with the same order.
+    """
+
+    # Normalize query into a list
     normalized_query: List[Event] = {}
+
+    Q_TYPE_DICT = "DICT"
+    Q_TYPE_LIST = "LIST"
+    Q_TYPE_VALUE = "VALUE"
```
ianspektor, line 57: Wdyt about making this an enum and defining it outside the function?

temporian/core/evaluator.py
```diff
@@ -26,26 +26,50 @@
...
+
     if isinstance(query, Event):
+        # The query is a single value
         normalized_query = [query]
-
-    elif isinstance(query, set):
-        normalized_query = list(query)
+        query_type = Q_TYPE_VALUE
```
ianspektor, line 62: Shouldn't `Q_TYPE_VALUE` be for the first case (`query` is an `Event`) and this be `Q_TYPE_SET`?

temporian/core/evaluator.py
```diff
@@ -73,10 +100,25 @@ def evaluate(
...
         for input_event, input_event_spec in input_data.items()
     }
-    # evaluate schedule
+
+    # Evaluate schedule
     outputs = evaluate_schedule_fn(materialized_input_data, schedule)
 
-    return {event: outputs[event] for event in normalized_query}
+    # Convert the result "outputs" into the same format as the query.
+    if query_type == Q_TYPE_DICT:
```
ianspektor, line 108: The definition of the `Q_TYPE` enum/constants seems a bit overkill to me if the only purpose is to be able to do this comparison instead of doing using `isinstance(query, X)` again

temporian/core/evaluator.py
```diff
@@ -73,10 +100,25 @@ def evaluate(
...
+        }
+
+    elif query_type == Q_TYPE_LIST:
+        return [outputs[k] for k in query]
+
+    elif query_type == Q_TYPE_VALUE:
+        return outputs[query]
+
+    else:
+        raise ValueError("Internal error")  # DO NOT SUBMIT: Correct exception
```
ianspektor, line 121: Leaving comment for this not to go under the radar

temporian/implementation/numpy/operators/assign.py
```diff
@@ -19,9 +19,16 @@
 from temporian.implementation.numpy.data.event import NumpyEvent
 from temporian.implementation.numpy.data.event import NumpyFeature
 from temporian.implementation.numpy.data.sampling import NumpySampling
+from temporian.core.operators.assign import AssignOperator
 
 
+# TODO: Rename to AssignImplementation
 class NumpyAssignOperator:
+    # TODO: Remove the optionallity on "op".
+    def __init__(self, op: AssignOperator = None) -> None:
```
ianspektor, line 28: Why are we making it optional now? 👀

temporian/implementation/numpy/operators/core_mapping.py
```diff
@@ -1,9 +1,19 @@
...
+)
 
+# TODO: Use a registration mechanism instead.
+# TODO: Make sure the implementation key is defined only one (in the operator
+# definition).
 OPERATOR_IMPLEMENTATIONS = {
     "SELECT": NumpySelectOperator,
     "SUM": NumpySumOperator,
     "ASSIGN": NumpyAssignOperator,
+    SmaDef.implementation_key(): SmaImp,  # Simple moving average
```
ianspektor, line 18: If we're making this change I would change all of them together (or since not in scope of this PR, open an issue addressing it)

temporian/implementation/numpy/operators/tests/simple_moving_average.py
ianspektor, line 1: rename to `simple_moving_average_test.py` for consistency with existing ones

we should also rename the folder from `tests` to `test` here for consistency too

temporian/test/prototype_test.py
```diff
@@ -75,7 +75,7 @@ def setUp(self) -> None:
             ],
         ).set_index(["product_id", "timestamp"])
 
-    def test_prototoype(self) -> None:
+    def disabled_test_prototoype(self) -> None:
```
ianspektor, line 78: why?

ianspektor (approved):

ianspektor (commented):
SMA implementation seems solid 👍🏼 nice work!

Take a careful look at the merge conflicts, some stuff implemented here has already been implemented into core in some of diego's PRs (e.g. operator implementations receiving core operator instead of unrolled arguments)

temporian/core/evaluator.py
achoum: Good idea.
*Done

achoum:
Thanks for the thorough review

PR #28, opened 2023-03-13
Select tests

diegomarvid:
Add Numpy Implementation test for Select operator

### DIFF
--- a/temporian/implementation/numpy/operators/tests/select_test.py
+++ b/temporian/implementation/numpy/operators/tests/select_test.py
@@ -1,0 +1,100 @@
+# Copyright 2021 Google LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


temporian/implementation/numpy/operators/tests/select_test.py
```diff
@@ -0,0 +1,100 @@
...
+    def test_select_one_feature(self) -> None:
+        """Test correct select operator for one feature selection."""
+
+        selected_feature = "sales"
+
+        # drop all columns in features except selected_feature
+        new_df = self.df.copy()
+        new_df = new_df.drop(
+            columns=[col for col in self.features if col != selected_feature]
+        )
```
ianspektor, line 59: Having logic like this in a test should be avoided - you would need a test that tests that your test is doing what you think it's doing :) although more verbose, it is preferrable to copy paste the selected feature for the expected event

temporian/implementation/numpy/operators/tests/select_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+        selected_features = ["sales", "costs"]
+
+        # drop all columns in features except selected_feature
+        new_df = self.df.copy()
+        new_df = new_df.drop(
+            columns=[
+                col for col in self.features if col not in selected_features
+            ]
+        )
```
ianspektor, line 84: same

ianspektor (approved):

PR #41, opened 2023-03-18
Propagate operator

achoum:
The propagate operator allows to duplicate features over an index.


temporian/core/operators/select.py
```diff
@@ -28,8 +28,15 @@ def __init__(self, event: Event, feature_names: Union[str, List[str]]):
         super().__init__()
 
         # store selected feature names
-        if isinstance(feature_names, str):
+        if isinstance(feature_names, list):
+            pass
+        elif isinstance(feature_names, str):
             feature_names = [feature_names]
+        else:
```
ianspektor, line 35: nit, subjective, this would make it cleaner, avoiding the if/pass clause:
```
if (isinstance(feature_names, str):
    feature_names = [feature_names]
    
if not isinstance(feature_names, list):
    raise ...
```

temporian/implementation/numpy/operators/propagate.py
```diff
@@ -0,0 +1,72 @@
...
+import numpy as np
+import itertools
+
+from temporian.implementation.numpy.data.event import NumpyEvent, NumpyFeature
+from temporian.implementation.numpy.data.sampling import NumpySampling
+from temporian.core.operators.propagate import Propagate
+
+
+class PropagateNumpyImplementation:
+    """Numpy implementation for the propagat operator."""
```
ianspektor, line 15:
```suggestion
    """Numpy implementation for the propagate operator."""

```

temporian/implementation/numpy/operators/propagate.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, op: Propagate) -> None:
+        assert isinstance(op, Propagate)
+        self._op = op
```
ianspektor, line 19: `self.operator` to keep it consistent with existing ones? else if `_op` is better (which I agree on) I would change all of the existing ones to it

temporian/implementation/numpy/operators/propagate.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __call__(
+        self,
+        event: NumpyEvent,
+        add_event: NumpyEvent = None,
```
ianspektor, line 24: Why the `None` default here? line 35 would raise if None

temporian/implementation/numpy/operators/tests/propagate_test.py
```diff
@@ -0,0 +1,157 @@
...
+from temporian.core.data import event as event_lib
+from temporian.core.data import feature as feature_lib
+from temporian.core.data import dtype as dtype_lib
+
+
+class PropagateOperatorTest(absltest.TestCase):
+    def setUp(self):
+        pass
+
+    def test_base(self):
```
ianspektor, line 35: Take a look at how we declared core and numpy events in `calendar_day_of_month_test` using `NumpyEvent.from_dataframe()` and `.schema()`. It would be nice to try and keep tests that test similar stuff (in this case, a numpy implementation) in a similar format. We can discuss on what's the best format in our next meeting 🗣️

ianspektor: using .schema() specifically allows to skip the core definitions, which makes it much more concise

temporian/implementation/numpy/operators/tests/propagate_test.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+        # Expected output
+        expected_sampling = NumpySampling(
+            index=["x", "c", "d"],
+            data={
+                ("X1", 1, 1): np.array([0.1, 0.2, 0.3], dtype=np.float64),
+                ("X1", 1, 2): np.array([0.1, 0.2, 0.3], dtype=np.float64),
+                ("X1", 2, 1): np.array([0.1, 0.2, 0.3], dtype=np.float64),
+                ("X1", 2, 2): np.array([0.1, 0.2, 0.3], dtype=np.float64),
```
ianspektor, line 97: I _think_ this `(2, 2)` value shouldn't be here, since it isn't present in the original data, and we decided we wanted the first implementation at least to propagate on the existing pairs and not on the cross product?

temporian/implementation/numpy/operators/propagate.py
```diff
@@ -0,0 +1,72 @@
...
+            index_uniques = []
+            for idx, key in enumerate(added_index):
+                assert key == src_add_event_item[idx].name
+
+            index_uniques = [
+                np.unique(src_add_event_item[idx].data)
+                for idx in range(len(added_index))
+            ]
+
+            for extra_index in itertools.product(*index_uniques):
```
ianspektor, line 52: See comment below in the test - let's discuss tomorrow about the behavior we want when selecting multiple columns to propagate over. This is taking the product of the unique values in each column we propagate to, we might want to propagate to the tuples that actually appear together in those columns.

temporian/core/operators/propagate.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+operator_lib.register_operator(Propagate)
+
+
+def propagate(
+    event: Event,
+    add_index: Union[Event, str, List[str]],
+) -> Event:
+    """Extands index and propagates feature values.
```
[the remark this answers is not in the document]
achoum: *Done

temporian/core/operators/propagate.py
```diff
@@ -0,0 +1,160 @@
...
+    features "f1" and "f2".
+
+    Constraints:
+
+    - If `add_index` is an event, `event` and `add_index` have the same index
+      and same sampling.
+
+    Args:
+        event: The event to propagate.
+        add_index: The features to index over. If add_index` is a list of
```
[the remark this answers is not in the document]
achoum: *Done

temporian/core/operators/propagate.py
```diff
@@ -0,0 +1,160 @@
...
+          strings, those string should refer to existing features of `event`
```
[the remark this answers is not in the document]
achoum: *Done

temporian/core/operators/propagate.py
```diff
@@ -0,0 +1,160 @@
...
+from temporian.core.data.sampling import Sampling
+
+
+class Propagate(Operator):
+    """Propagate operator."""
+
+    def __init__(
+        self,
+        event: Event,
+        add_index: Event,
```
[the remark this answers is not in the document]
achoum: "to" is the right one.
*Rename argument to "to".

temporian/implementation/numpy/operators/propagate.py
achoum: *Done

temporian/implementation/numpy/operators/propagate.py
achoum: I don't have preference between _op or _operator. It seems the "operator" version is currently more common in the code, so I'll update to that. However, I see value in the "_": This is a private attribute.

*Change to _operator

temporian/implementation/numpy/operators/propagate.py
achoum: Nice catch. This was some outdated code.

*Remove None

temporian/implementation/numpy/operators/propagate.py
achoum: As from discussion.

*Remove product.

temporian/implementation/numpy/operators/tests/propagate_test.py
achoum: Good point. See next comment.

ianspektor (approved):
Thanks for all the updates!

PR #49, opened 2023-03-20
Improve `to_dataframe` efficiency 

diegomarvid:
The functionality to convert from a NumpyEvent to a DataFrame was working too slow.
On the MINI config of the colab, the time for conversion was around 1 minute. Improved it to 224ms. It probably can be further improved considering converting from DataFrame to NumpyEvent takes 22ms.

### DIFF
--- a/temporian/implementation/numpy/data/event.py
+++ b/temporian/implementation/numpy/data/event.py
@@ -241,16 +241,11 @@ class NumpyEvent:
 
         for index, features in self.data.items():
             timestamps = self.sampling.data[index]
-
-            for i, timestamp in enumerate(timestamps):
-                # add row to dictionary
-                row = (
-                    list(index)
-                    + [feature.data[i] for feature in features]
-                    + [timestamp]
-                )
-                for i, column_name in enumerate(columns):
-                    data[column_name].append(row[i])
+            data["timestamp"].extend(timestamps)
+            for feature in features:
+                data[feature.name].extend(feature.data)
+            for index_key in self.sampling.index:
+                data[index_key].extend(index * len(timestamps))
 
         # Converting dictionary to pandas DataFrame
         df = pd.DataFrame(data)


temporian/implementation/numpy/data/event.py
```diff
@@ -214,48 +214,29 @@ def to_dataframe(self) -> pd.DataFrame:
...
-        df = pd.DataFrame(data=[], columns=columns)
+        columns = self.sampling.index + self.feature_names + ["timestamp"]
+        for column_name in columns:
+            data[column_name] = []
 
-        # append every feature to the dataframe. without index
         for index, features in self.data.items():
             timestamps = self.sampling.data[index]
 
             for i, timestamp in enumerate(timestamps):
```
achoum, line 228: Would it be more efficient to inverse the two for loops (since the internal one currently has a dictionary query)?

Something along the lines of (the following python code was not tested).

```
for index, features in self.data.items():
	timesamples = self.sampling.data[index]
	data[timestamp_key].extends(timesamples )
	for feature in features:
		data[feature.name].extends(features.data)
	for index_key in index_keys:
		data[index_key].extends( index * len(timesamples))
```

achoum: It is not every day you can say you implemented a 250x speed-up :)

achoum (approved):

temporian/implementation/numpy/data/event.py
```diff
@@ -219,48 +218,24 @@ def to_dataframe(self) -> pd.DataFrame:
...
-            index_name: type(first_index[i])
-            for i, index_name in enumerate(self.sampling.index)
-        }
-
-        # get timestamp dtype
-        first_timestamp = self.sampling.data[first_index][0]
-        sampling_dtype = {"timestamp": first_timestamp.dtype}
-
-        df = df.astype({**features_dtypes, **index_dtypes, **sampling_dtype})
+            timesamples = self.sampling.data[index]
```
getchebarne, line 230: nit: `timesamples` -> `timestamps`

temporian/implementation/numpy/data/event.py
```diff
@@ -219,48 +218,24 @@ def to_dataframe(self) -> pd.DataFrame:
...
+            data["timestamp"].extend(timesamples)
+            for feature in features:
+                data[feature.name].extend(feature.data)
+            for index_key in self.sampling.index:
+                data[index_key].extend(index * len(timesamples))
+
+        # Converting dictionary to pandas DataFrame
+        df = pd.DataFrame(data)
```
getchebarne, line 238: Does this keep the original `NumpyEvent's` index? i.e. is the output `DataFrame` indexed by the same columns as the event?

getchebarne (approved):

temporian/implementation/numpy/data/event.py
diegomarvid: We talked about it and decided that it was best for the output DataFrame to not have any index

PR #50, opened 2023-03-20
Support String in `NumpyEvent.from_dataframe()`

diegomarvid:
Convert np.objects in DataFrame to np.strings

- [x] Raise exception when non string object is passed

### DIFF
--- a/temporian/implementation/numpy/data/tests/test_df_to_event.py
+++ b/temporian/implementation/numpy/data/tests/test_df_to_event.py
@@ -61,20 +61,44 @@ class DataFrameToEventTest(absltest.TestCase):
     def test_string_column(self) -> None:
         df = pd.DataFrame(
             [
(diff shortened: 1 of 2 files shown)


temporian/implementation/numpy/data/tests/test_df_to_event.py
```diff
@@ -61,21 +63,45 @@ def test_correct(self) -> None:
     def test_string_column(self) -> None:
         df = pd.DataFrame(
             [
-                [666964, 1.0, "740.0"],
-                [666964, 2.0, "508.0"],
-                [574016, 3.0, "573.0"],
+                [666964, 1.0, "740"],
+                [666964, 2.0, np.nan],
```
achoum, line 67: Following on today's discussion. Can we have a nan in a string array? If so, is that efficient?

achoum (approved):
Awesome.

PR #51, opened 2023-03-20
DropIndexOperator

getchebarne:
This PR implements the `drop_index` operator. It works similarly to pandas' `reset_index`method. In our case, it collapses the existing index into a single level containing all the data. The resulting event is sorted by timestamp, in increasing order. 

Depends on #80.


temporian/core/operators/reset_index.py
```diff
@@ -0,0 +1,78 @@
...
+                sampling=output_sampling,
+                creator=self,
+            ),
+        )
+        self.check()
+
+    @classmethod
+    def build_op_definition(cls) -> pb.OperatorDef:
+        return pb.OperatorDef(
+            key="SELECT",
```
diegomarvid, line 57: `RESET_INDEX` ?

temporian/core/operators/reset_index.py
```diff
@@ -0,0 +1,78 @@
...
+
+from temporian.core import operator_lib
+from temporian.core.data.event import Event
+from temporian.core.data.feature import Feature
+from temporian.core.data.sampling import Sampling
+from temporian.core.operators.base import Operator
+from temporian.proto import core_pb2 as pb
+
+
+class ResetIndexOperator(Operator):
```
achoum, line 25: As from today discussion. Could the reset index operator be implemented as an AddIndex and aRemoveIndex operators (or something equivalent), where the tp.reset_index would be implemented as a RemoveIndex operator with all the available index selected?

temporian/core/operators/reset_index.py
```diff
@@ -0,0 +1,78 @@
...
+    """ResetIndex operator."""
+
+    def __init__(self, event: Event):
```
achoum, line 28: Idea: Add option to drop the index data.

temporian/core/operators/reset_index.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__()
+
+        # input event
+        self.add_input("event", event)
+
+        # output features
+        output_features = [
+            Feature(name=feature_name)
```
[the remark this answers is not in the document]
getchebarne: Good catch. I left this unresolved, we can discuss it on tomorrow's meeting.

temporian/implementation/numpy/operators/reset_index.py
```diff
@@ -0,0 +1,79 @@
...
+            # 2. from already existing features (appended)
+            for i, src_feat in enumerate(event.data[src_idx_lvl]):
+                dst_features[i + src_idx_len].data[
+                    ptr : ptr + num_timestamps
+                ] = src_feat.data
+
+            ptr += num_timestamps
+
+        # finally, sort data according to timestamps (increasingly)
+        sorted_idxs = np.argsort(dst_sampling_data, kind="mergesort")
```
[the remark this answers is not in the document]
getchebarne: To my understanding, it's the most efficient algorithm to sort an array of already-sorted subarrays.

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,113 @@
...
+                pb.OperatorDef.Input(key="event"),
+            ],
+            outputs=[pb.OperatorDef.Output(key="event")],
+        )
+
+
+operator_lib.register_operator(DropIndexOperator)
+
+
+def drop_index(
```
diegomarvid, line 110: This function is part of the API. Would be great to have an extensive docstring for better documentation 😀

temporian/implementation/numpy/operators/reset_index.py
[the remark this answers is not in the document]
achoum: SG

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+from temporian.proto import core_pb2 as pb
+
+
+class DropIndexOperator(Operator):
+    """DropIndex operator."""
+
+    def __init__(
+        self,
+        event: Event,
+        labels: Optional[Union[List[str], str]] = None,
```
achoum, line 32: Nit: What about moving the logic related to support for different types for the "labels" argument in the "drop_index" function (instead of the DropIndexOperator class)?

Arguments with multiple supported types is great for simple APIs, but I think it makes the code harder to read and more prone to error when extending.

We can chat about that today.

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+        keep: bool = True,
```
achoum, line 33: Defining default values in two separate places (here and in the api function) could lead to some inconsistency (imagine the default value of one change).

What about only using default values in the api function?

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+        self.add_attribute("keep", keep)
+
+        # output features
+        output_features = self._generate_output_features(event, labels, keep)
+
+        # output sampling
+        output_sampling = Sampling(
+            index=[
+                index_name
+                for index_name in event.sampling().index()
```
achoum, line 54: Can you check that all the items in "labels" are present in the index?

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+        super().__init__()
+
+        # process labels
+        labels = self._process_labels(event, labels)
+
+        # input event
+        self.add_input("event", event)
+
+        # attributes
+        self.add_attribute("labels", labels)
```
achoum, line 44: add_attribute does not support None type. Is the None type necessary here?

Note: If the None type means "dropping all the index", I would have this logic in the api function (instead of the api class).

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+                sampling=output_sampling,
+                creator=self,
+            ),
+        )
+        self.check()
+
+    def _process_labels(
+        self, event: Event, labels: Optional[Union[List[str], str]]
+    ) -> List[str]:
+        if labels is None:
```
achoum, line 73: Move this logic in the api function.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,181 @@
...
+IndexMetadata = Dict[
+    str, Union[int, List[List[NumpyFeature]], List[np.ndarray]]
+]
+
+
+def _impl(
+    event: NumpyEvent, drop_idx_names: List[str], keep: bool
+) -> NumpyEvent:
+    """
+    Implementation function to drop specified index names from a NumpyEvent.
```
achoum, line 20: Nit: Function documentation often reads better when it starts with a 3-rd person verb. For example in this case, this would be:

```
Drops part of the index from an event.
...
```

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,133 @@
...
+        return labels
+
+    def _generate_output_features(
+        self, event: Event, labels: List[str], keep: bool
+    ) -> List[Feature]:
+        output_features = [
+            Feature(name=feature.name, dtype=feature.dtype)
+            for feature in event.features()
+        ]
+        if keep:
```
achoum, line 94: If keep is true, can you ensure the input event did not already have a feature with the same name?

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,181 @@
...
+        dst_idx_metadata[dst_idx_lvl]["block_length"] += this_block_length
+        dst_idx_metadata[dst_idx_lvl]["timestamps"].append(timestamps)
+        dst_idx_metadata[dst_idx_lvl]["features"].append(
+            drop_feats + event.data[src_idx_lvl]
+        )
+
+    # allocate memory for destination sampling & event
+    # sampling
+    dst_samp_data = {
+        idx_lvl: np.empty(metadata["block_length"])
```
achoum, line 100: Set dtype to float64

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,181 @@
...
+from temporian.implementation.numpy.data.event import NumpyEvent
+from temporian.implementation.numpy.data.event import NumpyFeature
+from temporian.implementation.numpy.data.sampling import NumpySampling
+
+IndexMetadata = Dict[
+    str, Union[int, List[List[NumpyFeature]], List[np.ndarray]]
+]
+
+
+def _impl(
```
achoum, line 16: This is a great function :)

temporian/implementation/numpy/operators/tests/drop_index_test.py
```diff
@@ -0,0 +1,188 @@
...
+from temporian.implementation.numpy.operators.drop_index import (
+    DropIndexNumpyImplementation,
+)
+
+
+class DropIndexNumpyImplementationTest(absltest.TestCase):
+    def setUp(self) -> None:
+        # input event
+        self.input_evt = Event(
+            features=[Feature("sales", dtype=float)],
```
achoum, line 33: Use one of temporian dtypes?

temporian/implementation/numpy/operators/tests/drop_index_test.py
```diff
@@ -0,0 +1,188 @@
...
+                    [1, 2, 0.4, 16.0],
+                    [0, 1, 0.5, 11.0],
+                    [0, 1, 0.6, 12.0],
+                ],
+                columns=["store_id", "item_id", "timestamp", "sales"],
+            ),
+            index_names=["store_id", "item_id"],
+        )
+        # instance core operator
+        operator = DropIndexOperator(self.input_evt, labels=[])
```
achoum, line 164: In would raise an error if the user wants to drop nothing.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,181 @@
...
+    # index positions to drop
+    drop_idx_pos = [
+        pos
+        for pos, idx_name in enumerate(src_idx_names)
+        if idx_name not in dst_idx_names
+    ]
+    # intialize empty dict mapping destination index levels to block lengths,
+    # features, and timestamps
+    dst_idx_metadata: Dict[tuple, IndexMetadata] = defaultdict(
+        lambda: {"block_length": 0, "timestamps": [], "features": []}
```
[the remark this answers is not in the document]
ianspektor: using three dicts for `dst_idx_block_length`, `dst_idx_timestamps` and `dst_idx_features` would also be a good option imo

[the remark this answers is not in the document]
ianspektor: else a NamedTuple is also a good option, a class might be a bit overkill since it'd just be a data container

temporian/implementation/numpy/operators/drop_index.py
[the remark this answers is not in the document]
ianspektor: agree ☝🏼 and in this case I would even argue that having this logic inside the class is fine, since it's the only thing the class is actually implementing

ianspektor (approved):
Great work 👏🏼👏🏼👏🏼

temporian/core/operators/drop_index.py
getchebarne: Great suggestion.

temporian/core/operators/drop_index.py
getchebarne: Agreed!

temporian/implementation/numpy/data/sampling.py
```diff
@@ -27,6 +27,13 @@ def has_repeated_timestamps(self) -> bool:
 
         return False
 
+    @property
+    def dtypes(self) -> Dict[str, type]:
+        first_idx_lvl = next(iter(self.data))
```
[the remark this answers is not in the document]
getchebarne: When you say create a function, should the function's code be something like:
```
@property
def _first_idx_lvl(self):
    return next(iter(self.data))
```

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,181 @@
...
+        The modified NumpyEvent with the specified index names dropped.
+    """
+    # source index names
+    src_idx_names = event.sampling.index
+
+    # source sampling data
+    src_samp_data = event.sampling.data
+
+    # destination index names
+    dst_idx_names = [
```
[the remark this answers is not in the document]
getchebarne: I like it!

temporian/implementation/numpy/operators/tests/drop_index_test.py
```diff
@@ -0,0 +1,188 @@
...
+from temporian.implementation.numpy.data.event import NumpyEvent
+from temporian.implementation.numpy.operators.drop_index import (
+    DropIndexNumpyImplementation,
+)
+
+
+class DropIndexNumpyImplementationTest(absltest.TestCase):
+    def setUp(self) -> None:
+        # input event
+        self.input_evt = Event(
```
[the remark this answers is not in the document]
getchebarne: Thanks!

temporian/implementation/numpy/data/sampling.py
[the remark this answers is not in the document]
achoum: Yes, probably. I would make it a function instead of a property though.

temporian/implementation/numpy/operators/drop_index.py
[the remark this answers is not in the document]
achoum: +1 for NamedTuple.

I would avoid to create three different dictionaries as to avoid doing unnecessary dictionary queries.

temporian/implementation/numpy/operators/tests/drop_index_test.py
```diff
@@ -0,0 +1,188 @@
...
+            operator = DropIndexOperator(self.input_evt, labels="dept_id")
+
+        # one non-existing label
+        with self.assertRaisesRegex(KeyError, r"\['dept_id'\]"):
+            operator = DropIndexOperator(
+                self.input_evt, labels=["store_id", "item_id", "dept_id"]
+            )
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
achoum: My bad, I wanted to mean that the NumpyEvent contains no timestamps.

achoum (approved):

temporian/implementation/numpy/operators/drop_index.py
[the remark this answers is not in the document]
getchebarne: aren't namedtuples inmutable?

Issue #46, opened 2023-03-20
Convert methods to `property` where appropiate


ianspektor:
Methods with no arguments that just return a value from the class' internal state should be made properties via the `@property` decorator. This makes the syntax cleaner when calling the getter and makes it clear that the method is just returning one of the class' values.

We can additionally use the `@property.setter` where applicable.

Examples:
- core sampling: index, creator, set_creator
- core feature: name, dtype, sampling, creator, set_sampling, set_creator
etc.

diegomarvid:
Agree, `schema()` should be a property as well

ianspektor:
> schema() should be a property as well
not too sure about that, since it's creating and returning a new object

PR #53, opened 2023-03-21
Unix timestamp samplings

ianspektor:
Add `is_unix_timestamp` attr to samplings

Didn't include UTC in the name of the attribute since unix time is implicitly UTC - there's no non-UTC unix timestamps. Lmk if you still think it would be a good idea to have it in the attribute name, IMO it makes it harder to read.

### DIFF
--- a/temporian/core/data/sampling.py
+++ b/temporian/core/data/sampling.py
@@ -22,9 +22,11 @@ class Sampling(object):
         self,
         index: List[str],
         creator: Optional[Any] = None,
+        is_unix_timestamp: bool = False,
     ):
         self._index: List[str] = index
         self._creator = creator
+        self._is_unix_timestamp = is_unix_timestamp
 
     def __repr__(self):
         return f"Sampling<index:{self._index},id:{id(self)}>"
@@ -37,3 +39,6 @@ class Sampling(object):
 
     def set_creator(self, creator):
         self._creator = creator
+
+    def is_unix_timestamp(self) -> bool:
+        return self._is_unix_timestamp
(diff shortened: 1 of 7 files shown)


temporian/implementation/numpy/data/event.py
```diff
@@ -205,7 +209,9 @@ def from_dataframe(
                 for feature in feature_columns
             ]
 
-        numpy_sampling = NumpySampling(index=index_names, data=sampling)
+        numpy_sampling = NumpySampling(
+            index=index_names, data=sampling, is_unix_timestamp=True
```
achoum, line 213: Can the timestamps be provided as a float or int column? If so, I think the is_unix_timestamp should be False.

achoum (approved):
Looks good.
Left a new nits.

temporian/core/data/sampling.py
```diff
@@ -37,3 +39,6 @@ def creator(self):
 
     def set_creator(self, creator):
         self._creator = creator
+
+    def is_unix_timestamp(self) -> bool:
```
diegomarvid, line 43: Can this be a @property as it doesn't receive any attributes?

diegomarvid (approved):
Great!

getchebarne (approved):

temporian/implementation/numpy/data/event.py
ianspektor: Good catch, will fix and add test

temporian/core/data/sampling.py
ianspektor: Not addressing as part of this PR since all of `Sampling`'s methods already not properties, and https://github.com/google/temporian/issues/46 is open to deal with this globally

PR #54, opened 2023-03-21
move calendar operators to subdir in core and numpy

ianspektor:


temporian/core/operators/BUILD
```diff
@@ -14,7 +14,6 @@ py_library(
         ":arithmetic",
         ":assign",
         ":base",
```
achoum, line 16: PR43 change things, but we likely still need to register all the opts here: //temporian/core/operators/calendar:day_of_month", etc.

achoum (approved):
Looks good

diegomarvid (approved):
LGTM. Just to think about it, should we create a subfolder in tests for Calendar operators? I don't think it's needed right now but we may have several tests with Calendar & Window operators.

getchebarne (approved):

PR #55, opened 2023-03-21
Support Duration List in Lag

diegomarvid:
Support duration list in lag operator to handle several lags at once.

### DIFF
--- a/temporian/core/operators/lag.py
+++ b/temporian/core/operators/lag.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 """Lag operator."""
+from typing import List
+from typing import Union
 
 from temporian.core import operator_lib
 from temporian.core.data.duration import Duration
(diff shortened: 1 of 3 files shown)


temporian/core/operators/arithmetic.py
```diff
@@ -76,9 +76,6 @@ def __init__(
         if not isinstance(resolution, Resolution):
             raise ValueError("resolution must be a Resolution.")
 
-        if event_1.sampling() != event_2.sampling():
-            raise ValueError("event_1 and event_2 must have same sampling.")
```
ianspektor, line 80: Adding item to tomorrow's agenda to discuss best way to handle this

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+            f" {type(duration)}"
+        )
+
+    used_duration = duration if not should_leak else -duration
 
     return LagOperator(
         event=event,
-        duration=duration,
+        duration=used_duration,
     ).outputs()["event"]
```
ianspektor, line 122: This logic could be significantly simplified if we converted a single duration to a list first thing
```
if isinstance(duration, Duration):
    duration = [duration]

if not isinstance(duration, list):
    raise ...

<logic assuming duration is now a list of durations>
```

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
 
 
+def lag(
+    event: Event, duration: Union[Duration, List[Duration]]
+) -> Union[Event, List[Event]]:
```
ianspektor, line 127: Would be better to return a single event with one feature per each of the created lags. I think at some point we even talked about operators only being able to output events (probably before you joined us).

ianspektor (changes requested):

ianspektor (approved):

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+        duration: Duration to lag by. Can be a list of Durations.
+
+    Returns:
+        Lagged event. If a list of Durations is provided, a list of lagged
+        events is returned.
+    """
+    return _base_lag(event=event, duration=duration)
 
 
 def leak(event: Event, duration: Duration) -> Event:
```
achoum, line 141: Update pytype accordingly

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
-    if duration <= 0:
-        raise ValueError("duration must be greater than zero")
+    """Leak operator.
```
achoum, line 142: Can you an an explanation what "leak" means here?
Same for lag

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
 operator_lib.register_operator(LagOperator)
```
achoum, line 86: In "build_op_definition", the duration is defined as a float64. Can you make sure (through all the code) that the build_op_definition matches the implementation (pytype, isintance, etc.). I think, currently, some of the pytype are outdated.

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+    duration: Union[Duration, List[Duration]],
+    should_leak: bool = False,
+) -> Event:
+    """Base Lag function."""
 
-    return LagOperator(
-        event=event,
-        duration=duration,
-    ).outputs()["event"]
+    if isinstance(duration, (Duration, int)):
```
achoum, line 96: (Duration, int) is written multiple times. Can you find a way to have a single definition of the supported types?

temporian/implementation/numpy/operators/tests/lag_test.py
```diff
@@ -72,11 +77,73 @@ def test_correct_lag(self) -> None:
...
+
+        event = numpy_input_event.schema()
+
+        # lag multiple durations
+        lags = lag(event=event, duration=[1, 2])
+
+        # assign multiple lags to output event
+        output_event = event
+        for lagged_event in lags:
+            output_event = assign(output_event, lagged_event)
```
achoum, line 133: Can you test directly the output of the lag/leak operator (instead of using `assign`)?

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
 operator_lib.register_operator(LagOperator)
 
 
-def lag(event: Event, duration: Duration) -> Event:
-    if duration <= 0:
-        raise ValueError("duration must greater than zero")
+def _base_lag(
```
achoum, line 89: Why creating an implementation function here? It seems this is the entirety of the

achoum, line 89: This function is used both for lag and leak. This is the lag/leak file. Therefore, it is probably okay to call this method: "_implementation".

temporian/core/operators/lag.py
```diff
@@ -84,21 +86,67 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+    event: Event,
+    duration: Union[Duration, List[Duration]],
+    should_leak: bool = False,
```
achoum, line 92: s/should_leak/leak or is_leak

achoum (approved):
Looks good.
Left a few comments.

PR #60, opened 2023-03-23
SetIndexOperator

getchebarne:
This PR implements the `set_index` logic. This operator can be used to append a level to the existing index, or wipe the existing index in favor of a new one. Depends on #51.

### DIFF
--- a/temporian/core/operators/set_index.py
+++ b/temporian/core/operators/set_index.py
@@ -15,23 +15,23 @@ class SetIndexOperator(Operator):
     def __init__(
         self,
         event: Event,
-        labels: Optional[Union[List[str], str]] = None,
+        feature_names: Optional[Union[List[str], str]] = None,
         append: bool = False,
     ) -> None:
         super().__init__()
 
-        # process labels
-        labels = self._process_labels(event, labels)
+        # process feature_names
+        feature_names = self._process_feature_names(event, feature_names)
 
         # input event
         self.add_input("event", event)
 
         # attributes
-        self.add_attribute("labels", labels)
+        self.add_attribute("feature_names", feature_names)
         self.add_attribute("append", append)
 
         # output features
-        output_features = self._generate_output_features(event, labels)
+        output_features = self._generate_output_features(event, feature_names)
 
         # output sampling
 
@@ -40,10 +40,14 @@ class SetIndexOperator(Operator):
                 (index_name, index_dtype)
                 for index_name, index_dtype in event.sampling.index_dtypes.items()
             ]
-            + [(index_name, event.dtypes[index_name]) for index_name in labels]
+            + [
+                (index_name, event.dtypes[index_name])
+                for index_name in feature_names
+            ]
             if append
             else [
-                (index_name, event.dtypes[index_name]) for index_name in labels
+                (index_name, event.dtypes[index_name])
+                for index_name in feature_names
             ]
         )
         # output event
@@ -57,30 +61,30 @@ class SetIndexOperator(Operator):
         )
         self.check()
 
-    def _process_labels(
+    def _process_feature_names(
         self,
         event: Event,
-        labels: Optional[Union[List[str], str]],
+        feature_names: Optional[Union[List[str], str]],
     ) -> List[str]:
-        if isinstance(labels, str):
-            labels = [labels]
-        missing_labels = [
+        if isinstance(feature_names, str):
+            feature_names = [feature_names]
+        missing_feature_names = [
             label
-            for label in labels
+            for label in feature_names
             if label not in [feature.name for feature in event.features]
         ]
-        if missing_labels:
-            raise KeyError(missing_labels)
+        if missing_feature_names:
+            raise KeyError(missing_feature_names)
 
-        return labels
+        return feature_names
 
     def _generate_output_features(
-        self, event: Event, labels: List[str]
+        self, event: Event, feature_names: List[str]
     ) -> List[Feature]:
         output_features = [
             Feature(name=feature.name, dtype=feature.dtype)
             for feature in event.features
-            if feature.name not in labels
+            if feature.name not in feature_names
         ]
         return output_features
 
@@ -90,7 +94,7 @@ class SetIndexOperator(Operator):
             key="SET_INDEX",
             attributes=[
                 pb.OperatorDef.Attribute(
-                    key="labels",
+                    key="feature_names",
                     type=pb.OperatorDef.Attribute.Type.REPEATED_STRING,
                     is_optional=False,
                 ),
@@ -111,6 +115,6 @@ operator_lib.register_operator(SetIndexOperator)
 
 
 def set_index(
-    event: Event, labels: Optional[List[str]] = None, append: bool = True
+    event: Event, feature_names: List[str], append: bool = False
 ) -> Event:
-    return SetIndexOperator(event, labels, append).outputs["event"]
+    return SetIndexOperator(event, feature_names, append).outputs["event"]
--- a/temporian/core/operators/set_index.py
+++ b/temporian/core/operators/set_index.py
@@ -34,12 +34,17 @@ class SetIndexOperator(Operator):
         output_features = self._generate_output_features(event, labels)
 
         # output sampling
+
         output_sampling = Sampling(
             index=[
-                index_name for index_name in event.sampling().index() + labels
+                (index_name, index_dtype)
+                for index_name, index_dtype in event.sampling.index_dtypes.items()
             ]
+            + [(index_name, event.dtypes[index_name]) for index_name in labels]
             if append
-            else labels
+            else [
+                (index_name, event.dtypes[index_name]) for index_name in labels
+            ]
         )
         # output event
         self.add_output(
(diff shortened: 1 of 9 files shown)


temporian/implementation/numpy/operators/set_index.py
```diff
@@ -0,0 +1,186 @@
...
+            tuple(x)
+            for x in zip(*[src_idx_lvl_feats[i].data for i in dst_idx_pos])
+        ]
+        for target_dst_idx_suff in set(dst_idx_suffs):
+            # find all occurrences of destination suff in source event
+            dst_idx_suff_pos = [
+                idx
+                for idx, dst_idx_suff in enumerate(dst_idx_suffs)
+                if dst_idx_suff == target_dst_idx_suff
+            ]
```
ianspektor, line 54: I "feel" creating `dst_idx_suffs` + creating `dst_idx_suff_pos` could be done in the same pass through the `src_idx_lvl_feats[i].data` array - but its hard to say for sure just from looking at the code. Leaving the thought here in case it makes sense and you want to give it a go.

temporian/implementation/numpy/operators/set_index.py
```diff
@@ -0,0 +1,186 @@
...
+                for feat_name, feat_pos in dst_feat_pos.items()
+            ]
+            # fill sampling data
+            dst_samp_data[dst_idx_lvl] = event.sampling.data[src_idx_lvl][
+                dst_idx_suff_pos
+            ]
+
+    # finally, sort according to timestamps. TODO: this is merging sorted
+    # arrays, we should later improve this code by avoiding the full sort
+    _sort_by_timestamp(dst_event_data, dst_samp_data)
```
ianspektor, line 72: Might be wrong, but won't this be sorted already? Since we're taking subarrays of already sorted arrays

ianspektor (approved):

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,220 @@
...
+        index_names: List[str],
+        keep,
+    ) -> None:
+        super().__init__()
+
+        # input event
+        self.add_input("event", event)
+
+        # attributes
+        self.add_attribute("index_names", index_names)
```
achoum, line 39: Nit: Would a shorted name also work here?

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,220 @@
...
+    ]
+    if missing_index_names:
+        raise KeyError(
+            f"{missing_index_names} are missing from the input event's index."
+        )
+
+    return index_names
+
+
+def drop_index(
```
achoum, line 165: This is a very good documentation :).

temporian/core/operators/set_index.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @classmethod
+    def build_op_definition(cls) -> pb.OperatorDef:
+        return pb.OperatorDef(
+            key="SET_INDEX",
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="labels",
+                    type=pb.OperatorDef.Attribute.Type.REPEATED_STRING,
+                    is_optional=False,
```
achoum, line 90: Nit: The "is_optional" is optional and defaults to false.

temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +94,13 @@ def feature_names(self) -> List[str]:
         # indexes should have the same names
         return [feature.name for feature in self._first_index_features]
 
+    @property
+    def dtypes(self) -> Dict[str, type]:
```
achoum, line 98: Nit: Functions that are only for the implementation can be made private using an "_" prefix.

temporian/implementation/numpy/data/event.py
```diff
@@ -94,6 +94,13 @@ def feature_names(self) -> List[str]:
         # indexes should have the same names
         return [feature.name for feature in self._first_index_features]
 
+    @property
```
achoum, line 97: s/type/DType ? 
https://github.com/google/temporian/blob/main/temporian/core/data/dtype.py#L27

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,161 @@
...
+
+class DropIndexNumpyImplementation:
+    def __init__(self, operator: DropIndexOperator) -> None:
+        """
+        Initializes an instance of the DropIndexNumpyImplementation class.
+
+        Args:
+            operator: The DropIndexOperator instance.
+        """
+        self.operator = operator
```
achoum, line 24: I proposed and summitted a design where there operator attribute is handled by an abstract implementation class.
See https://github.com/google/temporian/blob/main/temporian/implementation/numpy/operators/glue.py#L29 for an example.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,161 @@
...
+        # index positions to drop
+        drop_index_pos = [
+            pos
+            for pos, idx_name in enumerate(src_index_names)
+            if idx_name not in dst_index_names
+        ]
+        # intialize empty dict mapping destination index levels to block lengths,
+        # features, and timestamps
+        dst_index_metadata: Dict[tuple, IndexMetadata] = defaultdict(
+            lambda: {"block_length": 0, "timestamps": [], "features": []}
```
achoum, line 64: Similarly as for the DropIndex implementation. I would use a class or named tuple instead.

temporian/core/operators/drop_index.py
```diff
@@ -0,0 +1,220 @@
...
+            if self.attributes()["keep"]
+            else feature_names
+        )
+
+    @property
+    def dst_index_names(self) -> List[str]:
+        return [
+            name
+            for name in self.inputs()["event"].sampling().index()
+            if name not in self.attributes()["index_names"]
```
achoum, line 102: I would cache the dictionary query i.e. stored "self.attributes()["index_names"]" in a variable.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        # loop over source index levels gathering destination index metadata
+        for src_index_lvl, timestamps in event.sampling.data.items():
+            # destination index level
+            dst_index_lvl = tuple((src_index_lvl[i] for i in keep_index_pos))
+
+            # number of samples in this index level
+            this_block_length = len(timestamps)
+
+            # collapse index level
```
achoum, line 74: What about change the comment to something like: Create a feature for the dropped index.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,161 @@
...
+                        idx_name,
+                        np.array(
+                            [src_index_lvl[i]] * this_block_length,
+                            dtype=dst_dtypes[idx_name],
+                        ),
+                    )
+                    for idx_name, i in zip(drop_index_names, drop_index_pos)
+                ]
+                if keep
+                else []
```
achoum, line 87: Moving the if outside of the [] would help the readability.

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -0,0 +1,161 @@
...
+            pos
+            for pos, idx_name in enumerate(src_index_names)
+            if idx_name not in dst_index_names
+        ]
+        # intialize empty dict mapping destination index levels to block lengths,
+        # features, and timestamps
+        dst_index_metadata: Dict[tuple, IndexMetadata] = defaultdict(
+            lambda: {"block_length": 0, "timestamps": [], "features": []}
+        )
+        # loop over source index levels gathering destination index metadata
```
achoum, line 66: Long variable names are informative, but they can also make reading the code more difficult.
If there is not possible confusion, what about "src_index"?

temporian/implementation/numpy/operators/test/drop_index_test.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        # instance core operator
+        operator = DropIndexOperator(
+            self.input_evt, index_names=["store_id"], keep=False
+        )
+        # instance operator implementation
+        operator_impl = DropIndexNumpyImplementation(operator)
+
+        # call operator
+        op_numpy_output_evt = operator_impl(event=self.numpy_input_evt)["event"]
```
achoum, line 201: Can you call "call" instead?

op_numpy_output_evt = operator_impl.call(event=self.numpy_input_evt)["event"]


See example: https://github.com/google/temporian/blob/main/temporian/implementation/numpy/operators/test/lag_test.py#L82

achoum (approved):
Looks good.
Left a few comments.

temporian/implementation/numpy/operators/set_index.py
getchebarne: it is but you would also need another loop to count the number of elements of each suffix, so that you can later allocate the appropriate memory for each of them

temporian/implementation/numpy/operators/set_index.py
getchebarne: not necessarily - they would also need to be sorted suffix-level-wise for that to happen

temporian/implementation/numpy/data/event.py
getchebarne: this might be useful for the end user too I think

temporian/implementation/numpy/operators/drop_index.py
getchebarne: I think that keeping the distinction between an index and an index level of that index provides more readability

PR #61, opened 2023-03-23
Check np.object column is only of type str

diegomarvid:
In this PR I change the DataFrame to NumpyEvent functionality to only allow for string-only columns. All numpy objects columns will be rejected if they have mixed values.

### DIFF
--- a/temporian/implementation/numpy/data/event.py
+++ b/temporian/implementation/numpy/data/event.py
@@ -174,15 +174,17 @@ class NumpyEvent:
 
         # check column dtypes, every dtype should be a key of DTYPE_MAPPING
         for column in df.columns:
-            # if dtype is object, check if it is a string column
+            # if dtype is object, check if it only contains string values
             if df[column].dtype.type is np.object_:
                 if not df[column].apply(lambda x: isinstance(x, str)).all():
                     raise ValueError(
-                        "String columns with mixed types are not supported"
+                        "Object columns are only allowed if they contain only"
+                        " string values"
                     )
-                # convert string column to np.string_
+                # convert object column to np.string_
                 df[column] = df[column].astype(np.string_)
 
+            # convert pandas' StringDtype to np.string_
             elif df[column].dtype.type is str:
                 df[column] = df[column].astype(np.string_)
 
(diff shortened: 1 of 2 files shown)


temporian/implementation/numpy/data/event.py
```diff
@@ -174,15 +174,14 @@ def from_dataframe(
 
         # check column dtypes, every dtype should be a key of DTYPE_MAPPING
         for column in df.columns:
-            # if dtype is object, convert to string
+            # if dtype is object, check if it is a string column
             if df[column].dtype.type is np.object_:
```
ianspektor, line 178: What happens if the column is already `np.string_`? We should accept that too

ianspektor: adding it to DTYPE_MAPPING like we discussed might be a solution, make sure to test to confirm

temporian/implementation/numpy/data/event.py
```diff
@@ -174,17 +174,22 @@ def from_dataframe(
 
         # check column dtypes, every dtype should be a key of DTYPE_MAPPING
         for column in df.columns:
-            # if dtype is object, convert to string
+            # if dtype is object, check if it is a string column
```
ianspektor, line 177: `check if it only contains string values`?

temporian/implementation/numpy/data/event.py
```diff
@@ -174,17 +174,22 @@ def from_dataframe(
...
-                    df[column] = df[column].astype(np.string_)
-                except ValueError as exc:
+                if not df[column].apply(lambda x: isinstance(x, str)).all():
                     raise ValueError(
-                        f"Column {column} has dtype object, but cannot be"
-                        " converted to string."
-                    ) from exc
-
-            elif df[column].dtype.type not in DTYPE_MAPPING:
+                        "String columns with mixed types are not supported"
```
ianspektor, line 181: `object columns are only allowed if they contain only string values` or similar?

temporian/implementation/numpy/data/event.py
```diff
@@ -174,17 +174,22 @@ def from_dataframe(
...
+                    )
+                # convert string column to np.string_
+                df[column] = df[column].astype(np.string_)
+
+            elif df[column].dtype.type is str:
```
ianspektor, line 186: would add a comment here stating that this checks for pandas' StringDtype

ianspektor (approved):

PR #63, opened 2023-03-23
Migrate from `np.string_` to `np.str_`

diegomarvid:
- [x] Fix `to_dataframe()` error when NumpyEvent had multiple str indexes.
- [x] Fix `from_dataframe()` error when NumpyEvent had multiple str indexes.

We now use np.str_ instead of np.string_. No binary string is used throughout the code. We may want to change to binary string in the future for more efficient encoding (Unicode vs ASCII).

### DIFF
--- a/temporian/implementation/numpy/data/event.py
+++ b/temporian/implementation/numpy/data/event.py
@@ -262,8 +262,12 @@ class NumpyEvent:
             data["timestamp"].extend(timestamps)
             for feature in features:
                 data[feature.name].extend(feature.data)
-            for index_key in self.sampling.index:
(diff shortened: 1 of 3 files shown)


achoum:
This PR seems to be only about the fix. Should it be reviewed and submitted just with the fixes?

diegomarvid:
> This PR seems to be only about the fix. Should it be reviewed and submitted just with the fixes?

Yeah, initially I wanted to implement the Base Window and started by changing the tests with the pd DataFrame format and encounter some errors. I may change the name of the PR and do another one for the rest.

ianspektor (approved):

PR #67, opened 2023-03-25
Event plotting

achoum:
Please review from PR by comparing it to branch `gbm_sampling`


temporian/implementation/numpy/data/plotter.py
```diff
@@ -0,0 +1,156 @@
...
+    colors = get_cmap("tab10").colors
+
+    px = 1 / plt.rcParams["figure.dpi"]
+
+    num_plots = 0
+    for event in events:
+        if index not in event.data:
+            raise ValueError(
+                f"Index '{index}' does not exist in event. Check the available"
+                " indexes with 'event.index()' and provide one of those index"
```
diegomarvid, line 74: Can't we list `event.index()` in the error for better clarity?

Maybe say something like:
```python
f"Index '{index}' does not exist in the event. Possible indexes: {event.index()}"
```

ianspektor (approved):

temporian/implementation/numpy/data/plotter.py
achoum: Good idea.
*Done

achoum:
Thanks ! :)

PR #70, opened 2023-03-26
Op validation in unit tests

achoum:
Please review from PR by comparing it to branch `gbm_prefix`


ianspektor (approved):
Like the idea of checking outputs in tests 👍🏼 

nit: IMO it would make it clearer and more explicit if we called the implementation normally as we did before, and then explicitly called `_check_output`, since the developer wouldn't need to go check what `run_with_check` does and returns

PR #71, opened 2023-03-29
Moving Sum & Moving Count Operators

diegomarvid:
- [x] Implemented Moving Sum.
- [x] Implemented Moving Count.
- [x] Changed Core Base Window so each Window Op can decide their feature output dtype.

### DIFF
--- a/temporian/core/operators/window/base.py
+++ b/temporian/core/operators/window/base.py
@@ -50,14 +50,21 @@ class BaseWindowOperator(Operator, ABC):
 
         self.add_input("event", event)
 
+        output_features = [  # pylint: disable=g-complex-comprehension
+            Feature(
+                name=f"{self.prefix}_{f.name()}",
+                dtype=self.get_feature_dtype(f),
+                sampling=effective_sampling,
+                creator=self,
+            )
+            for f in event.features()
+        ]
+
         # output
         self.add_output(
             "event",
             Event(
-                features=self.output_features(
-                    event=event,
-                    sampling=effective_sampling,
-                ),
+                features=output_features,
                 sampling=effective_sampling,
                 creator=self,
             ),
@@ -93,10 +100,15 @@ class BaseWindowOperator(Operator, ABC):
         """Get the key of the operator definition."""
 
     @abstractmethod
-    def output_features(
-        self, event: Event, sampling: Sampling
-    ) -> list[Feature]:
-        """Get the list of features in the output event."""
+    def get_feature_dtype(self, feature: Feature) -> str:
+        """Get the dtype of the output feature.
+
+        Args:
+            feature: feature to get the dtype for.
+
+        Returns:
+            str: The dtype of the output feature.
+        """
 
     @property
     @abstractmethod
--- a/temporian/core/operators/window/base.py
+++ b/temporian/core/operators/window/base.py
@@ -21,6 +21,7 @@ from typing import Optional
 from temporian.core.data.duration import Duration
 from temporian.core.data.event import Event
 from temporian.core.data.feature import Feature
+from temporian.core.data.sampling import Sampling
 from temporian.core.operators.base import Operator
 from temporian.proto import core_pb2 as pb
 
@@ -49,21 +50,14 @@ class BaseWindowOperator(Operator, ABC):
 
         self.add_input("event", event)
 
-        output_features = [  # pylint: disable=g-complex-comprehension
-            Feature(
-                name=f"{self.output_feature_prefix}_{f.name()}",
-                dtype=f.dtype(),  # TODO: This might not be correct
-                sampling=effective_sampling,
-                creator=self,
-            )
-            for f in event.features()
-        ]
-
         # output
         self.add_output(
             "event",
             Event(
-                features=output_features,
+                features=self.output_features(
+                    event=event,
+                    sampling=effective_sampling,
+                ),
                 sampling=effective_sampling,
                 creator=self,
             ),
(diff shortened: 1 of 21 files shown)


temporian/core/operators/window/moving_sum.py
```diff
@@ -0,0 +1,93 @@
...
+    sampling: Optional[Event] = None,
+) -> Event:
+    """Moving Sum.
+
+    For each sampling, and for each feature independently, returns at time "t"
+    the sum of the feature in the time windows [t-window, t].
+
+    If "sampling" is provided, applies the operator for each timestamps of
+    "sampling". If "sampling" is not provided, apply the operator for each
+    timestamps of "event".
```
ianspektor, line 73: same grammatical nit as the same line in moving_count

temporian/core/operators/window/moving_sum.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Missing values are ignored.
+
+    If the window does not contain any values (e.g., all the values are missing,
+    or the window does not contain any sampling), outputs missing values.
```
ianspektor, line 78: In response to note in description: I would definitely output NaN instead of zero if all the values in the window are NaN (which is also what this docstring states)

temporian/core/operators/window/base.py
```diff
@@ -98,8 +92,13 @@ def build_op_definition(cls) -> pb.OperatorDef:
     def operator_def_key(cls) -> str:
         """Get the key of the operator definition."""
 
-    @classmethod
+    @abstractmethod
+    def output_features(
+        self, event: Event, sampling: Sampling
+    ) -> list[Feature]:
+        """Get the list of features in the output event."""
```
ianspektor, line 99: wouldn't `output_features_dtypes(cls, input_event: Event) -> List[DType]` or similar be enough?

temporian/implementation/numpy/operators/window/moving_count.py
```diff
@@ -0,0 +1,59 @@
...
+from temporian.core.operators.window.moving_count import (
+    MovingCountOperator,
+)
+from temporian.implementation.numpy.operators.window.base import (
+    BaseWindowNumpyImplementation,
+)
+
+
+class MovingCountNumpyImplementation(BaseWindowNumpyImplementation):
+    """Numpy implementation of the moving sum operator."""
```
ianspektor, line 27:
```suggestion
    """Numpy implementation of the moving count operator."""
```

temporian/implementation/numpy/operators/window/moving_count.py
```diff
@@ -0,0 +1,59 @@
...
+        ignoring the 'nan' values.
+
+        Args:
+            values: A 2D NumPy array with shape (n, m) where each row represents
+                a  window of data points. 'n' is the length of the feature, and
+                'm' is the size of the window. The array can contain 'nan'
+                values as padding.
+
+        Returns:
+            np.array: A 1D NumPy array with shape (n,) containing the moving
```
ianspektor, line 50:
```suggestion
            np.array: A 1D NumPy array with shape (n,) containing the
```

temporian/implementation/numpy/operators/window/moving_count.py
```diff
@@ -0,0 +1,59 @@
...
+
+class MovingCountNumpyImplementation(BaseWindowNumpyImplementation):
+    """Numpy implementation of the moving sum operator."""
+
+    def __init__(self, operator: MovingCountOperator) -> None:
+        super().__init__(operator)
+
+    def _apply_operation(self, values: np.array) -> np.array:
+        """
+        Calculates the moving count of the values in each row of the input array.
```
ianspektor, line 34:
```suggestion
        Calculates the count of the values in each row of the input array.
```

temporian/implementation/numpy/operators/window/moving_count.py
```diff
@@ -0,0 +1,59 @@
...
+
+        The input array should have a shape (n, m), where 'n' is the length of
+        the feature and 'm' is the size of the window. Each row represents a
+        window of data points, with 'nan' values used for padding when the
+        window size is  smaller than the number of data points in the time
+        series. The function  computes the moving count for each row (window) by
```
ianspektor, line 40:
```suggestion
        series. The function computes the count for each row (window),
```

temporian/implementation/numpy/operators/window/moving_count.py
```diff
@@ -0,0 +1,59 @@
...
+                a  window of data points. 'n' is the length of the feature, and
+                'm' is the size of the window. The array can contain 'nan'
+                values as padding.
+
+        Returns:
+            np.array: A 1D NumPy array with shape (n,) containing the moving
+                    count for each row (window) in the input array.
+
+        """
+        return np.count_nonzero(~np.isnan(values), axis=1).astype(np.int64)
```
ianspektor, line 54: this means we're treating `0` as a missing value right? I don't think we should

temporian/implementation/numpy/operators/window/moving_sum.py
```diff
@@ -0,0 +1,59 @@
...
+
+class MovingSumNumpyImplementation(BaseWindowNumpyImplementation):
+    """Numpy implementation of the moving sum operator."""
+
+    def __init__(self, operator: MovingSumOperator) -> None:
+        super().__init__(operator)
+
+    def _apply_operation(self, values: np.array) -> np.array:
+        """
+        Calculates the moving sum of the values in each row of the input array.
```
ianspektor, line 34: Same nits about removing "moving" from the other docstring - this function is just applying a sum over rows, not a moving sum

temporian/implementation/numpy/operators/window/moving_count.py
diegomarvid: It doesn't. We generate a boolean array where False or 0 is that the value is not a np.nan and True or 1 for np.nan values. If we invert the boolean array, we have 1 for each non-nan value. And we sum all of them. It should also be equivalent to the sum of the array.

ianspektor (approved):

PR #72, opened 2023-03-29
Benchmarks

ianspektor:


achoum:
It seems the benchmark can run well with sh_binary (instead of py_binary). However, event in this case, scalene is not reporting memory usage. Maybe this is another issue?

See https://github.com/google/temporian/blob/gbm_benchmark/benchmarks/BUILD

Running: `bazel run benchmarks:run` in branch `gbm_benchmark`.

ianspektor:
> See https://github.com/google/temporian/blob/gbm_benchmark/benchmarks/BUILD

Thanks for the help @achoum! Merged your work into this branch.

Managed to do something similar by running scalene in `bazel-bin/benchmarks/run.runfiles/temporian/benchmarks` but memory isn't reported either, integrated a new tool.

temporian/implementation/numpy_cc/operators/sample.so
ianspektor, line 1: Should this file be tracked? @achoum

achoum: No. This is an rest of testing things :)

benchmark/BUILD
```diff
@@ -0,0 +1,23 @@
...
+py_binary(
+    name = "basic",
+    srcs = ["basic.py"],
+    srcs_version = "PY3",
+    deps = ["//temporian"],
+)
+
+sh_binary(
+    name = "time",
+    srcs = ["scalene.sh"],
```
achoum, line 15: Nit: Calling the .sh with the same key as the rule. e.g., s/scalene.sh/time.sh

benchmark/mprof.sh
```diff
@@ -0,0 +1,11 @@
+export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
```
achoum, line 1: What about adding "set -vex" (or a subset of it) to make sure it runs correctly?

benchmark/mprof.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# TODO: provide flag to `bazel run benchmark:memory` to choose between
```
achoum, line 3: Info / for later: Flags can be passed to binaries as follow:
bazel run benchmark:memory -- <flags to pass to the :memory binary>

benchmark/scalene.sh
```diff
@@ -0,0 +1,3 @@
+export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+
+scalene benchmark/basic.py
```
achoum, line 3: I found the option "--reduced-profile" to be useful (it only shows the lines that consume a lot of resources). Like for mprof.sh, this could be documented here.

achoum (approved):
This is awesome.

PR #76, opened 2023-03-31
Dtype Promotion

diegomarvid:
In this PR I introduce a type promotion hierarchy for Temporian DTypes. This enables us to dynamically understand the output of an operator if the resulting dtype follows [numpy type promotion rules.](https://numpy.org/doc/stable/reference/generated/numpy.result_type.html)

This addition enables the Arithmetic operator to support different feature types.

Expected dtype outputs:

| dtype1  | dtype2  | Output dtype |
|---------|---------|--------------|
| INT32   | INT32   | INT32        |
| INT32   | INT64   | INT64        |
| INT32   | FLOAT32 | FLOAT32      |
| INT32   | FLOAT64 | FLOAT64      |
| INT64   | INT64   | INT64        |
| INT64   | FLOAT32 | FLOAT64      |
| INT64   | FLOAT64 | FLOAT64      |
| FLOAT32 | FLOAT32 | FLOAT32      |
| FLOAT32 | FLOAT64 | FLOAT64      |
| FLOAT64 | FLOAT64 | FLOAT64      |
| STRING | * | ERROR      |
| * | STRING | ERROR      |

>__Note__:The division operator should always output FLOAT64 with any input dtype.

### DIFF
--- a/temporian/core/data/dtype.py
+++ b/temporian/core/data/dtype.py
@@ -38,3 +38,43 @@ def MissingValue(dtype: DType) -> Any:
         return ""
     else:
         raise ValueError(f"Non implemented type {dtype}")
+
+
+def get_resulting_dtype(dtype1: DType, dtype2: DType) -> DType:
+    """
+    Get the resulting dtype when combining two arrays with the specified dtypes.
+
+    This function takes two dtypes as input and returns the expected
+    output dtype based on a simplified dtype hierarchy.
+
+    Args:
+        dtype1: The dtype of the first array.
+        dtype2: The dtype of the second array.
+
+    Returns:
+        DType: The resulting dtype
+
+    Raises:
+        ValueError: If either of the input dtypes is not supported.
+    """
+    dtype_hierarchy = {
+        INT32: 1,
+        INT64: 2,
+        FLOAT32: 3,
+        FLOAT64: 4,
+    }
+
+    if dtype1 not in dtype_hierarchy or dtype2 not in dtype_hierarchy:
+        raise ValueError(
+            "Invalid dtype(s). Supported dtypes: int32, int64, float32, float64"
+        )
+
+    # Find the highest hierarchy value between the two input dtypes
+    max_hierarchy = max(dtype_hierarchy[dtype1], dtype_hierarchy[dtype2])
+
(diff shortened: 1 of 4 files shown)


temporian/core/data/dtype.py
```diff
@@ -38,3 +38,45 @@ def MissingValue(dtype: DType) -> Any:
...
+
+    if dtype1 not in dtype_hierarchy or dtype2 not in dtype_hierarchy:
+        supported_dtypes = ", ".join(dtype_hierarchy.keys())
+        raise ValueError(
+            f"Invalid dtype(s): [{dtype1}, {dtype2}]. Supported dtypes:"
+            f" {supported_dtypes}"
+        )
+
+    # Find the highest hierarchy value between the two input dtypes
+    max_hierarchy = max(dtype_hierarchy[dtype1], dtype_hierarchy[dtype2])
```
achoum, line 75: int32 => float32 or int64 => float32 is possibly a drop in precision.

I like this idea, but what about adding it to our next sync for the details?
For example, why not int32=>float32 and int64=>float64, or doing something like numpy (your link: https://numpy.org/doc/stable/reference/generated/numpy.result_type.html), or possibly just disallowing operations between different types (like in go.

temporian/core/data/dtype.py
```diff
@@ -38,3 +38,51 @@ def MissingValue(dtype: DType) -> Any:
...
+
+    # handle special case with int64 and float32
+    if (dtype1 == INT64 and dtype2 == FLOAT32) or (
```
getchebarne, line 78: nit: `dtype1` -> `dtype_1`, `dtype2` -> `dtype_2`

temporian/core/data/dtype.py
```diff
@@ -38,3 +38,51 @@ def MissingValue(dtype: DType) -> Any:
...
+        dtype1 == FLOAT32 and dtype2 == INT64
+    ):
+        return FLOAT64
```
getchebarne, line 81: I think this check should go before line 75 (`max_hierarchy = max(dtype_hierarchy[dtype1], ...`)

getchebarne (approved):

PR #78, opened 2023-04-03
General improvements

### LINKED ISSUE #46: Convert methods to `property` where appropiate

ianspektor:
Closes #46 

- Use `@property` and `@property.setter` where applicable
- Moved `NumpyFeature` to its own `feature` module (was in `event` module before)
- Fixed circular import problem so that now `creator` properties in core can be of type `Operator`
- Removed unused imports and unused build deps
- Change `"Event"` and `"NumpyEvent"` return types to `Event` and `NumpyEvent`

### DIFF
--- a/benchmark/scripts/basic.py
+++ b/benchmark/scripts/basic.py
@@ -92,7 +92,7 @@ def main():
     )
 
     # Print output's first row, useful to check reproducibility
-    print(res._first_index_features)
+    print(res.first_index_features())
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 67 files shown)


achoum (approved):
Awesome

getchebarne (approved):
awesome 💯 LGTM

PR #80, opened 2023-04-03
Change core and NumPy index to dict mapping names to dtypes

getchebarne:
This PR changes the definition of the core and NumPy index from a list of string to a dictionary mapping `strings` to `dtype`s. This is needed in order to specify kept feature `dtype`s when using the `DropIndex` operator. 

For future work: implement this logic as separate `Index` and `NumpyIndex` classes.

### DIFF
--- a/temporian/core/data/event.py
+++ b/temporian/core/data/event.py
@@ -14,9 +14,10 @@
 
 """An event is a collection (possibly empty) of timesampled feature values."""
 
-from typing import Any, List, Optional
+from typing import Any, Dict, List, Optional
 
 from temporian.core.data.feature import Feature
+from temporian.core.data.sampling import IndexDtypes
 from temporian.core.data.sampling import Sampling
 from temporian.utils import string
 
@@ -97,7 +98,7 @@ class Event(object):
 
 def input_event(
     features: List[Feature],
-    index: List[str] = [],
+    index: Dict[str, IndexDtypes] = {},
     name: Optional[str] = None,
     sampling: Optional[Sampling] = None,
 ) -> Event:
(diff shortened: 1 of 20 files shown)


temporian/core/data/event.py
```diff
@@ -97,7 +98,7 @@ def set_creator(self, creator):
 
 def input_event(
     features: List[Feature],
-    index: List[str] = [],
+    index: Dict[str, IndexDtypes] = {},
```
achoum, line 101: As per today's meeting, if this is too hard to do in this PR, can you add a todo to transform this into a List?

temporian/core/operators/propagate.py
```diff
@@ -46,18 +46,18 @@ def __init__(
...
+            set(event.sampling().index())
         )
         if len(overlap_features) > 0:
             raise ValueError(
                 "to contains feature names already present in the"
                 f" index: {list(overlap_features)}"
             )
 
-        new_index = event.sampling().index() + self._added_index
+        new_index = {**event.sampling().index(), **self._added_index}
```
achoum, line 60: Afterthoughts: I would implement the index list immediately. This index dictionary add a lot of changes. Note: You can also create a method sampling().index_keys() to minimize the change.

getchebarne: okay. I was thinking of doing something like this:

```
IndexDtypes = Union[dtype_lib.INT32, dtype_lib.INT64, dtype_lib.STRING]


class IndexLevel(NamedTuple):
    name: str
    dtype: IndexDtypes


class Sampling(object):
    def __init__(
        self,
        index: List[Tuple[str, IndexLevel]],
        ...
     ):
     self._index = [IndexLevel(name, dtype) for name, dtype in index] # convert regular tuples to IndexLevel
     ...
``` 

I prefer the user being able to specify the `index` with native python objects rather than having to call `IndexLevel`, which seems too verbose. What do you think?

PR #81, opened 2023-04-04
Split arithmetic operators into subclasses

### LINKED ISSUE #74: Change Arithmetic Operators to Abstract class
Do the same process with Arithmetic operators that we did with Calendar & Window

DonBraulio:
Closes #74

* Split `arithmetic.py` files (in core & implementation) -> `arithmetic/base.py, addition.py, multiplication.py, ...`
* Unlike the calendar and window operators (which also have subfolders), I added the imports in `arithmetic/__init__.py`, in order to be able to do `from temporian.core.operators.arithmetic import AdditionOperator, SubtractionOperator, ...` instead of referring to each sub-folder and need separate import lines or instantiate like `addition.AdditionOperator`, `subtraction.SubtractionOperator`, which seems cumbersome to me.

UPDATE:
* Added FloorDiv operator (`//`), and true division is now forbidden between integer types.
* Added arithmetic tests with multiple index levels and multiple features, including dtype float32.
* Added test to check the link between magic methods and arithmetic operators (`+,*,-,/,//`).
* Removed Resolution parameter (was not implemented yet anyway).


temporian/core/operators/BUILD
```diff
@@ -1,23 +1,26 @@
 package(
-    default_visibility = ["//visibility:public"],
-    licenses = ["notice"],
+    default_visibility=["//visibility:public"],
```
achoum, line 2: This is a strange formatting. BUILD files are generally treated like python files.
Can you check?

temporian/core/operators/all_operators.py
```diff
@@ -3,7 +3,7 @@
 # pylint: disable=unused-import
 from temporian.core.operators.arithmetic import divide
 from temporian.core.operators.arithmetic import multiply
-from temporian.core.operators.arithmetic import substract
+from temporian.core.operators.arithmetic import subtract
```
achoum, line 6: Nice catch.

temporian/core/operators/all_operators.py
```diff
@@ -3,7 +3,7 @@
 # pylint: disable=unused-import
 from temporian.core.operators.arithmetic import divide
 from temporian.core.operators.arithmetic import multiply
-from temporian.core.operators.arithmetic import substract
+from temporian.core.operators.arithmetic import subtract
 from temporian.core.operators.arithmetic import sum
```
achoum, line 7: Maybe this is the right time to rename "sum" into "add" or "addition" (like the operator name).

temporian/core/operators/arithmetic/BUILD
```diff
@@ -0,0 +1,76 @@
...
+    srcs=["__init__.py"],
+    srcs_version="PY3",
+    deps=[
+        ":base",
+        ":addition",
+        ":division",
+        ":subtraction",
+        ":multiplication",
+    ],
+)
```
achoum, line 20: Can you check if this rule is still used.
If not, can you remove it?

temporian/implementation/numpy/test/registered_operators_test.py
```diff
@@ -23,7 +23,10 @@ class RegisteredOperatorsTest(absltest.TestCase):
     def test_base(self):
         # Note: The operators are stored alphabetically.
         expected_implementations = [
-            "ARITHMETIC",
+            "ADDITION",
```
achoum, line 26: Nit: Sort the operator names alphabetically

achoum (approved):
Great!

temporian/core/operators/BUILD
getchebarne: @DonBraulio if you're using vscode, you can install the bazel extension to have automatic formatting on `BUILD` files

temporian/core/operators/all_operators.py
getchebarne: +1, otherwise we're overwriting a python built-in function

temporian/core/operators/arithmetic/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from temporian.core.operators.arithmetic.subtraction import (
+    subtract,
+    SubtractionOperator,
+)
+from temporian.core.operators.arithmetic.multiplication import (
+    multiply,
+    MultiplicationOperator,
+)
+
+# pylint: enable=unused-import
```
getchebarne, line 18: I thought that all `__init__.py` files should be empty? cc: @achoum

temporian/implementation/numpy/operators/all_operators.py
```diff
@@ -1,7 +1,10 @@
 """Imports all the packages containing operator implementations."""
 
 # pylint: disable=unused-import
-from temporian.implementation.numpy.operators import arithmetic
+from temporian.implementation.numpy.operators.arithmetic import addition
+from temporian.implementation.numpy.operators.arithmetic import subtraction
+from temporian.implementation.numpy.operators.arithmetic import multiplication
+from temporian.implementation.numpy.operators.arithmetic import division
```
getchebarne, line 7: nit: can we move these imports below the `temporian.implementation.numpy.operators` ones?

temporian/implementation/numpy/operators/arithmetic/base.py
```diff
@@ -13,24 +13,29 @@
...
+    def _do_operation(self, event_1_data, event_2_data):
+        """
+        Perform the actual arithmetic operation corresponding to the subclass
+        """
+
     def __call__(
         self, event_1: NumpyEvent, event_2: NumpyEvent
     ) -> Dict[str, NumpyEvent]:
-        """Sum two Events.
+        """Apply the corresponding arithmetic operation to two Events.
```
getchebarne, line 38: nit: `to` -> `between the`

temporian/implementation/numpy/operators/arithmetic/division.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class DivisionNumpyImplementation(BaseArithmeticNumpyImplementation):
+    """Actual numpy implementation to divide first event by the second one"""
+
+    def __init__(self, operator: DivisionOperator) -> None:
+        super().__init__(operator)
+
+    def _do_operation(self, event_1_data, event_2_data):
+        return event_1_data / event_2_data
```
getchebarne, line 29: to discuss: do we want alert the user if the division produced `np.nan` values?

cc: @achoum @ianspektor

getchebarne (approved):
great work! 
nit: can we add multiple index levels to the test events? we only have one currently (`store_id = 0`). Just to make sure the logic works index level-wise

temporian/core/operators/all_operators.py
DonBraulio: Yeah, if there's no good reason to use `sum` I think it's better too. I'm on it.

temporian/core/operators/arithmetic/__init__.py
DonBraulio: It's related with what I commented in the PR description. Maybe it's something to discuss.

temporian/core/operators/arithmetic/addition.py
```diff
@@ -0,0 +1,65 @@
...
+        event_1: First event
+        event_2: Second event
+        resolution: If resolution is Resolution.PER_FEATURE_IDX each feature
+            will be sum index wise. If resolution is Resolution.PER_FEATURE_NAME
+            each feature of event_1 will be sum with the feature in event_2
+            with the same name. Defaults to Resolution.PER_FEATURE_IDX.
+
+
+    Returns:
+        Event: Sum of event_1 and event_2 according to resolution.
```
ianspektor, line 59: nit: replace "sum" with "add" in docstring too for consistency

temporian/implementation/numpy/operators/arithmetic/base.py
```diff
@@ -13,24 +13,29 @@
...
+from temporian.core.operators.arithmetic.base import BaseArithmeticOperator
 from temporian.implementation.numpy.data.event import NumpyEvent
 from temporian.implementation.numpy.data.event import NumpyFeature
-from temporian.implementation.numpy import implementation_lib
 from temporian.implementation.numpy.operators.base import OperatorImplementation
 
 
-class ArithmeticNumpyImplementation(OperatorImplementation):
-    def __init__(self, operator: ArithmeticOperator) -> None:
+class BaseArithmeticNumpyImplementation(OperatorImplementation, ABC):
```
ianspektor, line 25: I think `ABC` parent class isn't needed here since `OperatorImplementation` has it already

temporian/implementation/numpy/operators/arithmetic/division.py
ianspektor: probably, I think torch for example does so, might be the same with other libs

temporian/core/operators/arithmetic/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+    DivisionOperator,
+)
+from temporian.core.operators.arithmetic.subtraction import (
+    subtract,
+    SubtractionOperator,
+)
+from temporian.core.operators.arithmetic.multiplication import (
+    multiply,
+    MultiplicationOperator,
+)
```
ianspektor, line 16: I'd rename these operators to the simple present form of the verb, rather than the noun (e.g. `DivisionOperator` -> `DivideOperator`) to keep it consistent with existing ones (we have `PropagateOperator`, `SelectOperator`, `GlueOperator`, etc. rather `PropagationOperator`, `SelectionOperator`, etc.)

ianspektor: same for the naming of the numpy implementations

ianspektor (approved):

temporian/implementation/numpy/operators/arithmetic/division.py
DonBraulio: Numpy already shows a RuntimeWarning on divide by zero (only once per run), and the result is `np.inf` in that case.

temporian/implementation/numpy/operators/arithmetic/base.py
DonBraulio: You're right, nice catch!

temporian/core/operators/arithmetic/__init__.py
DonBraulio: Yep, I agree, will do.

temporian/implementation/numpy/operators/arithmetic/floordiv.py
```diff
@@ -0,0 +1,35 @@
...
+
+class FloorDivideNumpyImplementation(BaseArithmeticNumpyImplementation):
+    """Actual numpy implementation to divide of first event by the second one
+    and take the result floor"""
+
+    def __init__(self, operator: FloorDivideOperator) -> None:
+        super().__init__(operator)
+
+    def _do_operation(self, event_1_feature, event_2_feature):
+        return event_1_feature.data // event_2_feature.data
```
diegomarvid, line 30: You ended the normal division with this to ensure same types:

```python
.astype(
            event_1_feature.data.dtype
        )
```
Why is it not here? Is this needed on the / and not on //? or is not needed on both?

temporian/implementation/numpy/operators/arithmetic/addition.py
```diff
@@ -25,8 +25,8 @@ class AdditionNumpyImplementation(BaseArithmeticNumpyImplementation):
     def __init__(self, operator: AdditionOperator) -> None:
         super().__init__(operator)
 
-    def _do_operation(self, event_1_data, event_2_data):
-        return event_1_data + event_2_data
+    def _do_operation(self, event_1_feature, event_2_feature):
```
getchebarne: can we add dtypes here?

temporian/implementation/numpy/operators/arithmetic/floordiv.py
DonBraulio: Nice catch. It was needed only in `/` because numpy would promote ints to floats automatically. Now, since it's forbidden the integer division, it's not needed there either (I'll remove it).
With this `//` operator, the output type is always the same as the input type.

diegomarvid: Float // float does still return float right?

temporian/implementation/numpy/operators/arithmetic/base.py
```diff
@@ -26,7 +26,9 @@ def __init__(self, operator: BaseArithmeticOperator) -> None:
         super().__init__(operator)
 
     @abstractmethod
-    def _do_operation(self, event_1_feature, event_2_feature):
+    def _do_operation(
```
diegomarvid, line 29: Can we also add that the return type is `np.ndarray`?

PR #83, opened 2023-04-05
Add index data type

getchebarne:
As per our last sync, this PR changes the definition of the core and NumPy index from a list of string to a list of tuples containing index level names and dtypes. This is needed in order to specify kept feature `dtype`s when using the `DropIndex` operator. 

For future work: implement this logic as separate `Index` and `NumpyIndex` classes. PR #51 depends on this.

### DIFF
--- a/temporian/core/data/sampling.py
+++ b/temporian/core/data/sampling.py
@@ -15,23 +15,34 @@
 """A sampling."""
 
 from __future__ import annotations
-from typing import List, Optional, TYPE_CHECKING
+from typing import List, NamedTuple, Optional, Tuple, TYPE_CHECKING, Union
 
+from temporian.core.data import dtype as dtype_lib
 
 if TYPE_CHECKING:
     from temporian.core.operators.base import Operator
 
 
+IndexDtypes = Union[dtype_lib.INT32, dtype_lib.INT64, dtype_lib.STRING]
+
+
+class IndexLevel(NamedTuple):
+    name: str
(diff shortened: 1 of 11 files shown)


temporian/core/data/sampling.py
```diff
@@ -53,3 +68,14 @@ def is_unix_timestamp(self) -> bool:
...
+    def _parse_index(
+        self, index: List[Union[Tuple[str, IndexDtypes], IndexLevel]]
+    ) -> List[IndexLevel]:
+        if all([isinstance(index_level, IndexLevel) for index_level in index]):
+            return index
+
+        return [
+            IndexLevel(index_name, index_dtype)
+            for index_name, index_dtype in index
+        ]
```
ianspektor, line 81: very nit: we could iterate over the list just once and for each item check if its an `IndexLevel`, else convert it, with a single comprehension

ianspektor (approved):

temporian/core/data/sampling.py
```diff
@@ -15,33 +15,48 @@
...
     def __repr__(self):
         return f"Sampling<index:{self.index},id:{id(self)}>"
 
     @property
-    def index(self) -> List[str]:
+    def index(self) -> List[IndexLevel]:
         return self._index
 
+    @property
+    def index_names(self) -> List[str]:
```
achoum, line 57: Nit: What about index_keys?

achoum (approved):
Awesome

PR #87, opened 2023-04-11
Boolean Operators Old

diegomarvid:
Implementation of the equal operator.

We currently only perform equality when the value and the feature have the same **general** dtype. This means we only compare integer with integer, float with float, string with string, and boolean with boolean.

This is helpful because if we define for example a value of 10 for comparison in Python. The default will be int64. But we may still want to compare with int columns of dtype np.int16, np.int32, np.int64. This is especially handy because the value is a scalar and is not expected to consider precision on Python scalars.

I had to introduce a new dtype BOOLEAN: np.bool_ for the operator's output. Feel free to give comments on this new dtype.

Currently, we have two different core classes: Boolean Feature and Boolean Scalar. We have two different classes because the inputs are very different in each case. However, the implementation for a boolean operation is the same for a numpy array and for a scalar. This is why even though we have two different Boolean Core Operators, we have the same NumpyImplementation.

In order to support this there has to be some code in the NumpyImplementation that enables the operator to receive two kinds of inputs: scalars and numpy events.  

Feel free to give more ideas on possible refactors for better code!


temporian/core/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+        self.check()
+
+    @classmethod
+    def build_op_definition(cls) -> pb.OperatorDef:
+        return pb.OperatorDef(
+            key="EQUAL",
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="value",
+                    # dtype depends on value dtype, this is not always float64
```
diegomarvid, line 64: How can I address this?

achoum: I think making the attribute type a list of allowed types (repeated Type allowed_types = 2;) could be a good solution. Let's chat about it.

temporian/core/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+"""Equal operator."""
+from temporian.core import operator_lib
+from temporian.core.data.dtype import BOOLEAN
+from temporian.core.data.event import Event
+from temporian.core.data.feature import Feature
+from temporian.core.operators.base import Operator
+from temporian.proto import core_pb2 as pb
+
+
+class EqualOperator(Operator):
```
achoum, line 24: To differentiate it again an equal operator between two features, could we find a more precise name?

temporian/implementation/numpy/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+        def are_same_general_type(dtype1, dtype2) -> bool:
+            """Check if two dtypes are of the same general type."""
+
+            # convert temporian dtypes to numpy dtypes
+            if dtype1 in DTYPE_REVERSE_MAPPING:
+                dtype1 = np.dtype(DTYPE_REVERSE_MAPPING[dtype1])
+            if dtype2 in DTYPE_REVERSE_MAPPING:
+                dtype2 = np.dtype(DTYPE_REVERSE_MAPPING[dtype2])
+
+            if dtype1.kind == "i" and dtype2.kind == "i":  # Both are integers
```
achoum, line 29: Nit: Add todo to use the "is_integer" method once PR#86 is submitted.

temporian/implementation/numpy/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class EqualNumpyImplementation(OperatorImplementation):
+    """New event with booleans where features are equal to a value."""
+
+    def __init__(self, operator: EqualOperator) -> None:
+        super().__init__(operator)
+
+    def __call__(self, event: NumpyEvent) -> Dict[str, NumpyEvent]:
+        def are_same_general_type(dtype1, dtype2) -> bool:
```
achoum, line 20: Could this check be made in the operator definition (!=operator implementation) in such a way that an error is raised if the two types are not compatible (instead of always returning false)?

temporian/implementation/numpy/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+
+        for index_value, features in event.data.items():
+            equal_features = []
+            for feature in features:
+                # compare general dtypes. all ints, all floats, all strings.
+                if are_same_general_type(feature.dtype, value_dtype):
+                    equal_feature = np.equal(feature.data, value)
+                else:
+                    equal_feature = np.full(
+                        feature.data.shape, False, dtype=np.bool_
```
achoum, line 66: This would be dangerous if someone does not realize that two types are not compatible.

achoum (commented):
Looks good. Left a few comments.

temporian/core/operators/boolean/base_feature.py
```diff
@@ -0,0 +1,114 @@
...
+    def operator_def_key(cls) -> str:
+        """Get the key of the operator definition."""
+
+    def feature_name(self, feature: Feature) -> str:
+        """Returns the name of the feature to be created."""
+        return f"{feature.name}_{self.operation_name}_{self.value_name}"
+
+    @property
+    @abstractmethod
+    def operation_name(self) -> str:
```
achoum, line 113: Nit: Since this is only used for the creation of the new feature name, could this name of this property indicate this.

temporian/core/operators/equal.py
```diff
@@ -0,0 +1,81 @@
...
+            Feature(
+                name=f"{f.name}_equal_{value}",
+                dtype=BOOLEAN,
+                sampling=f.sampling,
+                creator=self,
+            )
+            for f in event.features
+        ]
+
+        output_sampling = event.sampling
```
achoum, line 45: NIt: No need to create a output_sampling variable.

temporian/implementation/numpy/operators/boolean/base.py
```diff
@@ -0,0 +1,84 @@
...
+        ), (
+            "Expected operator to be of type BaseBooleanScalarOperator or"
+            f" BaseBooleanFeatureOperator, got {type(operator)}"
+        )
+        super().__init__(operator)
+
+    def __call__(
+        self, event_1: NumpyEvent, event_2: NumpyEvent = None
+    ) -> Dict[str, NumpyEvent]:
+        # if operator is a Scalar Operator, value is the scalar
```
achoum, line 36: I would separate the scalar and feature in two different implementation.

Note that the feature implementation could derive the arithmetic base class.
Instead, what about creating an arithmetic_scalar base class? This way, we can then easily implement operator such as "tp.add(a["x"], 5).

achoum (commented):
This is a great start. I left comments about the high level code structure.

PR #82, opened 2023-04-05
Docstrings

ianspektor:
- [X] Write docstring guidelines
- [X] Write example docstring (core lag & leak operators)
- [x] Write all remaining docstrings for public API
- [X] Do some work towards improving existing docstrings in internal classes and functions

Tried to strike some balance between not being too opinionated but not leaving too much space for creativity either, since it's consistency we're after 🤓 eager to get feedback

PR #89, opened 2023-04-14
Filter Operator

diegomarvid:
Add a new filter operator that filters an event base on a condition. The condition is an event with a boolean feature. The event timestamps where the condition is False will be filtered.

>__Note__: Currently I check that both `Event` and `Condition` must have the same sampling. This means that both should have the same index as well. Do you all agree with this design decision?

Example usage

```python
import pandas as pd
import temporian as tp


event_df = pd.DataFrame(
            [
                [1.0, 10.0],
                [2.0, np.nan],
                [3.0, 12.0],
                [4.0, 13.0],
                [5.0, 14.0],
                [6.0, 15.0],
            ],
            columns=["timestamp", "sales"],
        )

event_data = tp.from_dataframe(event_df)
event = event_data.schema()

condition_df = pd.DataFrame(
            [
                [1.0, True],
                [2.0, True],
                [3.0, True],
                [4.0, False],
                [5.0, False],
                [6.0, False],
            ],
            columns=["timestamp", "low_sales"],
        )

condition_data = tp.from_dataframe(condition_df)
condition = condition_data.schema()

filtered_event = tp.filter_event(event, condition)

expected_df = pd.DataFrame(
            [
                [1.0, 10.0],
                [2.0, np.nan],
                [3.0, 12.0],
            ],
            columns=["timestamp", "sales_filtered_by_low_sales"],
        )

expected_event = tp.from_dataframe(expected_df)

assert filtered_event == expected_event

```


achoum:
> Currently I check that both Event and Condition must have the same sampling. This means that both should have the same index as well. Do you all agree with this design decision?

Yes.

temporian/core/data/event.py
```diff
@@ -89,6 +89,10 @@ def __floordiv__(self, other):
     def sampling(self) -> Sampling:
         return self._sampling
 
+    @sampling.setter
+    def sampling(self, sampling: Sampling):
```
achoum, line 93: If this setter is mostly used to create event in several statement (my opposition to the modification of an existing event), would be make more sense not to have a setting, and to ensure the sampling is given in the constructor?

temporian/core/operators/filter.py
```diff
@@ -0,0 +1,118 @@
...
+operator_lib.register_operator(FilterOperator)
+
+
+# Added `_event` to the function name to avoid name collision with the
+# `filter` function in the `builtins` module.
+def filter_event(
+    event: Event,
+    condition: Event,
+) -> Event:
+    """Filter operator.
```
achoum, line 107: Nit: Should this be conjugated to the 3rd person?

For example, ""Filters out events for which the condition is false."""

https://github.com/google/temporian/pull/82/files#diff-ad7254a7a0a912467b0cc68a46f2f6f5560e26c992d14372c4bf7c5b85275795

temporian/core/operators/filter.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def feature_name(self, feature: Feature) -> str:
+        return f"{feature.name}_filtered_by_{self.condition_name}"
+
+
+operator_lib.register_operator(FilterOperator)
+
+
+# Added `_event` to the function name to avoid name collision with the
```
achoum, line 101: This is a good point. Like for the "glue" operator, can you think if there would be candidate short & single-word operator name instead?

temporian/implementation/numpy/operators/filter_event.py
```diff
@@ -0,0 +1,57 @@
...
+        event_filtered_data = {}
+
+        for index, features in condition.data.items():
+            # get boolean mask from condition
+            condition_feature = features[0].data
+
+            # filtered sampling
+            filtered_sampling = event.sampling.data[index][condition_feature]
+
+            # if filtered sampling is empty, skip
```
achoum, line 30: This is a good point. Modeling-wise, do you think returning an empty sampling would be better or worst?

achoum (approved):
Awesome

temporian/implementation/numpy/operators/filter_event.py
diegomarvid: If I return an empty sampling we cannot have the following functionality:

```python
event_df = pd.DataFrame(
            [
                [1.0, 10.0, "A"],
                [2.0, np.nan, "A"],
                [3.0, 12.0, "B"],
                [4.0, 13.0, "B"],
                [5.0, 14.0, "C"],
                [6.0, 15.0, "C"],
            ],
            columns=["timestamp", "sales", "product"],
        )

condition_df = pd.DataFrame(
            [
                [1.0, True, "A"],
                [2.0, True, "A"],
                [3.0, True, "B"],
                [4.0, False, "B"],
                [5.0, False, "C"],
                [6.0, False, "C"],
            ],
            columns=["timestamp", "low_sales", "product"],
        )

expected_df = pd.DataFrame(
            [
                ["A", 1.0, 10.0],
                ["A", 2.0, np.nan],
                ["B", 3.0, 12.0],
            ],
            columns=[
                "product",
                "timestamp",
                "sales_filtered_by_low_sales",
            ],
        )
```

We would probably have to add rows for each sampling in the DF and not add any values on the features. I think it makes it a bit more counterintuitive.

temporian/core/operators/filter.py
diegomarvid: what about `tp.pick(event, condition)`?

It could confuse with the select operator maybe. I don't think people will use tp.select if they have the "[]" option but it's still a valid counter argument.

diegomarvid: Pandas has a [`filter()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.filter.html) function for a DataFrame. If it's object-oriented we shouldn't have a problem I don't know if there's a way around this problem if we are using the functional mode (tp.filter()).

temporian/implementation/numpy/operators/filter_event.py
getchebarne: I prefer the alternative of removing the index level (what is currently implemented).

temporian/core/operators/filter.py
```diff
@@ -0,0 +1,116 @@
...
+            attributes=[],
+            inputs=[
+                pb.OperatorDef.Input(key="event"),
+                pb.OperatorDef.Input(key="condition"),
+            ],
+            outputs=[pb.OperatorDef.Output(key="event")],
+        )
+
+    def feature_name(self, feature: Feature) -> str:
+        return f"{feature.name}_filtered_by_{self.condition_name}"
```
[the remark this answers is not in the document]
diegomarvid: @achoum @DonBraulio

PR #91, opened 2023-04-14
Cast operator

### LINKED ISSUE #90: Cast operator

DonBraulio:
This PR implements the cast operator.
* `cast(event, target)` maps to `CastOperator(event, from_dtypes, from_features, to_dtype)`, only allowing one of them at the same time.
* Perform overflow checks by default, but don't ever check from/to boolean or string types  (see [comment](https://github.com/google/temporian/pull/91/files#diff-954ab77f3882c4971a5048544ce0a04cfb17166371da6f3fe2c617c8f6e9a50cR30))
* Patch to add `BOOLEAN` dtype so that I can check that cast too (it's also included in another PR).
* Added cast operator to the benchmark, to see overhead of overflow check (`float64,float64->float32,int32`):

```
================================================================
Name                              Wall time (s)    CPU time (s)
================================================================
...
----------------------------------------------------------------
simple_moving_average:100            0.00074       0.00074
simple_moving_average:10000          0.00109       0.00109
simple_moving_average:1000000        0.00921       0.00921
----------------------------------------------------------------
select_and_glue:100                  0.00035       0.00035
select_and_glue:10000                0.00037       0.00037
select_and_glue:1000000              0.00037       0.00037
----------------------------------------------------------------
calendar_day_of_month:100            0.00010       0.00010
calendar_day_of_month:10000          0.00435       0.00435
calendar_day_of_month:1000000        0.42776       0.42776
----------------------------------------------------------------
sample:e100_s100                     0.00052       0.00052
...
sample:e1000000_s1000000             0.01291       0.01167
----------------------------------------------------------------
propagate:100                        0.00050       0.00049
propagate:10000                      0.00550       0.00550
propagate:1000000                    0.40087       0.40087
----------------------------------------------------------------
cast(check=False):100                0.00026       0.00026
cast(check=True):100                 0.00084       0.00084
cast(check=False):1000000            0.00126       0.00126
cast(check=True):1000000             0.00372       0.00372
================================================================
```

Closes #90


temporian/core/operators/cast.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__()
+
+        # Check that all origin keys are DType or feature_names
+        if not isinstance(to, DType):
+            for origin_key in to:
+                if (
+                    not isinstance(origin_key, DType)
+                    and origin_key not in event.feature_names
+                ):
+                    raise ValueError(f"Invalid key to cast: {origin_key}")
```
diegomarvid, line 44: Could we raise a more explicit error?

```python
raise ValueError(f"Invalid key to cast: {origin_key}. Expected dtype or feature name.")
```

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,128 @@
...
+                    origin_key not in ALL_TYPES  # Also here
+                    and origin_key not in event.feature_names
+                ):
+                    raise ValueError(f"Invalid key to cast: {origin_key}")
+
+        # Convert any input format to feature_name->target_dtype
+        target_dtypes = {}
+        for feature in event.features:
+            if to in ALL_TYPES:  # isinstance(to, DType):
+                target_dtypes[feature.name] = to
```
diegomarvid, line 50: Could you add a short comment or description to understand each condition?

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,128 @@
...
+                pb.OperatorDef.Input(key="event"),
+            ],
+            outputs=[pb.OperatorDef.Output(key="event")],
+        )
+
+
+operator_lib.register_operator(CastOperator)
+
+
+def cast(
```
diegomarvid, line 123: Could we have a docstring here explaining the function? This is a Public API function.

temporian/implementation/numpy/operators/test/cast_test.py
```diff
@@ -0,0 +1,245 @@
...
+        self.assertTrue(self.numpy_expected_event == operator_output["event"])
+
+    def test_cast_mixed_keys(self) -> None:
+        """Test correct casting, some cols by origin dtype,
+        others by feature name, without overflow check.
+        """
+
+        operator = CastOperator(
+            event=self.input_event,
+            to={
```
diegomarvid, line 158: I just feel this may be confusing in the future or error-prone.

What do you think of the possibility of having two attributes on CastOperator:

1. `from_name` or `from_feature_name`
2. `from_dtype`

With this, we have two explicit dictionaries. It may make the code cleaner and it's more explicit for the user.

I'm open to suggestions or critics.

achoum:
>> Should I avoid reusing the feature/event instances, recreate all features and just reuse feature.data when possible?
>> If yes, also take the chance and add a cast_ prefix?

Changing the name of a feature makes it harder for the user to reference features by name, as the operator needs to know the new naming schema. I think not changing the name of the feature make sense in many situations, and the cast operator is one of them. As a counter example, it makes sense to generate a new feature name for arithmetic operators.


>> just reuse feature.data when possible
In the current implementation, reusing feature.data will avoid copy (which is great). However, it might not always be the case. When possible, when we want to avoid copy, I would always return the input feature (instead of creating a new feature on the existing feature.data).

diegomarvid:
If our value proposition is _"Safety, not error-prone, avoid common mistakes."_ I think we should check overflow by default. If we really don't want the computation overhead for this then we should at least give a warning every time the function is run without this flag in True.

DonBraulio:
> If our value proposition is "Safety, not error-prone, avoid common mistakes." I think we should check overflow by default. If we really don't want the computation overhead for this then we should at least give a warning every time the function is run without this flag set to True.

I agree, although I think it's indeed quite an overhead with the current implementation. But maybe it makes sense to prioritize safety over performance with the default parameters. I'll default to `True` and we can discuss it further tomorrow.

temporian/implementation/numpy/operators/test/cast_test.py
DonBraulio: Not sure about this one.
IMO the usage is more concise this way, and it's also pretty explicit. It would need some sort of precedence anyway with two separate dicts (if a feature matches feature_name and dtype), or raise error with two arguments at the same time. Also, `from_dtype` could be a dtype or a dict.

But I'm open to what the majority feels as more intuitive. Opinions @achoum @getchebarne ?

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,191 @@
...
+                    and origin_key not in event.feature_names
+                ):
+                    raise KeyError(
+                        f"Invalid key to cast: {origin_key}. "
+                        "Expected dtype or feature name."
+                    )
+
+        # Convert any input format to feature_name->target_dtype
+        target_dtypes = {}
+        for feature in event.features:
```
getchebarne, line 55: nit: can we move this logic to a separate `parse_target` method?

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+import temporian.core.data.dtype as tp_dtypes
```
getchebarne, line 2: move import to third party section

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,87 @@
...
+
+from typing import Dict
```
getchebarne, line 4: move import to standard section

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,87 @@
...
+class CastNumpyImplementation(OperatorImplementation):
+    def __init__(self, operator: CastOperator) -> None:
+        super().__init__(operator)
+        self.dtype_limits = {
+            tp_dtypes.INT32: np.iinfo(np.int32),
+            tp_dtypes.INT64: np.iinfo(np.int64),
+            tp_dtypes.FLOAT32: np.finfo(np.float32),
+            tp_dtypes.FLOAT64: np.finfo(np.float64),
+            tp_dtypes.STRING: np.finfo(np.float64),  # Use highest dtype
+        }
```
getchebarne, line 24: I think `dtype_limits` should be a class variable

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def _can_overflow(
+        self, origin_dtype: tp_dtypes.DType, dst_dtype: tp_dtypes.DType
+    ):
```
getchebarne, line 28: can we add dtypes here?

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,87 @@
...
+        return (
+            self.dtype_limits[origin_dtype].max
+            > self.dtype_limits[dst_dtype].max
+        )
+
+    def _check_overflow(self, data, origin_dtype, dst_dtype):
```
getchebarne, line 34: also here

temporian/implementation/numpy/operators/drop_index.py
```diff
@@ -160,3 +161,8 @@ def __call__(self, event: NumpyEvent) -> Dict[str, NumpyEvent]:
                 dst_event_data, NumpySampling(dst_index_names, dst_samp_data)
             )
         }
+
+
+implementation_lib.register_operator_implementation(
+    DropIndexOperator, DropIndexNumpyImplementation
```
getchebarne, line 167: thanks :))

temporian/proto/core.proto
```diff
@@ -157,7 +157,7 @@ message OperatorDef {
   }
 
   message Attribute {
-    // String identifier of the atrribute. Should be unique.
+    // String identifier of the atribute. Should be unique.
```
getchebarne, line 160: typo `atribute` -> `attribute`

getchebarne (approved):

temporian/proto/core.proto
DonBraulio: hehe amazing to fix a typo and generate a new one 🤦

temporian/implementation/numpy/operators/cast.py
DonBraulio: sorry for this one, got lazy and used auto import

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,105 @@
...
+        if np.any(
+            (data < self._dtype_limits[dst_dtype].min)
+            | (data > self._dtype_limits[dst_dtype].max)
+        ):
+            raise ValueError(
+                f"Overflow casting {origin_dtype}->{dst_dtype} {data=}"
+            )
+
+    def __call__(self, event: NumpyEvent) -> Dict[str, NumpyEvent]:
+        target_dtypes = self.operator.attributes["target_dtypes"]
```
achoum, line 58: Could those be made properties in the operator definition?

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,105 @@
...
+        if all(
+            orig_dtype is target_dtypes[feature_name]
+            for feature_name, orig_dtype in event.dtypes.items()
+        ):
+            return {"event": event}
+
+        # Create new event, some features may be reused
+        output = NumpyEvent(data={}, sampling=event.sampling)
+
+        for feat_idx, feature_name in enumerate(event.feature_names()):
```
achoum, line 71: Performance wise, it would be best to have the feature look inside of the index loop (instead of the opposite)

temporian/implementation/numpy/operators/BUILD
```diff
@@ -45,65 +48,69 @@ py_library(
 )
 
 py_library(
-    name = "glue",
-    srcs = ["glue.py"],
+    name = "cast",
```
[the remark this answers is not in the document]
DonBraulio: Yeah, I changed it to make it in alphabetic order so that it's easier to spot any missing test, but hadn't thought about that, going back.

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+class CastOperator(Operator):
+    """Type cast operator."""
+
+    def __init__(
+        self,
+        event: Event,
+        to_dtype: Optional[DType] = None,
+        from_dtypes: Optional[Mapping[DType, DType]] = None,
```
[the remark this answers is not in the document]
DonBraulio: But now that `DType` is an enum we shouldn't have this problem, right?
```
>>> a = DType.INT32
>>> isinstance(a, DType)
True
>>> type(a) is DType
True
```

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,105 @@
...
+        self._dtype_limits = {
+            DType.INT32: np.iinfo(np.int32),
+            DType.INT64: np.iinfo(np.int64),
+            DType.FLOAT32: np.finfo(np.float32),
+            DType.FLOAT64: np.finfo(np.float64),
+        }
+
+        self._nocheck_dtypes = [DType.BOOLEAN, DType.STRING]
+
+    def _can_overflow(self, origin_dtype: DType, dst_dtype: DType) -> bool:
```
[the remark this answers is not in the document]
DonBraulio: Anything that is not 0 is `True`:
```
>>> np.array([0, 1, -1, int(1e10), int(1e10) + 1]).astype(np.bool_)
array([False,  True,  True,  True,  True])
```

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,259 @@
...
+                " should be provided to CastOperator."
+            )
+
+        # Check: to_dtype is actually a dtype
+        if to_dtype is not None and not isinstance(to_dtype, DType):
+            raise ValueError(f"Cast: expected DType but got {type(to_dtype)=}")
+
+        # Check: from_dtypes is a dict {dtype: dtype}
+        if from_dtypes is not None and (
+            not isinstance(from_dtypes, Mapping)
```
diegomarvid, line 109: does `isinstance()` work with Mapping?

I know it doesn't work with something like this:

```python
if isinstance(x, Union[int, float])
```

temporian/implementation/numpy/operators/cast.py
DonBraulio: Ok this is interesting, because I thought so too and got a bit obsessed with it at this point, because the benchmarks show otherwise, consistently.

So, the main thing with the proposed version, is that you need to call `_can_overflow()` at each event_idx/feature. So I made a dict with `can_overflow` results pre-calculated, and made sure to have only one `event_index` lookup per row in this new implementation.
But the benchmark still favored the current version.

Then I thought that maybe adding 10 features, having these repeated event_idx lookups per feature would favor the proposed version, but the difference is even larger in favor of this current version:
```
for feature, for event (current), with 10 features
----------------------------------------------------------------
cast(check=False):100                0.00036       0.00036
cast(check=True):100                 0.00097       0.00097
cast(check=False):1000000            0.00115       0.00115
cast(check=True):1000000             0.00370       0.00370
================================================================

for event, for feature (proposed), with 10 features:
----------------------------------------------------------------
cast(check=False):100                0.00172       0.00172
cast(check=True):100                 0.00241       0.00241
cast(check=False):1000000            0.00349       0.00349
cast(check=True):1000000             0.00662       0.00644
================================================================
```

In summary, it seems like being able to factor out all these small dict lookups out of the second loop (`from_features[feature_name]`, `event.dtypes[feature_name]`, reverse dtype map, and check overflow), speeds up the code despite having to lookup by `event_index` for each feature. Which is consistent with the O(1) of the hashmap lookup, but still seems anti-intuitive.

Any thoughts?

temporian/core/operators/cast.py
DonBraulio: Yes 👍 
```
>>> a = {"hello": 1}
>>> isinstance(a, Mapping)
True
```

temporian/core/operators/cast.py
```diff
@@ -0,0 +1,259 @@
...
+                target_dtypes[feature.name] = to_dtype
+            elif from_features is not None:
+                # Cast by feature name (use same feature.dtype if not found)
+                target_dtypes[feature.name] = from_features.get(
+                    feature.name, feature.dtype
+                )
+            elif from_dtypes is not None:
+                # Cast by dtype (use same feature.dtype if not found)
+                target_dtypes[feature.name] = from_dtypes.get(
+                    feature.dtype, feature.dtype
```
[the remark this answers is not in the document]
DonBraulio: Had my doubts that it would be clearer, but now that I see it, it is 😄

temporian/implementation/numpy/operators/cast.py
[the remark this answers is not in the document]
achoum: Sg then

temporian/implementation/numpy/operators/cast.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __call__(self, event: NumpyEvent) -> Dict[str, NumpyEvent]:
+        target_dtypes = self.operator.attributes["target_dtypes"]
+        check = self.operator.attributes["check_overflow"]
+
+        # Reuse event if actually no features changed dtype
+        if all(
+            orig_dtype is target_dtypes[feature_name]
+            for feature_name, orig_dtype in event.dtypes.items()
+        ):
```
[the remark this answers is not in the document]
achoum: I think the second option is the right one, and we need to figure how to configure the pytype correctly.

Is adding "assert isintance(...)" before solve the typing issue?

temporian/implementation/numpy/operators/cast.py
achoum: Great analysis. Let's follow the benchmark :).

Possibly, you can add a TODO or NOTE for later re-visitation (i.e., maybe the EventV2 will change things).

achoum (approved):
Awesome

diegomarvid:
Great moment to send part of a conversation with ChatGPT:

image

PR #95, opened 2023-04-17
NumpyEvent V2

getchebarne:
This PR introduces the new `NumpyEvent`. This change deprecates the use of `NumpyFeature` and `NumpySampling`, as features and now represented by `np.ndarrays` and timestamps belong to the `NumpyEvent` itself. 

TODO: 
 - [x] reactivate `Sampling` checks in base NumPy Operator.


temporian/core/operators/propagate.py
```diff
@@ -40,7 +40,7 @@ def __init__(
 
         # TODO: Constraint the type of features supported in "add_event".
 
-        if event.sampling != to.sampling:
+        if event.sampling is not to.sampling:
```
diegomarvid, line 43: Why do we want to check that the reference is the same and not use the equals? Using the equals is better because we could have two different references with the same schema. Is there any reason for this change?

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+    DType.FLOAT32: np.float32,
+    DType.INT64: np.int64,
+    DType.INT32: np.int32,
+    DType.STRING: np.str_,
+    DType.BOOLEAN: np.bool_,
+}
+
+
+@dataclass
+class IndexData:
```
diegomarvid, line 48: Can we have a small docstring describing what this class is?

temporian/implementation/numpy/data/event.py
```diff
@@ -135,16 +202,12 @@ def from_dataframe(
             ... )
             >>> event = NumpyEvent.from_dataframe(df, index_names=["product_id"])
         """
-        if index_names is None:
```
diegomarvid, line 138: Actually, we have to leave the None in there. If we have a mutable object as a default value, then that object reference will be shared between different calls of the function and will end up with unexpected results. 

[https://web.archive.org/web/20201002220217/http://effbot.org/zone/default-values.htm ](https://web.archive.org/web/20201002220217/http://effbot.org/zone/default-values.htm )

temporian/implementation/numpy/data/event.py
```diff
@@ -199,144 +262,190 @@ def from_dataframe(
...
-        """Sequence of available indexes."""
+            # check same features
+            index_data_other = __o[index_key]
+            for feature_self, feature_other in zip(
+                index_data_self.features, index_data_other.features
+            ):
+                if feature_self.dtype.type != feature_other.dtype.type:
+                    return False
+
+                # check if the array has a float dtype. If so, compare with
```
diegomarvid, line 421: do we want event with different dtypes but same values to be the same?

diegomarvid (approved):
Great work!!

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+    ) -> None:
+        if isinstance(features, np.ndarray):
+            features = [features]
+
+        self.features = features
+        self.timestamps = timestamps
+
+    def __len__(self) -> int:
+        return len(self.timestamps)
+
```
DonBraulio, line 65: I'd really like to have the `__getitem__` magic here (returning a tuple (timestamp, row_features)), and also an `.item()` method that only works when `IndexData` has 1 element and raises ValueError otherwise (like [torch](https://pytorch.org/docs/stable/generated/torch.Tensor.item.html)). I'd imagine that in many use cases, there's only 1 row per index, and it's clumsy to do `timestamp, features = event[0][0]`.
We can discuss this further anyway if it's not trivial.

DonBraulio (approved):
I really like this change!
Some of my comments may need further discussion, feel free to ignore them and merge after fixing the easy ones 😄

temporian/core/operators/lag.py
```diff
@@ -36,6 +36,10 @@ def __init__(
     ):
         super().__init__()
 
+        self._duration = duration
+        self._prefix = "lag" if duration > 0 else "leak"
```
achoum, line 40: Should this be removed? (lag has not auto naming anymore

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+
+@dataclass
+class IndexData:
+    features: List[np.ndarray]
+    timestamps: np.ndarray
+
+    def __init__(
+        self,
+        features: Union[np.ndarray, List[np.ndarray]],
```
achoum, line 54: I would not allow for the user to feed a single np.array.

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+    # tolerance levels for comparing float features in __eq__
+    rtol: float = 1e-9
+    atol: float = 1e-9
+
     def __init__(
         self,
-        data: Dict[Tuple, List[NumpyFeature]],
-        sampling: NumpySampling,
+        data: Dict[Tuple, IndexData],
+        feature_names: Union[str, List[str]],
```
achoum, line 75: Same here (data and feature_names). I would only allow Lists.
Overloading argument types make the code harder to read.

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+        index_names: Union[str, List[str]],
+        is_unix_timestamp: bool = False,
```
achoum, line 77: Remove the default value so we cannot forget to set is_unix_time.

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+        self.features = features
+        self.timestamps = timestamps
+
+    def __len__(self) -> int:
+        return len(self.timestamps)
+
 
 class NumpyEvent:
+    # tolerance levels for comparing float features in __eq__
+    rtol: float = 1e-9
```
achoum, line 69: What about having a "Near" function instead, and having rtol and atol being argument of this "near" function?

temporian/implementation/numpy/data/event.py
achoum, line 75: Can we avoid duplicating the schema data? (feature name, index names, is_unix_timestamp)

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+    def index_names(self) -> List[str]:
+        return self._index_names
+
+    @property
+    def is_unix_timestamp(self) -> bool:
+        return self._is_unix_timestamp
+
+    @property
+    def _first_index_features(self) -> List[np.ndarray]:
+        return list(self._data.values())[0].features
```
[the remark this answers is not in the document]
achoum: I would go further. Since there is not guaranty that a feature exist or that there is an index (in both cases, this function will crash), I would remove the "_first_index_features" function and implement codes that currently call "_first_index_features" differently.

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
 
     @property
```
achoum, line 108: Non trivial time complexity => This should not be a property.

Also, it is not a great idea to duplicate data (higher memory consumption, more complex logic to make sure all the copies of the data remain in sync).

I think this should just be a function

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
 
-        return next(iter(self.data))
+    def iterindex(self) -> Iterable[Tuple[Tuple, IndexData]]:
+        yield from self.data.items()
 
-    def first_index_features(self) -> List[NumpyFeature]:
-        if self.first_index_value() is None:
-            return []
-        return self.data[self.first_index_value()]
+    def index_dtypes(self) -> Dict[str, DType]:
```
achoum, line 127: Can we have "dict" somewhere in the function name?

Do we really need all those accessors (iterator, extract dtype to dict, extract name to list, etc.). I think the code (both of this method and of the callers) would improve by having a single simple accessor.

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
-    def sampling(self, sampling: NumpySampling) -> None:
-        self._sampling = sampling
+    @property
+    def feature_count(self) -> int:
+        return len(self._feature_names)
 
-    def first_index_value(self) -> Optional[Tuple]:
-        if self.data is None or len(self.data) == 0:
-            return None
+    def copy(self) -> NumpyEvent:
```
achoum, line 121: I think se should not have any copier or setters in the API.

The library is purely functional (https://en.wikipedia.org/wiki/Purely_functional_programming) so modifiers and copier should not be necessary. 

If so currently needs it, can you add a TODO for removal?

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
-        # indexes should have the same names
-        return [feature.name for feature in self.first_index_features()]
+        return next(iter(self._data))
+
+    def first_index_features(self) -> IndexData:
+        if self.first_index_key() is None:
+            return []
+        return self[self.first_index_key()]
 
     def schema(self) -> Event:
```
achoum, line 150: The schema should not be recomputed from partial data. Instead, let's (1) keep a reference to the existing schema or (2) to pass both the data and schema to all the functions that require both.

achoum (commented):
Some initial comments :)

temporian/implementation/numpy/data/event.py
```diff
@@ -25,81 +25,148 @@
...
+@dataclass
+class IndexData:
+    features: List[np.ndarray]
+    timestamps: np.ndarray
+
+    def __init__(
+        self,
+        features: Union[np.ndarray, List[np.ndarray]],
+        timestamps: np.ndarray,
+    ) -> None:
```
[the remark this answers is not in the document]
getchebarne: I like it.

temporian/implementation/numpy/data/event.py
getchebarne: I don't think so

temporian/implementation/numpy/data/event.py
```diff
@@ -199,144 +262,190 @@ def from_dataframe(
...
         return (
-            f"data ({len(self.data)}):\n{data_repr}\nsampling:\n{sampling_repr}"
+            "data:"
+            f"\n\t\tindex_names={self.index_names}"
+            f"\n\t\tfeature_names={self.feature_names}"
+            f"\n{data_repr}"
         )
 
+    def __getitem__(self, index: Tuple) -> IndexData:
+        return self.data[index]
```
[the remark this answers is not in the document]
getchebarne: Good idea. We can discuss this further on tomorrow's sync.

temporian/implementation/numpy/data/event.py
getchebarne: why not? I think it can make the api less verbose (we parse features to a list in line 57, it won't be stored as a standalone array)

temporian/implementation/numpy/data/event.py
getchebarne: you mean as an operator or as an internal function to the numpy `event` module?

temporian/implementation/numpy/data/event.py
getchebarne: numpy events don't have a reference to their schema (e.g. when created from a DataFrame). Where would we get the schema data from?

temporian/implementation/numpy/data/event.py
getchebarne: I agree, it's a placeholder function (we need to improve index handling in the numpy backend in a future PR). I'll add a TODO as a reminder

temporian/implementation/numpy/data/event.py
DonBraulio: Just to clarify, I don't think this is important for the MVP, maybe it's one of those things that may come out from the UX testing. So feel free to stop reading this comment 😅 

So, in the general case, each `IndexData` actually contains a list of rows with `timestamp, features` pairs. So, for the general case we could access each row using `index_data[k]` instead of `index_data.timestamps[k]` + `index_data.features[k]`.
Aside of that, for the particular case where `index_data` only contains 1 row (i.e: the index is unique), there could be a shorthand method `index_data[0] -> index_data.item() ` which allows iteration over event rows like: `timestamp, features = event[p].item()`.

temporian/implementation/numpy/data/event.py
achoum: This is a nit, but here is my though process.
Keep in mind that this is your code, so you can disagree :).
When I disagree and I want to continue with the code submission, I generally just write "Ack" for acknowledged and close the conversation.

=============
Though process:

This function is not part of the API, so only is will use it. Therefore, code readability and validation is strictly more important than code ease of use. readability >> writing code.

Function with overloading arguments, especially when we start composing them, and especially when we don't have a type checker (yet), can hurt code readability and validation. For this reason, I generally try to avoid overloading arguments when not necessary.

In a way, this is similar to the op definitions. I always advocated for the op constructor not to have overloaded parameters, and instead, to keep overloading arguments strictly for the function API.

temporian/implementation/numpy/data/event.py
achoum: To were ever rtol is used. In other words, "rtol" should not be a property of a NumpyEvent.

ianspektor: agree with this. if rtol is used in `__eq__` only I wouldn't keep it as a class attribute. if parameters can't be added to `__eq__` which I think is the case, I would maybe do `rtol = None` by default in `__eq__` and have a `.equals` that the user can pass an rtol to if they need it?

temporian/implementation/numpy/data/test/event_test.py
```diff
@@ -2,47 +2,45 @@
...
+                    ],
+                    timestamps=np.array([0.1, 0.2, 0.3]),
+                ),
+                (2,): IndexData(
+                    features=[
+                        np.array([7, 8]),
+                        np.array([9, 10]),
+                    ],
+                    timestamps=np.array([0.4, 0.5]),
+                ),
```
ianspektor, line 26: love it 💕

temporian/implementation/numpy/data/event.py
getchebarne: I really like this idea. We can implement it in a future PR.

temporian/implementation/numpy/data/event.py
getchebarne: Very good explanation, I agree. Change's implemented.

PR #93, opened 2023-04-15
M5 example

achoum:
- Create an example of processing for the M5 dataset.
- Add "todos" for when all the necessary operators are available.
- Register the drop_index and set_index operators
- Simplify the drop_index operator implementation. Fix issues with string features / indexes with multiple characters.
- Allow plotting with multiple events.
- Create a "unique_sampling" operator.
- Modify the "propagate" operator. Now, the new index is defined by a "sampling" argument (instead of features). This version is significantly simpler to use and efficient on the M5 dataset.

PR #96, opened 2023-04-18
Optimize `from_df()`

diegomarvid:
Depends on #93 

- [x] Vectorized date conversion. -> 13x impact for 0 Indexes, negligible for multi-index.
- [x] Very slight improvement by changing apply to map for checking mixed types in object columns. -> Barely noticeable 
- [x]  Remove `get_group(group)`. Call `group[column_name]` directly. -> 2x impact with multiple indexes, doesn't apply to no index.

>__Warning__: Had to add np.int64 to PYTHON_DTYPE_MAPPING in order to pass the test after implementing the third point. Don't really understand the reason for this change yet.  Check [this commit](https://github.com/google/temporian/pull/96/commits/8857a8f9b9947b4a326f4b5e5184aa22a8f6b463).

Previous time:
```
================================================================
Name                              Wall time (s)    CPU time (s)
================================================================
from_dataframe:s:10000_numidx:0_numidxval:20_idx:int       0.00198       0.00542
from_dataframe:s:10000_numidx:0_numidxval:20_idx:str       0.00193       0.00193
from_dataframe:s:10000_numidx:1_numidxval:20_idx:int       0.00529       0.00528
from_dataframe:s:10000_numidx:1_numidxval:20_idx:str       0.00897       0.00897
from_dataframe:s:10000_numidx:3_numidxval:20_idx:int       0.69247       0.69235
from_dataframe:s:10000_numidx:3_numidxval:20_idx:str       1.14818       1.13795
from_dataframe:s:10000_numidx:5_numidxval:20_idx:int       1.28118       1.27048
from_dataframe:s:10000_numidx:5_numidxval:20_idx:str       2.31480       2.30958
```

New time:

```
================================================================
Name                              Wall time (s)    CPU time (s)
================================================================
from_dataframe:s:10000_numidx:0_numidxval:20_idx:int       0.00015       0.00015
from_dataframe:s:10000_numidx:0_numidxval:20_idx:str       0.00014       0.00014
from_dataframe:s:10000_numidx:1_numidxval:20_idx:int       0.00192       0.00192
from_dataframe:s:10000_numidx:1_numidxval:20_idx:str       0.00545       0.00545
from_dataframe:s:10000_numidx:3_numidxval:20_idx:int       0.36431       0.36395
from_dataframe:s:10000_numidx:3_numidxval:20_idx:str       0.55161       0.55124
from_dataframe:s:10000_numidx:5_numidxval:20_idx:int       0.63044       0.63038
from_dataframe:s:10000_numidx:5_numidxval:20_idx:str       1.14939       1.14907
```

Speed up:

```
================================================================
Name                              Wall time Speed up  CPU time Speed up
================================================================
from_dataframe:s:10000_numidx:0_numidxval:20_idx:int       13.2x           36.1x
from_dataframe:s:10000_numidx:0_numidxval:20_idx:str       13.8x           13.8x
from_dataframe:s:10000_numidx:1_numidxval:20_idx:int        2.8x            2.8x
from_dataframe:s:10000_numidx:1_numidxval:20_idx:str        1.6x            1.6x
from_dataframe:s:10000_numidx:3_numidxval:20_idx:int        1.9x            1.9x
from_dataframe:s:10000_numidx:3_numidxval:20_idx:str        2.1x            2.1x
from_dataframe:s:10000_numidx:5_numidxval:20_idx:int        2.0x            2.0x
from_dataframe:s:10000_numidx:5_numidxval:20_idx:str        2.0x            2.0x
```

We can see from the table that the most important aspect to optimize is the logic when there are multiple indexes. Grouping probably takes the most time.

>__Warning__: Couldnt run the profiler to profile from_dataframe.

### DIFF
--- a/temporian/implementation/numpy/data/event.py
+++ b/temporian/implementation/numpy/data/event.py
@@ -240,15 +240,22 @@ class NumpyEvent:
             Returns:
                 pd.Series: Converted timestamp column to Unix Epoch float.
             """
+            # check if timestamp column contains missing values and raise error
+            if timestamp_column.isna().any():
+                raise ValueError(
+                    f"Cannot convert timestamp column {timestamp_column.name} "
+                    "to Unix Epoch Float because it contains missing values."
+                )
+
             # if timestamp_column is already float64, ignore it
             if timestamp_column.dtype == "float64":
                 return timestamp_column
 
-            # if timestamp_column is int or float != float64, convert to float64
+            # if timestamp_column is int or float != float64 convert to float64
             if timestamp_column.dtype.kind in ("i", "f"):
                 return timestamp_column.astype("float64")
 
-            # string and objects will be converted to datetime and then to float
+            # string and objects will be converted to datetime, then to float
             timestamp_column = pd.to_datetime(timestamp_column, errors="raise")
             timestamp_column = timestamp_column.view("int64") / 1e9
             return timestamp_column
@@ -275,11 +282,8 @@ class NumpyEvent:
                 f"Timestamp column {timestamp_column} cannot be on index_names"
             )
 
-        # check if created sampling's values will be unix timestamps. #TODO:
-        # the user should also be able to specifiy wether it's a unix timestamp
-        is_unix_timestamp = pd.api.types.is_datetime64_any_dtype(
-            df[timestamp_column]
-        )
+        # check if created sampling's values will be unix timestamps
+        is_unix_timestamp = df[timestamp_column].dtype.kind not in ("i", "f")
 
         # convert timestamp column to Unix Epoch Float
         df[timestamp_column] = convert_timestamp_column_to_unix_epoch_float(
--- a/temporian/implementation/numpy/data/event.py
+++ b/temporian/implementation/numpy/data/event.py
@@ -258,9 +258,17 @@ class NumpyEvent:
         )
         # convert timestamp column to float
         # TODO: This is taking a lot of time. Don't use apply.
-        df[timestamp_column] = df[timestamp_column].apply(
-            convert_date_to_duration
-        )
+
+        # if timestamp column is of kind int convert to float
+        if df[timestamp_column].dtype.kind == "i":
+            df[timestamp_column] = df[timestamp_column].astype("float64")
+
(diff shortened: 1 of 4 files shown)


getchebarne (approved):
Very nice!

temporian/implementation/numpy/data/event.py
```diff
@@ -163,9 +172,18 @@ def from_dataframe(
         )
 
         # convert timestamp column to float
-        df[timestamp_column] = df[timestamp_column].apply(
-            convert_date_to_duration
-        )
+        # TODO: This is taking a lot of time. Don't use apply.
+
+        # if timestamp column is of kind int convert to float
```
achoum, line 177: Pandas can represent duration / timestamps with a type (e.g. int64) and a unit (e.g. s, µs, ns).
Are we sure this code works for all units?

achoum: Sure.

```python
import numpy as np
a = np.array([1490195805], dtype='datetime64[s]')
b = np.array([1490195805433502912]).astype('datetime64[ns]')

print(f"a:{a}")
print(f"b:{b}")

print(f"a.astype(int):{a.astype(int)}")
print(f"b.astype(int):{b.astype(int)}")
```

Results:

```
a:['2017-03-22T15:16:45']
b:['2017-03-22T15:16:45.433502912']
a.astype(int):[1490195805]
b.astype(int):[1490195805433502912]
```

Both "a" and "b", represent the same date. However, when casted to integer, the value is different because "a" represent the date in seconds since the epoche, while "b" is the time in ns since the epoche. When converting to temporian, we want all the dates to be a float64 in seconds.

I think the same exist in pandas types:

```python
a = pd.to_datetime(1490195805, unit='s')
b = pd.to_datetime(1490195805433502912, unit='ns')
```

diegomarvid: Okay so the user in the API will be able to pass any object as datetime and we will parse them with the function `pd,.to_datetime()` independent of what the user passes as an object (unless it's a plain int or float).

image

This is the output of what will happen if the user uses two datetimes with different units. In the end, both will have the same timestamp float but one will have more precision.

PR #97, opened 2023-04-19
Protobuf attributes & missing types

DonBraulio:
This PR is built on top of #91 (check only latest commit for diff) and adds some attribute and feature types that were missing or recently added:
* Added support and test for feature dtypes: `BOOLEAN` and `STRING`. Serialization for these types and also `FLOAT32` and `INT32` is tested by changing the output feature dtypes of operators `OpI102`, etc. Now all dtypes are covered.
* Added support and/or tests for attribute types: `BOOLEAN`,  `REPEATED_STRING` and `MAP_STR_STR`. Serialization of these attributes is done in a new case `test_serialize_attributes`.


temporian/proto/core.proto
```diff
@@ -71,6 +71,16 @@ message Operator {
       int64 integer_64 = 2;
       string str = 3;
       double float_64 = 4;
+      ListString list_str = 5;
```
getchebarne, line 74: I think we can use `repeated string` here

temporian/core/operators/base.py
```diff
@@ -135,6 +135,16 @@ def check(self) -> None:
                 if available_output not in [v.key for v in definition.outputs]:
                     raise ValueError(f'Unexpected output "{available_output}".')
 
+            # Check that expected attributes are present
+            for expected_attr in definition.attributes:
```
achoum, line 139: If possible, can we also check the type?

temporian/core/test/utils.py
```diff
@@ -200,3 +201,80 @@ def __init__(self, event: Event):
...
+                pb.OperatorDef.Input(key="event"),
+            ],
+            outputs=[
+                pb.OperatorDef.Output(key="output"),
+            ],
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="attr_int",
+                    type=pb.OperatorDef.Attribute.Type.INTEGER_64,
+                    is_optional=False,
```
achoum, line 221: Nit: is_optional is optional. The default value is False.

achoum (approved):
That's great :)

temporian/core/test/utils.py
DonBraulio: That's a self explanatory var name 😆

PR #98, opened 2023-04-21
Docs: setup mkdocs

MNLucas:
Initial setup for `temporian`'s documentation with a placeholder for the getting started guide and tutorials.

Preview: https://temporian.readthedocs.io/en/latest/


.readthedocs.yaml
achoum, line 1: Is it possible to move files in sub-directories to keep the top-level  from being encumbered.
For example, I would move all the read-the-doc files (both configuration and content) in the "docs"  directory.

achoum: In addition, could you document (or equivalently, create a script in "tools/") how to generate the documentation locally for preview?

achoum (approved):
Awesome :)

.readthedocs.yaml
ianspektor: added doc server instructions to readme

mkdocs.yml
```diff
@@ -0,0 +1,100 @@
...
+  - gen-files:
+      scripts:
+        # Generate the index page from the README.md.
+        - docs/gen_index.py
+        # Generate the reference pages from the docstrings.
+        - docs/gen_ref_pages.py
+  - literate-nav:
+      nav_file: SUMMARY.md
+    # Configuration to generate the reference pages from the docstrings.
+  - mkdocstrings:
```
diegomarvid, line 56: Great work! Is it viable to only show the docstrings/reference of the public functions/classes with which the users will interact? Currently, we have references for even the tests. I know there are ways to specify which functions/classes you want to show on the code. It may be too tedious to do this way but I don't know what alternative we have.

For more information we can check on this page: https://mkdocstrings.github.io/usage/

diegomarvid (approved):

mkdocs.yml
diegomarvid: Another thing to keep in mind is if these types of documentation are supposed to document all the code not only the public API functions/classes. If that is the case, you can ignore my comment.

diegomarvid: We could consider having separate sections: users and contributors. So users will not be exposed to the whole code base when searching for documentation. And still have the documentation necessary if anyone wants to understand better the library and contribute to it.

ianspektor: https://github.com/google/temporian/issues/104 opened this issue an hour ago stating we should remove test modules from docs 👍🏼 and agree that splitting public/internal ones could be a good idea too!

DonBraulio:
Monumental work 👏

PR #99, opened 2023-04-22
Example of notebook

achoum:
Example of notebooks:

- https://github.com/google/temporian/blob/gbm_notebook/example/toy_example.ipynb
- https://github.com/google/temporian/blob/gbm_notebook/example/vital_signs.ipynb


diegomarvid (approved):
Awesome notebook! I know this is a WIP PR but just wanted to leave a comment. When we plot an event with multiple features, if the name for each future is large then the y-axis label will overlap between different features.

For example here:

image

This could be easily fixed by not labeling the y-axis and setting a title for each plot. The y-label in that case could be changed to something shorter if we already have a title for an explanation. I know that the plot was automatically generated but we could rethink the function to avoid cases like this.

achoum:
I agree. Long feature names are an issue with this representation.
For this example, I'll simply use the "rename" op to create shorter features.

However, it would be interesting to brainstorm on other representations.
For instances, what about having the feature written horizontally like [here](https://framework.mathieu.guillame-bert.com/documentation/honey/pictures/ev.png)?

PR #101, opened 2023-04-24
Improve plotter

achoum:
- Index can be passed as a list of tuple, a tuple, a value, or a list of values.
- Automatically add markers in the plot when the sampling is not uniform.


getchebarne (approved):

DonBraulio (approved):
Nice one!

achoum:
Thanks

PR #105, opened 2023-04-27
Fixes & cleanup

DonBraulio:
Various fixes to errors I'm finding while working on an example (which is not ready but these fixes may be useful now):
 * Support to use `start_notebook.sh` in OSX, it was failing because `cp --parents ` option doesn't exist.
 * Benchmarks were failing because there was a very restrictive sampling check, which would fail if the samplings were expected to be different (because they don't reference the same `Sampling` object in core), but the timestamps actually were the same. I changed that because I find it very weird, and **only check that they samplings are equal when expected**. Please let me know if I'm missing something here.
 * Make all operators propagate the `is_unix_timestamp` property of sampling, which was mostly missing. Upon some discussion, this parameter is now **mandatory** in `Sampling()` too, as it already is in `NumpyEvent` (to avoid similar errors in the future).
 * Make sure that calls like `duration.weeks(4)` actually return a `float` and not `int` (since `Duration=float`)
 * Cleanup: Remove from numpy `feature.py` and `sampling.py` (no longer used in "V2"). Fixed BUILD files and some imports accordingly.


temporian/core/data/duration.py
```diff
@@ -35,23 +35,23 @@ def milliseconds(value: float) -> Duration:
 
 
 def seconds(value: float) -> Duration:
```
achoum, line 37: User will also want to feed values themselves. e.g.:

b = tp.moving_count(a, 5)


I think it would be better for Duration to be Union[float,int], and for attributes to do the conversion if necessary (as discussed during meeting).

tools/start_notebook.sh
```diff
@@ -21,8 +21,19 @@ bazel build -c opt //temporian
...
-( cd bazel-bin && find temporian \( -name "*.so" -o -name "*.py" \) -type f -exec cp --parents {} ${PKDIR}/ \; )
+find temporian -name "*.py" -type f -exec sh -c '
+    dir="$(dirname "$1")"
+    mkdir -p "${2}/${dir}"
+    cp "$1" "${2}/${dir}/$(basename "$1")"
+  ' _ {} ${PKDIR} \;
+
+( cd bazel-bin && \
+  find temporian \( -name "*.so" -o -name "*.py" \) -type f -exec sh -c '
+    dir="$(dirname "$1")"
```
achoum, line 32: This is complex :) I think rsync is available on Macos, can this be used instead?
Something like (I did not test):

( cd bazel-bin && \
find temporian \( -name "*.so" -o -name "*.py" \) -type f -exec rsync -R -arv {} ${PKDIR}/ \; )

achoum (approved):
Looks good. Let's wait for Ian code-wide PR (https://github.com/google/temporian/pull/111) to land first before merging this one?

PR #122, opened 2023-05-04
Rename arithmetic scalar files

ianspektor:


DonBraulio (commented):
I think it's easier to skip this PR because it collides with the major refactor and new operators I'm working on. I'm taking this criteria of adding the `_scalar` suffix there.

I like the warnings removal, let's not lose it:
```
    "missing-class-docstring",
    "missing-function-docstring",
```

PR #126, opened 2023-05-11
Getting started doc

ianspektor:
- Improved wording and syntax in Getting started document, and named it User Guide
- Updated and improved README. Sourced some stuff from our one pager.
- Decided against creating a separate Getting Started document - I think the README (which acts as our docs' index page) + User Guide covers it pretty well
- Removed `For developers` section from README when shown in docs index (filtered it out in `gen_index.py`)

- [x] Resolve TODOs
    - Left some missing graphs as TODOs 
- [x] Move intro to a separate `Getting Started` doc
- [x] Write `Memory management` section (though possibly not in this PR)
    - Not including this in this PR


DonBraulio (approved):
Awesome!
I think we should remove the TODOs and instead have a separate document (or even a task board), with things we'd like to improve or add to the documentation.
So that they can be discussed or addressed by anyone.

PR #127, opened 2023-05-12
Python types & tuple features

DonBraulio:
A few small things that were left behind:
- [x] Accept python types in `CastOperator`, `Feature`, `Sampling` and `input_event`.
- [x] Move `os.environ["DEBUG_MODE"]` to a centralized file (I put it on `utils/config.py`)x`
- [x] Specify features as tuples, like: `Node(features=[("f1", int), ("f2", float)])`


temporian/core/data/dtype.py
```diff
@@ -88,3 +88,32 @@ def from_python_type(cls, python_type: type) -> "DType":
             return DType.BOOLEAN
 
         raise ValueError(f"Non-implemented type {python_type}")
+
+
+def py_types_to_dtypes(target: Any) -> Any:
+    """
+    Convert any python types found to temporian types,
+    works recursively on List or Mapping.
```
ianspektor, line 96: Remove empty line at the start of docstring, use descriptive style:
```
"""Converts any...

"""
```

temporian/core/data/node.py
```diff
@@ -48,11 +48,19 @@ class Node(object):
 
     def __init__(
         self,
-        features: List[Feature],
+        features: List[Feature, FeatureTuple],
```
ianspektor, line 51: `List[Union[Feature, FeatureTuple]]`?

temporian/core/data/sampling.py
```diff
@@ -26,27 +26,30 @@
...
-    name: str
-    dtype: IndexDType
+AnyDType = Union[DType, IndexDType, type]
 
 
 class Index:
     def __init__(
         self,
-        levels: Union[Tuple[str, IndexDType], List[Tuple[str, IndexDType]]],
+        levels: Union[Tuple[str, AnyDType], List[Tuple[str, AnyDType]]],
```
ianspektor, line 52: an index's type cannot be a float, so using `AnyDType` is too permissive. If we need to support python dtypes we would need to allow `IndexDType` + [`int`, `str`]

ianspektor (approved):

PR #128, opened 2023-05-16
Add packaging configuration

rstz:
This PR adds tools for building a Python wheel of Temporian.

It includes the build script and also makes the following changes
- Bump Protobuf version past a breaking change in version 3.20
- Add documentation for builds
- Modify package description to be more conformant with Python's [Core metadata](https://packaging.python.org/en/latest/specifications/core-metadata/#author) (up for discussion)
- Remove unneeded setup.py


ianspektor (approved):

DonBraulio (commented):
Thanks for this 🙌 
I'm trying to build in macos arm64, but after a successful bazel compilation, I'm getting an error (image attached).

Apparently, it's because there's a `BUILD` file in the project root and poetry is also trying to create a file `build` (lowercase). The problem is that I'm on a case-insensitive file system (default for macos, almost sure 😬 ).

If there's an easy way to change that name, it would be great. Otherwise, we should probably clarify in the README that you need a case-sensitive filesystem to build.

Screen Shot 2023-05-17 at 11 27 23

ianspektor:
Confirm that build also works for intel macs, building in MacOS 13 with Rosetta (got same problem as @DonBraulio, deleted the BUILD file to patch and continue the process)

rstz:
The naming clash can be avoided either by deleting (but it might be useful to keep the license) or by renaming (see [Bazel docs](https://bazel.build/concepts/build-files)). Added a commit that renames BUILD to BUILD.bazel, which is equivalent

DonBraulio (approved):
Awesome! Worked now 🙌

PR #131, opened 2023-05-17
Publish to PyPI with GH Action

ianspektor:
- Added `publish.yaml` action which builds, tests and publishes wheels + source for all supported python versions for linux and macos
- Had to remove the usage of @classmethod, @property and @abstractmethod when used together (removed the property decorator) since py3.8 failed to understand it correctly
- Removed all uses of `list`, `dict`, `tuple` instead of `typing.List`, etc. as type hints for py3.8
- Had to make the `build_manylinux` script build for a single version, added `build_manylinux_all` which builds all of them
- Added all supported python versions to tests workflow
- Added manual confirmation step before publishing to pypi:
image

TODO:
- [x] add prod pypi token to `pypi` environment's secrets in the repo's settings
- [x] change `test-pypi` to `pypi` environment in publish step
- [x] change `branches:` to `release` in publish action
- [x] roll version back to `0.1.0`


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "temporian"
-version = "0.1.0"
+version = "0.1.3" # TODO: roll back to 0.1.0
 description = "Temporian is a library to pre-process temporal signals before their use as input features with off-the-shelf tabular machine learning libraries (e.g., TensorFlow Decision Forests)."
 authors = ["Google Inc."]
```
achoum, line 5: Nit: While you are at it, I would put the full team in the authors list.

achoum (approved):
That is great.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "temporian"
-version = "0.1.0"
+version = "0.1.3" # TODO: roll back to 0.1.0
```
rstz, line 3: Why not just move to 0.1.1 and release a new version with updated authors, maintainers etc?

tools/build.md
```diff
@@ -29,25 +29,39 @@ within the manylinux2014 container are encouraged.
...
-```
+
+```sh
 ./tools/start_compile_docker.sh
 ```
+
 Within the docker, run the following command
+
+```sh
+PYTHON_VERSION=<version> ./tools/build_manylinux_all.sh
```
rstz, line 40: Just build_manylinux.sh if you're specifying the python version.

tools/build.md
```diff
@@ -29,25 +29,39 @@ within the manylinux2014 container are encouraged.
...
 ```
-./tools/build_manylinux.sh
+
+where `<version>` is one of `38`, `39`, `310`, `311`.
+
+Or, in case you want to build for all supported Python versions
```
rstz, line 45: I've had weird caching issues with build_manylinux_all.sh end of last week, where bazel wound up building an incompatible package. I don't think it's worth diving deeper into it and instead just remove the build_manylinux_all.sh script since it does not provide that much extra value. WDYT?

rstz (approved):
Great work! Just a few minor things

rstz (approved):
Great work! Just a few minor things

tools/build.md
ianspektor: yep, typo, was supposed to be that one, good catch

tools/build.md
ianspektor: agreed! will do

pyproject.toml
ianspektor: I figured we could do that when we decided to do a new release - change the version name in main, then merge to release

DonBraulio (approved):
This is great 🙌🙌

rstz:
LGTM again, thank you for the changes :)

PR #133, opened 2023-05-19
Fix window accumulation

ianspektor:
Fix for tiny bug we discovered today. The `accumulate` function in `window.cc` didn't account for the possibility of there being equal timestamps in `evset_timestamps`, which caused two values in the same timestamp to end up with a different result in all window ops.

### DIFF
--- a/temporian/implementation/numpy_cc/operators/window.cc
+++ b/temporian/implementation/numpy_cc/operators/window.cc
@@ -34,27 +34,30 @@ py::array_t<OUTPUT> accumulate(const ArrayD &evset_timestamps,
 
   TAccumulator accumulator;
 
-  // Index of the first value in the RW.
+  // Index of the first value in the window.
   size_t begin_idx = 0;
+  // Index of the first value outside the window.
+  size_t end_idx = 0;
 
-  for (size_t sampling_idx = 0; sampling_idx < n_event; sampling_idx++) {
+  while (end_idx < n_event) {
     // Note: We accumulate values in (t-window_length, t] with t=
-    // v_timestamps[sampling_idx], and there may be several contiguous equal
+    // v_timestamps[end_idx], and there may be several contiguous equal
     // values in v_timestamps.
 
     // Add all values with same timestamp as the current one.
-    accumulator.Add(v_values[sampling_idx]);
-    size_t same_ts_idx = sampling_idx;
+    accumulator.Add(v_values[end_idx]);
+    const auto current_ts = v_timestamps[end_idx];
+    size_t first_diff_ts_idx = end_idx + 1;
     while (
-      same_ts_idx + 1 < n_event &&
-      v_timestamps[same_ts_idx] == v_timestamps[same_ts_idx + 1]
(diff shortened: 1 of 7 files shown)


temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -38,18 +38,36 @@ py::array_t<OUTPUT> accumulate(const ArrayD &evset_timestamps,
...
+    // Note: We accumulate values in (t-window_length, t] with t=
+    // v_timestamps[sampling_idx], and there may be several contiguous equal
+    // values in v_timestamps.
 
+    // Add all values with same timestamp as the current one.
     accumulator.Add(v_values[sampling_idx]);
+    size_t same_ts_idx = sampling_idx;
+    while (
+      same_ts_idx + 1 < n_event &&
+      v_timestamps[same_ts_idx] == v_timestamps[same_ts_idx + 1]
```
achoum, line 50: What about storing this value before the while loop on line 48. e.g.,

const auto ts = v_timestamps[same_ts_idx];
while (.... && ts == v_timestamps[same_ts_idx + 1]) { ...}

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -38,18 +38,36 @@ py::array_t<OUTPUT> accumulate(const ArrayD &evset_timestamps,
   size_t begin_idx = 0;
 
   for (size_t sampling_idx = 0; sampling_idx < n_event; sampling_idx++) {
+    // Note: We accumulate values in (t-window_length, t] with t=
+    // v_timestamps[sampling_idx], and there may be several contiguous equal
+    // values in v_timestamps.
 
+    // Add all values with same timestamp as the current one.
     accumulator.Add(v_values[sampling_idx]);
+    size_t same_ts_idx = sampling_idx;
```
achoum, line 47: I think the code below could be simpler if you store "sampling_idx+1" instead of "sampling_idx".

- No need for the +1 in the while loop.
- Need to do same_ts_idx++ after the .Add (no change in complexity).
- < instead of <= in for loop on line 65.
- Can replace the for loop on line 40 with a while loop (or a for loop without the "sampling_idx++" part).

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -38,18 +38,36 @@ py::array_t<OUTPUT> accumulate(const ArrayD &evset_timestamps,
   size_t begin_idx = 0;
 
   for (size_t sampling_idx = 0; sampling_idx < n_event; sampling_idx++) {
```
achoum, line 40: Nit: What about renaming "sampling_idx" with "end_idx" (more similar to begin_idx, more similar to the sampling based "accumulate" function.

achoum (approved):
Nice catch :)
That's is awesome.

temporian/implementation/numpy_cc/operators/window.cc
ianspektor: loved it!

PR #138, opened 2023-05-25
Added tools/test_install.py

DonBraulio:
Minimal test to use after installing temporian with pip.


tools/test_install.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
+""" Minimal script to test that temporian is installed and running well.
+
+Usage example:
+    pip install temporian
```
achoum, line 6: Nit: Add "-U" to make sure the last version of temporian is used.

PR #144, opened 2023-05-31
Add "memory usage" section in the __repr__ of an event set

achoum:
This PR is based on https://github.com/google/temporian/pull/141


temporian/implementation/numpy/data/event_set.py
```diff
@@ -498,3 +501,13 @@ def plot(self, *args, **wargs) -> Any:
         from temporian.implementation.numpy.data import plotter
 
         return plotter.plot(evsets=self, *args, **wargs)
+
+    def memory_usage(self) -> int:
```
rstz, line 505: How does this compare with (a recursive version of) sys.getsizeof() https://docs.python.org/3/library/sys.html#sys.getsizeof ?

rstz: Also: Maybe use __sizeof__ for it s.t. it can be used through sys.getsizeof()?

rstz (approved):

temporian/implementation/numpy/data/event_set.py
achoum: getsizeof  and __sizeof__ is surprisingly poorly documented, but I agree this is the right way to do.
For instance, surprisingly, sys.getsizeof of a dictionary grows with the size of the dictionary, but it does not in fact account for the content (key or values) of the dictionary :).

I've implemented it, and left the memory_usage function as an alias for discoverability.

PR #145, opened 2023-06-02
Dependencies and other cleanups

achoum:
This PR is based on https://github.com/google/temporian/pull/144


.vscode/settings.json
```diff
@@ -18,7 +18,7 @@
     "editor.defaultFormatter": "ms-python.black-formatter"
   },
   "editor.tabSize": 4,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
rstz, line 21: (not too familiar with vscode, but should this file be part of the repo at all?)

rstz (approved):

.vscode/settings.json
achoum: My version of code (possibly the same for you) has issues with black formatting, so I need to run it manually.
formatOneSave will messup files.

PR #149, opened 2023-06-07
tp.tick operator

achoum:
Based on https://github.com/google/temporian/pull/148 (tp.begin and tp.end operators)


temporian/core/operators/tick.py
```diff
@@ -0,0 +1,119 @@
...
+            inputs=[pb.OperatorDef.Input(key="input")],
+            outputs=[pb.OperatorDef.Output(key="output")],
+        )
+
+
+operator_lib.register_operator(Tick)
+
+
+def tick(input: Node, interval: Duration, rounding: bool = True) -> Node:
+    """Generates timestamps at regular intervaled in the range of a guide.
```
ianspektor, line 85: s/intervaled/intervals

temporian/core/operators/tick.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        input: Guide node. The start and end time boundaries to generate the new
+            timestamps are defined by the range of timestamps in `input`.
```
ianspektor, line 89: I think it could be useful (and easy to implement) to allow receiving two different nodes for the start and end dates? `tick(start: Node, end: Node, interval: Duration, ...`. 

Would be nice to have the second one be optional (in which case we use the first for both) but then we would need to have it as the third parameter after `interval` which might look weird

temporian/core/operators/tick.py
```diff
@@ -0,0 +1,119 @@
...
+        interval: Tick interval.
+        rounding: If false, the first tick is generated at the first timestamp
+            (similar to "tp.begin"). If true (default), ticks are generated on
+            timestamps that are multiple of "interval".
```
ianspektor, line 93: ` commas instead of " in "tp.begin" and "interval"

temporian/core/operators/tick.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def tick(input: Node, interval: Duration, rounding: bool = True) -> Node:
+    """Generates timestamps at regular intervaled in the range of a guide.
+
+    Args:
+        input: Guide node. The start and end time boundaries to generate the new
+            timestamps are defined by the range of timestamps in `input`.
+        interval: Tick interval.
+        rounding: If false, the first tick is generated at the first timestamp
```
ianspektor, line 91: I think this parameter's name could be improved - `align` or something similar?

ianspektor (approved):

PR #151, opened 2023-06-08
Docstring tests

DonBraulio:
NOTE: Since all the examples in the user guide must work now, I commented out with `<!-- TODO -->` the `has_leak` operation and some `evaluate()` variants, they're not visible now.
 
- [x] Create tests for docstring examples (bazel)
- [x] Create tests for markdown examples (bazel)
- [x] Fix code examples in docstrings
- [x] Fix code examples in 3 minutes guide
- [x] (in progress) Fix all code examples in user_guide
- [x] Explain how to create runnable examples in `CONTRIBUTING` (include details on: ignore part of results, SKIP flag, execution context).


temporian/core/data/node.py
```diff
@@ -502,18 +502,19 @@ def input_node(
     Usage example:
 
         ```
-        # Without index
-        a = source_node(features=[("f1", tp.float64), ("f2", tp.string)])
+        >>> # Without index
+        >>> a = input_node(features=[("f1", tp.float64), ("f2", tp.str_)])
```
achoum, line 506: tp.input_node?

Or is "tp" already imported?

achoum (approved):
Nice

temporian/core/data/node.py
DonBraulio: good catch. `input_node()` works here because we're on the module that defines it, but I'd agree it's best to use `tp.input_node` for consistency.

ianspektor (approved):

PR #154, opened 2023-06-09
Improve reference

ianspektor:
# Preview

https://temporian--154.org.readthedocs.build/en/154/reference/

# TODOs
- [x] fix all paths to reference sections in other parts of the docs
- [x] use cross references https://mkdocstrings.github.io/usage/#cross-references
- [ ] ~Add links to edit the file in github~
    - this is only easy to do if doc paths map directly to files somehow, which ours don't
- [x] add cross links to the user guide
- [x] show types (tp.float32, etc) in docs
- [ ] ~add cross links to operator docstrings~
    - will do in follow up PR
- [x] solve problem rendering operators with same function and file names


achoum:
This is awesome. This is my new favourite page  :).

I have a few comments from looking at the preview. I am writing about them here.

1.
Nit.
In the "Source code in" sections. Is it possible to get a link to the file on Github?

2.
Can custom type annotations (e.g. tp.Schema) be clickable?
For example, could the text "Schema", in the section "schema() -> Schema" of the page "tp.Node", redirect to the tp.Schema page?

3.
Nit
Is it possible to remove the "temporian" bullet on the page "https://temporian--154.org.readthedocs.build/en/154/reference/".
This is the only bullet with this indentation.

4.
Nit
Can the functions in tp.duration be visible in the left menu (i.e., tp.duration is unfoldable)?

5.
Nit
"Io" -> "IO" in the left menu.

6.
Can the python usage example be coloured syntactically? Maybe we just need to have the tag ```python in the documentation?

7.
Nit
Is it possible to create "links" in the documentation?
For example, in "tp.EventSet" plot, we write "Plots the event set. See tp.plot for details.".
Is there a way to get "tp.plot" clickable?

achoum (approved):

PR #159, opened 2023-06-15
Reference homepage

ianspektor:
https://temporian--159.org.readthedocs.build/en/159/reference/

# Important

Having this page means we need to remember to change the one-liner there if we change the docstring of any symbol in the public API (for most symbols, but not all of them, I just copied the first line of the docstring). If this is too much work I can look into ways to automate it, but same as with today's conversation, I think doc automation will only get us so far and will become too rigid in the short-medium term. Look for example at how I grouped the calendar or binary operators into groups in the same row - that'd be very hard to automate with heuristics without seriously restricting our codebase's structure.,


docs/src/reference/index.md
```diff
@@ -1,5 +1,92 @@
...
+| [`tp.EventSet`][temporian.EventSet] | Container for actual temporal data.                                                            |
+| [`tp.Schema`][temporian.Schema]     | Description of the data inside a [`Node`][temporian.Node] or [`EventSet`][temporian.EventSet]. |
+
+## Functions
+
+| Symbol                                    | Description                                                                            |
+| ----------------------------------------- | -------------------------------------------------------------------------------------- |
+| [`tp.evaluate()`][temporian.evaluate]     | Evaluates [`Nodes`][temporian.Node] on [`EventSets`][temporian.EventSet].              |
+| [`tp.plot()`][temporian.plot]             | Plots [`EventSets`][temporian.EventSet].                                               |
+| [`tp.event_set()`][temporian.event_set]   | Creates an [`EventSet`][temporian.EventSet] from arrays (list, numpy, pandas).         |
```
DonBraulio, line 21: nit: remove `pandas` from here (although it might work with pandas `Series`, I think it's more confusing than useful given `from_pandas`).

DonBraulio (approved):
Beautifullll 🤩 🙌

docs/src/reference/index.md
```diff
@@ -1,5 +1,92 @@
...
+| [`tp.unique_timestamps()`][temporian.unique_timestamps] | Removes events with duplicated timestamps from a [`Node`][temporian.Node].                  |
+
+### Binary operators
+
+| Symbols                                                                                                                                                                                                                                           | Description                                                                   |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| [`tp.add()`][temporian.add] [`tp.subtract()`][temporian.subtract] [`tp.multiply()`][temporian.multiply] [`tp.divide()`][temporian.divide] [`tp.floordiv()`][temporian.floordiv] [`tp.modulo()`][temporian.modulo] [`tp.power()`][temporian.power] | Compute an arithmetic binary operation between two [`Nodes`][temporian.Node]. |
```
DonBraulio, line 65: nit: Add a line like:

Aliases for `+-*/`, etc.

I'm writing some examples there to cover cast, resample, etc, but I feel like it's hard to find if you're using arithmetic symbols instead.

PR #160, opened 2023-06-16
Beam backend - IO utility + One op.

achoum:
This is the first step of the proposed beam backend. This PR introduce the following:
- A beam module "temporian.beam" containing the beam inference and IO utilities.
- The implementation of three operators (moving_sum, select and add_index).
- Testing utilities to ensure equality of the beam and "numpy" results.
- Proposal for the internal implementation of the Beam backend.

Some of the next steps after this PR include:
- Large scale tests.
- Implementation of all the operators.
- Utility to reduce the boiler plate code to implement horizontal operators.
- (Possibly) refinement of the API.


temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+    ]
+    feature_values = [
+        _cast_feature_value(row[feature.name], feature.dtype)
+        for feature in schema.features
+    ]
+    timestamp = np.float64(row[timestamp_key])
+    return tuple(index_values), (timestamp, tuple(feature_values))
+
+
+class _MergeTimestampsDP(beam.DoFn):
```
ianspektor, line 127: DP? a more expressive name + short docstring in the process method could help understand the merging logic

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+
+    This method is compatible with the output of `read_csv_raw` and the
+    Official Beam IO connectors.
+
+    When importing data from csv files, use `read_csv` to convert csv files
+    directly into event sets.
+
+    Unlike Temporian in-process event set import method (`tp.event_set`), this
+    method (`tpb.to_event_set`) requires for timestamps to be numerical values.
+    TODO: Add support for datetime timestamps.
```
ianspektor, line 163: move TODO to a normal comment, outside docstring

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+) -> beam.PCollection[BeamEventSet]:
+    """Converts a PCollection of key:value to a Beam event-set.
+
+    This method is compatible with the output of `read_csv_raw` and the
+    Official Beam IO connectors.
+
+    When importing data from csv files, use `read_csv` to convert csv files
+    directly into event sets.
+
+    Unlike Temporian in-process event set import method (`tp.event_set`), this
```
ianspektor, line 161: use cross-refs for navigable docs (in this case, s/`tp.event_set`/[`tp.event_set`][temporian.event_set])

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+
+    return (
+        pipe
+        | "Structure" >> beam.Map(_structure_fm, schema, timestamp_key)
+        # Group by index values and feature index
+        | "Aggregate" >> beam.GroupByKey()
+        # Build feature and timestamps arrays.
+        | "Merge timestamps"
+        >> beam.ParDo(_MergeTimestampsDP(len(schema.features)))
+    )
```
ianspektor, line 183: awesome

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+# Beam operations.
+StructuredRow = Tuple[
+    Tuple[BeamIndex, ...], Tuple[Timestamps, Tuple[np.generic, ...]]
+]
+
+# Unit of data for an event set used by all the operators implementations.
+#
+# Contains: the index+feature_idx, timestamps, and feature values.
+# The feature value can be None for event-set without features.
+BeamEventSet = Tuple[
```
ianspektor, line 49: If I understand correctly, this isn't a whole EventSet, but a single index value? In that case, I would propose changing its name to `IndexValue` or a better one, and a `BeamEventSet` or `PEventSet` would be a `PCollection[IndexValue]` for example

temporian/beam/operators/add_index.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+from typing import Dict, List, Tuple, Optional, Iterable
+
+from collections import defaultdict
+import apache_beam as beam
+import numpy as np
+
+from temporian.core.operators.add_index import (
+    AddIndexOperator as CurrentOperator,
```
ianspektor, line 23: why do this rename? the original symbol isn't used elsewhere

temporian/beam/io.py
ianspektor: I see now that it's not a whole index value, but a single feature inside an index value (my comment still applies). Also: does this mean the timestamps are repeated for every feature an event set has?

temporian/implementation/numpy/operators/base.py
```diff
@@ -37,6 +38,15 @@ def __call__(self, **inputs: EventSet) -> Dict[str, EventSet]:
     def output_schema(self, key: str) -> Schema:
         return self._operator.outputs[key].schema
 
+    def apply_feature_wise(
+        self,
+        src_timestamps: np.ndarray,
+        src_feature: np.ndarray,
+    ) -> np.ndarray:
+        """If implemented, the op can be executed feature wise."""
```
ianspektor, line 46: Executes the op on a single feature. Some operators might not implement this.

temporian/implementation/numpy/operators/window/base.py
```diff
@@ -90,3 +90,16 @@ def _compute(
                 kwargs["sampling_timestamps"] = sampling_timestamps
             dst_feature = implementation(**kwargs)
             dst_features.append(dst_feature)
+
+    def apply_feature_wise(
```
ianspektor, line 94: This looks like it can be merged with `_compute()` above

ianspektor (approved):
Went through the ops. Select and window look super clean, but I had a lot of trouble understanding how the add_index works, maybe just need to get used to the data structure inside the beam evsets. Some improvements in naming of the data structures and in docstrings might help. Amazing work as always, thank you!

temporian/beam/operators/add_index.py
achoum: Each such file handle a single operator.
There are some boilerplate code that only depends on the operator class.
By having a common class name, we make this boilerplate easier to read and type.

temporian/beam/evaluation.py
```diff
@@ -0,0 +1,161 @@
...
+        pipe: A Beam PCollection containing the input event set. Use
+            `tpb.read_csv` to read data from csv files, or use
+            `tpb.to_event_set` to import an event set from a dictionary of
+            key/values such as the output of Beam IO connectors
+            (https://beam.apache.org/documentation/io/connectors/).
+        input: Input node of a Temporian graph.
+        output: Output node of a Temporian graph.
+
+    Returns:
+        A Beam PCollection containing the output event set.
```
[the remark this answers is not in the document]
achoum: *Done

pyproject.toml
```diff
@@ -58,6 +58,7 @@ ipykernel = "^6.22.0"
 griffe = "0.26.0"
 jupyterlab = "^4.0.0"
 mkdocs-exclude-search = "^0.6.5"
+apache-beam = "^2.48.0"
```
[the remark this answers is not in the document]
achoum: Good point.
*Adding todo.

temporian/beam/implementation_lib.py
```diff
@@ -0,0 +1,46 @@
...
+            f"Unknown operator implementation '{key}' for Beam backend. It"
+            " seems this operator is only available for the in-process"
+            " Temporian backend. Available Beam operator implementations are:"
+            f" {list(_OPERATOR_IMPLEMENTATIONS.keys())}."
+        )
+    return _OPERATOR_IMPLEMENTATIONS[key]
+
+
+def registered_implementations() -> Dict[str, Any]:
+    """Lists the registered operator implementations."""
```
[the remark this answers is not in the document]
achoum: Good point.
The entire logic is common.
We could create a "registration" module containing this common logic.

*Added a todo.

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+# Type of an index element.
+#
+# In the numpy backend, index are represented as numpy primitives. However,
+# Beam does not support numpy primitive as index. Therefore, all index are
+# converted to python primitive of type "BeamIndex".
+BeamIndex = Union[int, float, str, bool]
+
+# Type of an index element OR a feature index (int).
+#
+# The feature index -1 is used to represent the timestamps of event-set without
```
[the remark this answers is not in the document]
achoum: `-1` is a value (it is not about "index-1").

*Rephrasing the documentation.

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+# features.
+BeamIndexAndFeature = BeamIndex
```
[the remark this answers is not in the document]
achoum: *Done

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+
+# An array of timestamps.
+Timestamps = np.float64
+
+# A single event / row during the conversion from dict of key/value to internal
+# the Temporian beam format for event sets. In a StructuredRow, index and
+# features are indexed by integer instead of string keys.
+#
+# Contains: the index, the timestamps, and the features.
```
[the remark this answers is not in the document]
achoum: *Done

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+# Note the double 2-items tuple (instead of a one 3-items tuple) to facilitate
+# Beam operations.
+StructuredRow = Tuple[
+    Tuple[BeamIndex, ...], Tuple[Timestamps, Tuple[np.generic, ...]]
+]
+
+# Unit of data for an event set used by all the operators implementations.
+#
+# Contains: the index+feature_idx, timestamps, and feature values.
+# The feature value can be None for event-set without features.
```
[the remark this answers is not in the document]
achoum: *Done

temporian/beam/io.py
achoum: This new naming is much better.

>> Duplicated timestamps.
Yes, this is an important implication. Essentially, we are traddingof performance for scalability.
*Added a comment about the duplication of timestamps.

temporian/beam/io.py
```diff
@@ -0,0 +1,299 @@
...
+    return np_type(value)
+
+
+def _cast_index_value(value: Any, dtype: DType) -> BeamIndex:
+    """Convert a user index value to the internal representation."""
+
+    return _cast_feature_value(value, dtype).item()
+
+
+def _structure_fm(
```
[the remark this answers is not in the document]
achoum: "feature map".
*Spelling it out.

temporian/beam/io.py
achoum: *Done

temporian/implementation/numpy/operators/base.py
achoum: *Done (update to keep under 80 chars).

temporian/implementation/numpy/operators/window/base.py
achoum: Almost.

- It would mean a new implementation class is instantiate for each feature (probably okay).
- "apply_feature_wise" does not support sampling. For the Beam implementation, sampling is a bit more complex as it requires two different streams of data.

achoum:
>  Went through the ops. Select and window look super clean, but I had a lot of trouble understanding how the add_index works, maybe just need to get used to the data structure inside the beam evsets. Some improvements in naming of the data structures and in docstrings might help. Amazing work as always, thank you!

Thanks, and that is a good point.
I am adding examples in the internal function documentation to improve understanding.

PR #161, opened 2023-06-16
Doc examples

DonBraulio:
Adding docstring examples to all public api methods.


ianspektor (approved):
Insane amount of work 👏🏼 Thanks!

temporian/core/data/duration.py
```diff
@@ -23,6 +23,26 @@
 def milliseconds(value: Union[int, float]) -> Duration:
     """Converts input value from milliseconds to a `Duration` in seconds.
 
+    Example:
```
achoum, line 26: Could you add two extra pieces of information:
- In Temporian, timestamps and durations are float64. The user chose the semantic of this value.
- In Temporian, datetime related functions (such as duration, calendar and plotting functions) assume that durations are expressed in seconds (https://temporian.readthedocs.io/en/latest/user_guide/#time-units). So it is a good idea for user to use second unit when it make sense.

temporian/core/data/duration.py
```diff
@@ -23,6 +23,26 @@
...
+        >>> evset = tp.event_set(
+        ...     timestamps=[0.5, 1.0, 1.2],
+        ...     features={"f1": [1, 5, -5]}
+        ... )
+        >>> source = evset.node()
+
+        >>> duration = tp.duration.milliseconds(250)
+        >>> duration
+        0.25
+        >>> result = tp.moving_sum(source, window_length=duration)
```
achoum, line 37: What about s/source/evset.node() here

temporian/core/data/duration.py
```diff
@@ -23,6 +23,26 @@
...
+        >>> result.evaluate({source: evset})
```
achoum, line 38: s/{source: evset}/evset

temporian/core/data/duration.py
```diff
@@ -39,6 +59,33 @@ def seconds(value: Union[int, float]) -> Duration:
...
+        >>> # Actually equivalent
+        >>> result = tp.moving_sum(source, window_length=3)
+
+        >>> result.evaluate({source: evset})
+        indexes: ...
+                timestamps: [1. 2. 6.]
+                'f1': [ 1 6 -5]
+        ...
+
+        ```
```
achoum, line 87: I would always use the same small example in all duration functions: print a duration and use a duration in a simple graph.

I would always use datetime timestamps.

temporian/core/evaluation.py
```diff
@@ -51,6 +51,60 @@ def evaluate(
 
     The result is returned in the same format as the `query` argument.
 
+
+    Single input output example:
+        ```python
+        >>> input_evset = tp.event_set(timestamps=[1, 2, 3], features={"f": [0, 4, 10]})
+        >>> input_node = input_evset.node()
+        >>> output_node = tp.moving_sum(input_node, 5)
```
achoum, line 59: I read the "# Equivalents" section as being equivalent to the bloc above. What about moving the "output_evset = tp.evaluate(" above the "# Equivalents"?

temporian/core/evaluation.py
```diff
@@ -51,6 +51,60 @@ def evaluate(
...
+
+        >>> # Equivalents
+        >>> output_evset = tp.evaluate(output_node, {input_node: input_evset})
```
achoum, line 62: Could you also show the version with `tp.evaluate(output_node, input_evset)`

temporian/core/evaluation.py
```diff
@@ -51,6 +51,60 @@ def evaluate(
...
+        >>> input_node = input_evset.node()
+
+        >>> # Check input name equals evset's name
+        >>> input_node.name
+        'my_input'
+
+        >>> output_node = tp.moving_sum(input_node, 5)
+
+        >>> # Find input node by name
+        >>> output_evset = output_node.evaluate(input_evset)
```
achoum, line 104: A name is not necessary for this version.

temporian/core/operators/add_index.py
```diff
@@ -143,22 +143,76 @@ def _normalize_index_to_set(
...
-
-        3. `add_index(input, feature_names=['X', 'Y'])`
-           Output `Node` will have index names ['A', 'B', 'C', 'X', 'Y'] and
-           features names ['Z'].
+    Usage example:
+        ```python
+        >>> a_evset = tp.event_set(
+        ...     timestamps=[1, 2, 1, 0, 1, 1],
+        ...     features={
+        ...         "store": [1, 1, 1, 2, 2, 2],
```
achoum, line 151: Ideally, all the example would use the same feature/indexes.
I am not sure using the store/product/sales actually helps understanding the example (unless we really use it everywhere). I would find it easier if features were just "f1, f2, ..." or "a", "b", "c". Ideally, all our examples would use the same implicit type of example.

What about something like:

evset = event_set(
    timestamps=[0.1, 0.2, 0.3, 0.4],
    features={
        "a": ["x", "x", "y", "y"],
        "b": [1, 1, 2, 3],
        "c": ["A", "B", "C", "D"],
        "d": [10, 11, 12, 13],
    }
)

tp.add_index(evset.node(), ["a", "b"])

or

evset = event_set(
    timestamps=[0.1, 0.2, 0.3, 0.4],
    features={
        "cat_1": ["x", "x", "y", "y"],
        "cat_2": ["A", "B", "C", "D"],
        "num_1": [1, 1, 2, 3],
        "num_2": [10, 11, 12, 13],
    }
)

achoum (approved):
This is awesome, and I am sure this will greatly help users :)

*Added some meta-nits

temporian/core/data/duration.py
ianspektor: adding to duration docs

PR #162, opened 2023-06-19
Redo reference

ianspektor:
# New docs structure
Docs now work as follows:
- The structure that the navigation follows is the structure of markdown files under `docs/src/reference`
- If a file there has content, the code ref generator doesn't touch it. Keep in mind you can still use mkdocstring identifiers (e.g. `::: temporian.to_csv`) for them to be auto-filled with that symbols' docstring. More info [here](https://mkdocstrings.github.io/usage/).
- If the file is empty, it is interpreted as a placeholder for the top-level symbol with its same name and its reference page is generated in its same path. E.g., if an empty `docs/src/reference/temporian/io/to_csv.md` file exists, the reference page for `temporian.to_csv` will be generated under `reference/temporian/io/to_csv/` in the docs.

# Tests
Added two tests:
- One that checks that the symbols available in `temporian.` are exactly the ones we expect
- One that checks that all of those symbols have an auto-generated page in the reference folder (so that we don't forget adding it when adding a new symbol)

# Bonus
- Managed to remove the `tp.core`, `tp.implementation`, etc. symbols by simply doing `del core` in the init file 👀
- Ordered the symbols in the left nav alphabetically + folders at the bottom


temporian/test/public_api_test.py
```diff
@@ -0,0 +1,142 @@
...
+        """Asserts that all symbols exposed under tp.<> have an auto generated
+        documentation page.
+
+        Note that this might need to change in the future, since for example we
+        might want to group some symbols under the same page."""
+        ref_path = Path("docs/src/reference")
+        ref_files = ref_path.rglob("*.md")
+        ref_files_names = {file.with_suffix("").name for file in ref_files}
+        for symbol in PUBLIC_API_SYMBOLS:
+            self.assertIn(symbol, ref_files_names)
```
DonBraulio, line 138: Should we also check that all `*.md` files have a corresponding symbol?

temporian/test/public_api_test.py
```diff
@@ -0,0 +1,142 @@
...
+    "notnan",
+    # WINDOW OPERATORS
+    "simple_moving_average",
+    "moving_standard_deviation",
+    "cumsum",
+    "moving_sum",
+    "moving_count",
+    "moving_min",
+    "moving_max",
+}
```
DonBraulio, line 118: Is there a way to test that all registered operators are exposed in the API?

docs/gen_ref_pages.py
```diff
@@ -1,111 +1,46 @@
 """
 Generate the code reference pages.
 
-Source: https://mkdocstrings.github.io/recipes/
+This script traverses the markdown files under docs/src/reference and for each:
+- If the file is not empty it is ignored and will appear in the docs as-is
```
DonBraulio, line 5: nit: `it is ignored` -> `it is used`

docs/gen_ref_pages.py
```diff
@@ -1,111 +1,46 @@
...
+    nav[list(nav_path.parts)] = ref_path.as_posix()
 
-while files_to_parse:
-    prefix, file = files_to_parse.pop()
+    # If file is empty we assume it's a top-level symbol and auto
+    # generate the mkdocstring identifier for it
+    if path.stat().st_size == 0:
+        with mkdocs_gen_files.open(full_ref_path, "w") as fd:
+            ident = "temporian." + nav_path.name
+            fd.write(f"::: {ident}")
```
DonBraulio, line 42: Are we using the file if it's not empty?
Can we have a print to show the list of non-empty files that have been found?

I'm mostly worried about potentially silent bugs generating some empty page.

DonBraulio (approved):
This reference is ✨✨✨

temporian/test/public_api_test.py
ianspektor: Nope, since we can have pages that aren't a symbol's (e.g. the index, or an operators home page, etc.)

temporian/test/public_api_test.py
ianspektor: Not that I can think of, since an operator has no reference to the API method that uses it

docs/gen_ref_pages.py
ianspektor: added print of non-empty files found

docs/gen_ref_pages.py
```diff
@@ -1,111 +1,58 @@
 """
 Generate the code reference pages.
```
achoum, line 2: What about adding a comment that tell that this script is executed when running `mkdocs serve -f docs/mkdocs.yml`?

docs/src/reference/index.md
```diff
@@ -18,7 +18,7 @@ Check the index on the left for a more detailed description of any symbol.
 | ----------------------------------------- | -------------------------------------------------------------------------------------- |
```
achoum, line 18: This front page is now awesome.

docs/gen_ref_pages.py
```diff
@@ -1,111 +1,58 @@
...
+    REFERENCE.rglob("*.md"), key=lambda p: str(len(p.parts)) + str(p)
+)
 
-non_parsable_imports = []
+non_empty_files = []
 
-# We need to be able to parse other files to allow wildcard imports
-# Storing pair of (prefix, path) to parse in a stack
-files_to_parse = [(None, SRC_PATH / "__init__.py")]
+for path in paths:
```
achoum, line 31: Would it be possible to check that all empty files point to a class / function ?

achoum (commented):
LGTM

docs/gen_ref_pages.py
ianspektor: mkdocstrings will fail if any of them doesn't 👍🏼

PR #164, opened 2023-06-20
Rename `index` to `indexes`

ianspektor:
Addresses action item from 01/06/23's meeting: renaming `index` to `indexes`, both in internal code and docs, so that we can easily talk about "indexes" and "features" and a single "index" and "feature",  for example. 

The renaming was already done in several key places (e.g. in `Node` and `Schema`).

Some other renamings (lmk if any of these don't make sense to you and we can discuss/rollback):
- `timestamps_column` -> `timestamps` in `tp.from_pandas()`, `tp.from_csv()` (normalized it with `tp.event_set()`)
- `index_names` and `index_columns` -> `indexes` in `tp.event_set()`, `tp.from_pandas()`, `tp.from_csv()`
- `get_arbitrary_index_value` -> `get_arbitrary_index_key` in `EventSet` since it returns a key
- `index_to_add`, `index_to_drop`, `index` -> `indexes` in `tp.add_index()`, `tp.drop_index()`, `tp.set_index()` respectively


DonBraulio (approved):
I like this change even tho I've to update my docstrings PR 😓 

TIL: indexes is ok as a plural form of index (as well as indices).

PR #165, opened 2023-06-23
hide pylint warnings for internal imports

ianspektor:
Pylint started marking all internal imports as missing since I manually removed the `core`, `implementation`, etc. symbols from the package, added it to ignores

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ disable = [
     "missing-function-docstring",
     "fixme",
     "invalid-name",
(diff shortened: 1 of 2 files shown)


ianspektor:
> Does it means the lint error is disabled for all the callers of temporian, or just for the temporian code?

just the temporian code 😄 the `[tool.XXX]` sections in the `pyproject.toml` are configuration sections for tools in our development environments

achoum (approved):
Sg then :)

PR #167, opened 2023-06-26
Eager mode + @tp.compile + tp.save(fn) + ...

ianspektor:
Implemented an `@operator` decorator that allows operators to receive `EventSets` in their `Node` args and return its evaluated result. 

Added a test to showcase the functionality with a couple of operators, will wait for review before extending to all other ones.

# Tasks

- [x] create @tp.compile and make it usable in any function
- [x] make @tp.compile accept receiving and returning tuples/lists/dicts of evsets
- [x] change all operator type hints to Generics + add @tp.compile decorator
- [x] rename `save` to `save_graph`, add new `save` that operates on a tp.function
- [x] extend `save` function to allow several EventSet params in signature
- [x] extend `save` function to allow returning dict of EventSets
- [x] add magic methods to EventSet
    - remove __setitem__ method
    - change __getitem__ for select op

# Further work

Not including these as part of this PR:
- [ ] update operator docstrings to talk about EventSets instead of Nodes
- [ ] update operator examples to use EventSets instead of Nodes
- [ ] allow calling `evset.operator()` or `node.operator()` (TBD)
- [ ] extend `save` function to allow receiving lists, dicts and tuples, and returning lists (behavior TBD)
- [ ] make @tp.compile Wrapper change Node type to EventSetOrNode, make operators and user functions be typed as Node


temporian/core/decorators.py
```diff
@@ -0,0 +1,86 @@
...
+            if node is not None:
+                kwargs[k] = node
+                inputs[node] = arg
+
+        output = fn(*args, **kwargs)
+
+        if is_eager:
+            evset = output.evaluate(inputs)
+            # Prevent .node() from creating a new sampling when called
+            evset._internal_node = output  # pylint: disable=protected-access
```
DonBraulio, line 47: Shouldn't we do this always inside of `evaluate()`?

achoum:
Ping me when you need the review.

temporian/core/serialization.py
```diff
@@ -313,14 +415,20 @@ def get_node(key):
...
         id=_identifier(src),
         sampling_id=_identifier(src.sampling_node),
         feature_ids=[_identifier(f) for f in src.feature_nodes],
         name=src.name,
-        creator_operator_id=_identifier_or_none(src.creator),
+        creator_operator_id=(
+            _identifier_or_none(src.creator)
+            if src.creator is not None and src.creator in operators
+            else None
+        ),
```
ianspektor, line 431: @achoum had to add these to remove creator operator references from input nodes to the graph. maybe it would make more sense to add it directly in the graph inference code? to discuss tomorrow. related to whether we want to "reset" an eventset's node's previous graph when calling .node()

temporian/core/serialization.py
```diff
@@ -38,11 +54,90 @@
...
+
+    node_inputs = {}
+    inputs = inputs.copy()
+    for k, v in inputs.items():
+        node_input = _process_fn_input(v)
+        if node_input is not None:
+            inputs[k] = node_input
+            node_inputs[k] = node_input
+
+    outputs = fn(**inputs)
```
achoum, line 105: Should we check exaustively the output type? For example, what is the returned value is a string, or a tuple with an EventSet and a string?

temporian/core/test/compilation_test.py
```diff
@@ -0,0 +1,99 @@
...
+# limitations under the License.
+
+from typing import Dict, List, Tuple
+from absl.testing import absltest
+
+import temporian as tp
+from temporian.implementation.numpy.data.event_set import EventSet
+
+
+class CompileTest(absltest.TestCase):
```
achoum, line 22: Could you also test the following situation? (to make sure a full and single graph is traced).
Note: To test that a graph is traced exactly once, you can instrument the traced functions.

1.
@tf.compile
def f(x)
    return tp.something(x)

@tf.compile
def g(x)
    return f(x) +x

2.
@tf.compile
def f(x, b:bool)
    if b:
        return tp.something(x)
    else:
        return tp.somethingelse(x)
3.
@tf.compile
def f(x, b:int)
    for _ in range(b):
        x = tp.something(x)
    return x

4.
(If possible)

class A:
    param:int = 5

    @tf.compile
    def f(x)
        return tp.something(x)

    @tf.compile
    def g(self, x)
        return tp.something(x, self.param)

temporian/core/test/compilation_test.py
```diff
@@ -0,0 +1,99 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, List, Tuple
+from absl.testing import absltest
+
+import temporian as tp
```
achoum, line 18: If possible, I would like to avoid calling this main API function in any of the files (including unit tests). 
If a unit test call "import temporian as tp", essentially this is not a unit test anymore (as it loads the entire temporian lib; possibly something very large).

temporian/core/compilation.py
```diff
@@ -0,0 +1,103 @@
...
+        " the node corresponding to each EventSet with .node(), or pass"
+        " EventSets only."
+    )
+
+    if isinstance(obj, EventSet):
+        if is_eager is None:
+            is_eager = True
+        elif not is_eager:
+            # If a Node had been received and we receive an EventSet, raise
+            raise ValueError(err)
```
achoum, line 91: It would be interresting to unit test all the failure modes.

temporian/test/api_test.py
```diff
@@ -60,6 +58,46 @@ def test_evaluation(self):
...
+        h2 = tp.resample(input=h1, sampling=evset_2)
+
+        # TODO: add magics to EventSet
+        # Create base class that both Node and EventSet inherit from
+        # h3 = evset_1 * 2.0 + 3.0 > 10.0
+
+        # TODO: enable __getitem__ for column selection in EventSet
+        # result = tp.glue(tp.prefix("sma_", h2["f2"]), evset_2)
+
+        result = tp.glue(tp.prefix("sma_", h2), evset_2)
```
achoum, line 82: Can you try to "del h1" before the computation of "result" (so it looks like a potential real code).

achoum (commented):
This is awesome :) !

Here is a first pass of comment.

docs/src/user_guide.md
```diff
@@ -1004,12 +1004,12 @@ a = evset.node()
...
+tp.save_graph(inputs={"input_a": a}, outputs={"output_b": b}, path="/tmp/my_graph.tem")
 
 # Equivalent.
 a.name = "input_a"
 b.name = "output_b"
-tp.save(inputs=a, outputs=[b], path="/tmp/my_graph.tem")
+tp.save_graph(inputs=a, outputs=[b], path="/tmp/my_graph.tem")
 
 # Load the graph.
 loaded_inputs, loaded_outputs = tp.load(path="/tmp/my_graph.tem")
```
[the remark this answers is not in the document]
ianspektor: adding to thursday's agenda

temporian/core/serialization.py
```diff
@@ -38,11 +54,90 @@
 
 
 def save(
+    fn: Callable[..., Union[EventSet, Dict[str, EventSet]]],
+    inputs: Dict[str, Any],
```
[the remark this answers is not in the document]
ianspektor: What would happen here?
```
def my_f(a: Node, b: Dict[str, Node]) -> Node:
    ...

a = event_set(...)
b = {"a": event_set(...)}
tp.save(my_f, inputs={"a": a, "b": b{)
```
I would need to check that the function parameter's names and dict's keys don't collide I think, which sounds undesirable. Taking to Thursday's meeting.

temporian/core/serialization.py
```diff
@@ -38,11 +54,90 @@
 
 
 def save(
+    fn: Callable[..., Union[EventSet, Dict[str, EventSet]]],
```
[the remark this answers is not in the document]
ianspektor: to discuss on thursday

temporian/core/serialization.py
```diff
@@ -38,11 +54,90 @@
...
+
+def _process_fn_input(input: Any) -> Optional[Node]:
+    # TODO: allow dicts and lists of stuff other than EventSet/Nodes
+    if isinstance(input, (list, dict, tuple)):
+        raise ValueError(
+            "`tp.save` does not support collections in a function's input."
+        )
+    if isinstance(input, Schema):
+        return create_node_with_new_reference(schema=input)
+    if isinstance(input, EventSet):
```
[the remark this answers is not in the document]
ianspektor: Why not?

temporian/core/serialization.py
```diff
@@ -38,11 +54,90 @@
...
+
+
+def _process_fn_input(input: Any) -> Optional[Node]:
+    # TODO: allow dicts and lists of stuff other than EventSet/Nodes
+    if isinstance(input, (list, dict, tuple)):
+        raise ValueError(
+            "`tp.save` does not support collections in a function's input."
+        )
+    if isinstance(input, Schema):
+        return create_node_with_new_reference(schema=input)
```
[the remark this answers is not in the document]
ianspektor: I don't see how input_node would serve me here since I only have the schema? And I want to create all new stuff since its an input node

temporian/core/serialization.py
```diff
@@ -110,9 +206,10 @@ def save(
 def load(
```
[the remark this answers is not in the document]
ianspektor: taking to thursday's

temporian/core/test/compilation_test.py
ianspektor: added TODO

achoum (approved):
Awesome

PR #169, opened 2023-06-27
Rename tp.evaluate to tp.run

achoum:


DonBraulio (approved):
LGTM 👍 

The notebooks are now being checked as part of the ci.
If these conflicts are complicated to solve, let me know and we can turn them off, merge as is and I can fix them in a following PR.

achoum:
Thanks for the approval.
It is not trivial to see the change that has been made in the ".ipynb" files.

+1 if you can resolve the conflicts in a next pr.
Let me push this one now, as the longer a code-width pr is is review, the harder it is to actually push it.

PR #173, opened 2023-06-28
Update 3_minutes.md

ktonthat:
Removed hyperlinks from headers


achoum (approved):
sg

DonBraulio (approved):
Ack - let's not use links in titles.

PR #176, opened 2023-06-29
Various small improvements

achoum:
- Print index key in plot
- Add support for equality / inequality in str_ features.
- Reduce title font size in plots.
- Unify defalut verbose value.
- tp.equal works both on node/node and node/scalar.
- Speedup tools/start_notebook.sh on Windows


temporian/implementation/numpy/operators/binary/relational.py
```diff
@@ -33,28 +35,43 @@ def __init__(self, operator: EqualOperator) -> None:
...
+        evset_2_feature: np.ndarray,
+        dtype: DType,
     ) -> np.ndarray:
-        # returns False on both NaNs
-        return np.equal(evset_1_feature.data, evset_2_feature.data)
+        if dtype == DType.STRING:
+            return np.char.equal(evset_1_feature.data, evset_2_feature.data)
+        else:
+            # returns False on both NaNs
+            return np.equal(evset_1_feature.data, evset_2_feature.data)
```
DonBraulio, line 47: Nice 👌

temporian/core/operators/binary/relational.py
```diff
@@ -118,6 +119,10 @@ def equal(
     Returns:
         Node containing the result of the comparison.
     """
+
+    if not isinstance(input_2, Node):
+        return equal_scalar(input=input_1, value=input_2)
```
DonBraulio, line 124: This is good.
I'm assuming we don't need this for the rest of the ops because nobody will use them directly.

DonBraulio (approved):
Nice 👍

temporian/core/operators/binary/relational.py
achoum: exactly my thought.

PR #178, opened 2023-07-02
Add tools/build_cleaner.py to fix BUILD dependencies

achoum:
Also some initial dep fix.


temporian/test/BUILD
```diff
@@ -66,16 +66,3 @@ py_test(
...
-    srcs = ["public_api_test.py"],
-    data = [
-        "//docs:reference",
-    ],
-    srcs_version = "PY3",
-    deps = [
-        # already_there/absl/testing:absltest
-        "//temporian",
-    ],
-)
```
ianspektor, line 81: 👀

achoum: I moved it in docs/BUILD (it makes my life a bit easier with the sync with internal version).

tools/build_cleaner.py
```diff
@@ -0,0 +1,463 @@
...
+        "//temporian/implementation/numpy_cc/operators:operators_cc",
+    ),
+}
+
+
+# Rule prefix for third party module dependencies.
+THIRD_PARTY_RULE_PREFIX = "# already_there/"
+
+# Filename of BUILD files.
+BUILD_FILENAME = "BUILD"
```
rstz, line 89: BUILD.bazel is an alternative name for the build file and we actually use it for the top-level build file due to a name collision when building the pip package on non case-sensitive systems. Maybe include it here

rstz (approved):
Very cool!

tools/build_cleaner.py
```diff
@@ -0,0 +1,463 @@
...
+This program does not modify BUILD files. Instead, it lists the dependencies to
+add and remove.
+
+How to use the build cleaner.
+
+    # Run the following command.
+    tools/build_cleaner.py
+
+If case of a false positive, you might have to update the configuration fields
+BUILD_IN_MODULES, THIRD_PARTY_MODULES, or EXTRA_SOURCE_TO_RULE.
```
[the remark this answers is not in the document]
achoum: *Done

tools/build_cleaner.py
achoum: Good point.
*Done

PR #179, opened 2023-07-03
Implement enumumerate() &  timestamps()

DonBraulio:


temporian/core/operators/timestamps.py
```diff
@@ -0,0 +1,107 @@
...
+    Datetime timestamps are converted to unix timestamps.
+
+    Integer timestamps example:
+        ```python
+        >>> from datetime import datetime
+        >>> evset = tp.event_set(
+        ...    timestamps=[1, 2, 3, 5],
+        ...    name='simple_timestamps'
+        ... )
+        >>> tp.timestamps(evset.node()).run(evset)
```
achoum, line 71: I like those examples.

Would it make sense to also create an example with an operation on the timestamps.

For example:

tp.timestamps(...) * 2
or  tp.timestamps(...) <= datetime("2020-01-02").timestamp()

temporian/implementation/numpy/operators/enumerate.py
```diff
@@ -0,0 +1,54 @@
...
+
+        output_schema = self.output_schema("output")
+
+        # Create output EventSet
+        output_evset = EventSet(data={}, schema=output_schema)
+
+        # Fill output EventSet's data
+        for index_key, index_data in input.data.items():
+            output_evset[index_key] = IndexData(
+                [np.indices(index_data.timestamps.shape).flatten()],
```
achoum, line 44: Would using numpy.arange be easier ?

https://numpy.org/doc/stable/reference/generated/numpy.arange.html

Nit: dtype=np.int64

achoum (commented):
Awesome :)

temporian/core/operators/timestamps.py
DonBraulio: I like it! I think the comparison is very useful.

DonBraulio:
Thanks! All suggestions implemented.

achoum (approved):
Awesome.
Looking forward to use those ops :).

Make sure to allign with Ian before submitting as he is working on a code-wide pr.

ianspektor:
Make sure to merge main and add @compile decorator to new ops before merging 👮🏼‍♂️

PR #180, opened 2023-07-03
Exec notebooks in RTD

DonBraulio:
- This PR moves the user_guide to a notebook.
- In order to test `user_guide.ipynb` as part of the CI, it needs to be executed in readthedocs (doctest not available for notebooks).
- Since its run in CI anyway, we can save them without outputs which makes it easier to merge.
- Took the chance and also run `getting_started.ipynb` with the same logic.

NOTE: The other notebooks are not being run or tested yet. The problem is that they download stuff and train models, which makes the CI setup very slow. I tried many workarounds but they add some complexity, so I want to discuss them on Thursday. Anyway, I'd merge this PR before solving that, to avoid blocking other tasks (e.g., update docs to eager mode).


achoum (approved):
That's is great.

ianspektor (approved):

PR #181, opened 2023-07-04
tp.has_leak function

achoum:


.vscode/settings.json
```diff
@@ -19,7 +19,7 @@
     "editor.defaultFormatter": "ms-python.black-formatter"
   },
   "editor.tabSize": 4,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
ianspektor, line 22: We need to stop changing this back and forth 😅 is the auto save causing trouble? Works fine on our end

.vscode/settings.json
```diff
@@ -82,5 +82,6 @@
     "queue": "cpp",
     "span": "cpp",
     "algorithm": "cpp"
-  }
+  },
+  "python.analysis.typeCheckingMode": "basic"
```
ianspektor, line 86: 🙌🏼 💯

temporian/core/evaluation.py
```diff
@@ -258,6 +258,59 @@ def build_schedule(
     return schedule
 
 
+def has_leak(
+    output: EvaluationQuery,
+    input: Optional[EvaluationQuery] = None,
+) -> bool:
+    """Tests if a node depends on a leak operator.
+
+    Tests if a [`Node`][temporian.Node] or collection of nodes depends on the
```
ianspektor, line 267: careful to change to EventSetNode when merging main

temporian/core/evaluation.py
```diff
@@ -258,6 +258,59 @@ def build_schedule(
     return schedule
 
 
+def has_leak(
+    output: EvaluationQuery,
```
ianspektor, line 262: EvaluationQuery seems to be creeping into the codebase, added TODO in my last PR to give it a better name

temporian/core/evaluation.py
```diff
@@ -258,6 +258,59 @@ def build_schedule(
...
+    input: Optional[EvaluationQuery] = None,
+) -> bool:
+    """Tests if a node depends on a leak operator.
+
+    Tests if a [`Node`][temporian.Node] or collection of nodes depends on the
+    only operator that can introduce a future leakage:
```
ianspektor, line 268: nit s/a future leakage/future leakage

temporian/core/evaluation.py
```diff
@@ -258,6 +258,59 @@ def build_schedule(
...
+
+    Single input output example:
+        ```python
+        >>> a = tp.input_node([("f", float)])
+        >>> b = tp.moving_sum(a, 5)
+        >>> c = tp.leak(b, 6)
+        >>> d = tp.prefix("my_prefix_", c)
+        >>> e = tp.moving_sum(d, 7)
+        >>> # The computation of "e" contains a leak.
+        >>> assert tp.has_leak(e)
```
ianspektor, line 279: I think it would read better in the examples to print the `True` or `False` results instead of assert them

temporian/core/evaluation.py
```diff
@@ -258,6 +258,59 @@ def build_schedule(
...
+        normalized_input = None
+    else:
+        normalized_input = _normalize_query(input)
+
+    normalized_output = _normalize_query(output)
+
+    graph = infer_graph(inputs=normalized_input, outputs=normalized_output)
+
+    for operator in graph.operators:
+        if operator.operator_key() == "LEAK":
```
ianspektor, line 308: maybe would make sense to define `"LEAK"` as a constant in leak.py and import it from here? to not have the hardcoded string in both places. or check the operator is an instance of the class?

ianspektor (approved):

.vscode/settings.json
achoum: For me, the issue is that BUILD files are python files but they are not currently formated as our other python files.
So each time I edit a BUILD file, the formatting is all changed.

Have you disabled the formating of BUILD files?

temporian/core/evaluation.py
achoum: sg

temporian/core/evaluation.py
achoum: *Replaced

temporian/core/evaluation.py
achoum: Good point.
*Use returned value

temporian/core/evaluation.py
achoum: Much better.
*Use LEAK constant defined in the leak operator definition

PR #182, opened 2023-07-04
rename Node to EventSetNode

ianspektor:


docs/src/reference/temporian/internals/core.mixins.EventSetOperationsMixin.md
ianspektor, line 1: Managed to include non-public symbols in the docs :) I think it's a nice way for the user to be able to see their docs, but not have it available under `tp.`. 

In this case for example its for showing the methods included in the `EventSetOperationsMixin`, so that the user can see what magics are implemented in the `EventSet` class (and the chainable operators will go here too).

achoum (approved):
Nice

PR #185, opened 2023-07-06
Scheduler optimization

achoum:
- Release unused memory during evaluation
- Deterministic op evaluation order


ianspektor (approved):

temporian/core/evaluation.py
```diff
@@ -208,7 +208,8 @@ def build_schedule(
     # Operators ready to be computed (i.e. ready to be added to "planned_ops")
     # as all their inputs are already computed by "planned_ops" or specified by
     # "inputs".
-    ready_ops: Set[Operator] = set()
+    ready_ops: List[Operator] = []
+    ready_ops_set: Set[Operator] = set()
```
DonBraulio, line 212: I think it's safe to remove `ready_ops_set` now we got `ready_ops`. I don't see its purpose.

temporian/core/evaluation.py
```diff
@@ -222,25 +223,50 @@ def build_schedule(
...
+            ready_ops.append(op)
+            ready_ops_set.add(op)
         else:
             # Some of the inputs are missing.
             op_to_num_pending_inputs[op] = num_pending_inputs
 
+    # Make evaluation order deterministic.
+    #
+    # Execute the op with smallest internal ordered id first.
+    ready_ops.sort(key=lambda op: op._internal_ordered_id, reverse=True)
```
DonBraulio, line 242: I'm not sure why the scheduling order is better for memory release with this step.
I get it only affects the ops that depend directly on the inputs (current ops in `ready_ops`) and executes them in instantiation order, but is that relevant in terms of memory allocation?

DonBraulio (approved):
LGTM.
Nice work on the memory release.
Left a comment because I don't understand why sorting only the first input ops is relevant.

temporian/core/operators/base.py
```diff
@@ -81,10 +83,18 @@ def __init__(self):
             attr.key: attr.type for attr in self._definition.attributes
         }
 
+        # Id of the operator object such that an operator A instantiated before
+        # an operator B will have a smaller _internal_ordered_id.
+        #
+        # _internal_ordered_id is used to ensure the deterministic graph
+        # evaluation.
+        self._internal_ordered_id = Operator.next_internal_id
+        Operator.next_internal_id += 1
```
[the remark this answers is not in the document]
achoum: The alternative solution would be to replace some of the set-inputs in the API and intermediate functions by lists. I don't see that as ideal.

The other benefit of this change is that "id" is now a small number that is easier for people to read in error messages. Ideally, I would like to update all the "ids" with this mechanism.

The third benefit is that an "id()" is not guaranteed to be unique for two objects with non-overlapping lifetimes. This is not an issue for our current code, but this is a property that is not great for what we use it for.

See https://docs.python.org/3/library/functions.html#id

temporian/core/evaluation.py
achoum: "ready_ops" and "ready_ops_set" contain the same data. Those two containers have different properties and costs (one is a list, the other is a set). The list is great for a FILO, while the set is great to check for presence of an item.

>> I think it's safe to remove ready_ops_set now we got ready_ops

You mean there is a situation where this code will not work? If so, can you give details?

achoum:
Thanks.
Submitting now.

Adding the open questions to the agenda of our next meeting.

PR #190, opened 2023-07-11
Internally, tp uses bytes instead of int32 unicodes

achoum:


rstz (approved):
lgtm

docs/src/reference/temporian/dtypes/bytes_.md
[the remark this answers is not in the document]
ianspektor: it's a placeholder for https://temporian--190.org.readthedocs.build/en/190/reference/temporian/dtypes/bytes_/ to be autogenerated by mkdocstrings

temporian/beam/io.py
```diff
@@ -1,6 +1,6 @@
 """Utilities to import/export Beam-Event-Set from/to dataset containers."""
 
-from typing import Iterable, Dict, Any, Tuple, Union, Optional
+from typing import Iterable, Dict, Any, Tuple, Union, Optional, List
```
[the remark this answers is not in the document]
ianspektor: unsupported in py 3.8 so we can't use them 🥲 had to modify them to the old syntax when making tests pass for all supported versions

temporian/core/data/dtype.py
```diff
@@ -124,5 +127,6 @@ def check_is_valid_index_dtype(dtype: DType):
 bool_ = DType.BOOLEAN
 """Boolean value."""
 
-str_ = DType.STRING
+bytes_ = DType.STRING
```
ianspektor, line 130: missing docstring

temporian/implementation/numpy/data/event_set.py
```diff
@@ -126,22 +145,22 @@ def normalize_features(
...
     if feature_values.dtype.type == np.string_:
-        feature_values = feature_values.astype(np.str_)
+        feature_values = feature_values.astype(np.bytes_)
 
     if feature_values.dtype.type == np.object_:
         logging.warning(
             (
                 'Feature "%s" is an array of numpy.object_ and was casted to'
-                " numpy.str_."
+                " numpy.string_ == numpy.bytes_."
```
ianspektor, line 159: would keep either one or the other, == looks like an expression

temporian/implementation/numpy/data/event_set.py
```diff
@@ -61,10 +61,29 @@
...
+IndexType = Tuple[IndexItemType, ...]
+
+
+def normalize_index_item(x: IndexItemType) -> IndexItemType:
+    if isinstance(x, str):
+        return x.encode()
+    return x
+
+
+def normalize_index(
```
ianspektor, line 78: would rename to normalize_index_key, since we're now using the word index to refer to an index column/index feature

ianspektor (approved):

PR #191, opened 2023-07-11
Implements add_index in cc

achoum:
- [ ]  cc implementation
- [ ] replace python implementation
- [ ] benchmark speed boost


temporian/implementation/numpy_cc/operators/add_index.cc
```diff
@@ -0,0 +1,256 @@
...
+    for (Idx row_idx = 0; row_idx < num_rows; row_idx++) {
+      const char *begin = ((char *)info.ptr) + row_idx * stride;
+      std::string_view value(begin, itemsize);
+      local_groups[remove_tailing_zeros(value)].push_back(row_idx);
+    }
+  } else {
+    for (const auto row_idx : selected_rows) {
+      const char *begin = ((char *)info.ptr) + row_idx * stride;
+      std::string_view value(begin, itemsize);
+      // DO NOT SUBMIT -> Remove extra 0s
```
DonBraulio, line 183: I guess this is intended to be removed

temporian/implementation/numpy/operators/add_index.py
```diff
@@ -32,24 +33,20 @@ def __call__(self, input: EventSet) -> Dict[str, EventSet]:
...
-            for new_index, example_idxs in new_index_to_value_idxs.items():
-                # Note: The new indexes added after the existing ones.
-                dst_index = src_index + new_index
+            for group_idx, group_key in enumerate(group_keys):
+                dst_index = src_index + group_key
                 assert isinstance(dst_index, tuple)
 
+                example_idxs = row_idxs[
+                    group_begin_idx[group_idx] : group_begin_idx[group_idx + 1]
+                ]
```
DonBraulio, line 49: The use of `index` (or `idx`) in this op is very confusing, as it refers to the temporian concept of index and also to array positions.
In the future we should consider alternative names for ordinal positions, like `pos`, `loc`, `ord`, `num`.
I think that `index` is better in any other context, but here it's ambiguous and also leads to these weird names like "example", I guess to avoid `index_idxs` (?)

DonBraulio (approved):
Awesome!
Feel free to ignore the comment about the ambiguity of `index` for this PR, it's just something to think about in the future.

temporian/implementation/numpy/operators/add_index.py
achoum: let's figure a way next meeting :).

So far, we are using "index" to talk about Temporian index, and "idx" to reference integers that define a position in a list/vector.

achoum:
Thanks!

PR #195, opened 2023-07-13
Verbose compile

ianspektor:


temporian/core/compilation.py
```diff
@@ -24,57 +24,103 @@
...
+) -> Any:
     """Compiles a Temporian function.
 
-    A Temporian function is a function that takes EventSetNodes as arguments and
-    returns EventSetNodes as outputs. Compiling it enables it to perform eager
-    evaluation, i.e., receive and return EventSets instead of EventSetNodes.
+    A Temporian function is a function that takes
+    [`EventSetNodes`][temporian.EventSetNode] as arguments and returns
+    [`EventSetNodes`][temporian.EventSetNode] as outputs. Compiling it enables
+    it to perform eager evaluation, i.e., receive and return
```
achoum, line 35: The other side of the coin is that with @compile, you run a function operating on tp.EventSet faster, since the operators run both on Node and EventSet. I wonder if this is also worth mentionning.

temporian/core/compilation.py
```diff
@@ -24,57 +24,103 @@
...
+    [`EventSets`][temporian.EventSet] instead of
+    [`EventSetNodes`][temporian.EventSetNode].
+
+    Example usage:
+    ```python
+    >>> @tp.compile
+    ... def f(x: EventSetNode) -> EventSetNode:
+    ...     return tp.prefix("pre_", x)
```
achoum, line 43: What about having two operators in the function? (e.g. prefix + moving_sum)

temporian/core/compilation.py
```diff
@@ -24,57 +24,103 @@
...
+            return outputs
+
+        wrapper.is_tp_compiled = True
 
-            return run(query=outputs, input=inputs_map)
+        return wrapper
 
-        return outputs
+    # Function is being called as a decorator
+    if fn is not None:
```
achoum, line 114: Fun

temporian/core/compilation.py
```diff
@@ -24,57 +24,103 @@
...
+        # Set default values for kwargs
+        verbose = 0
 
-    wrapper.is_tp_compiled = True
+        return _compile(fn)
 
-    return wrapper
+    # Else the function is being called as a function, so we return a decorator
+    # that will receive the function to compile.
+    verbose = verbose or 0
```
achoum, line 122: I would make the  value of verbose be 0 in the function header is possible.

achoum (approved):
Awesome

temporian/core/compilation.py
ianspektor: this logic was leftover from a previous worse implementation, good catch :)

PR #198, opened 2023-07-14
Rename and unify evaluation i/o types

ianspektor:
`run`'s i/o types weren't showed in docs, and were used in other modules. renamed them to more explanatory names, moved them to node/event set module, added docstrings to them, added them to docs.

also made the internal symbols show in docs not show full path in the nav bar.


achoum (approved):
Nice

PR #201, opened 2023-07-17
Eventsetornode hints

ianspektor:
All of these are now typed correctly:
image
image
image

And no type errors are thrown in operators nor in @compile
image
image
image

Only had to add a single type ignore (and its because of the hard-to-track logic we're doing to accept arguments in the decorator, but it might be possible to type compile correctly to avoid this, though I don't think its worth the trouble)
image


achoum:
The design is great.

achoum (approved):
Awesome

PR #203, opened 2023-07-18
Add to_tensorflow

achoum:
With tp.to_tensorflow, the use can convert an event-set to a tensorflow dataset.


pyproject.toml
```diff
@@ -63,6 +63,7 @@ apache-beam = "^2.48.0"
 pillow = "^9.5.0"
 cairosvg = "^2.7.0"
 colorama = "^0.4.6"
+tensorflow = "^2.12.0"
```
ianspektor, line 66: this should definitely be an optional dep 👀  check how the beam dependency is defined in the `[tool.poetry.extras]` section (and we should probably add a check to see if tf is installed before running any code that uses it)

ianspektor (approved):

pyproject.toml
achoum: *Done x 2

PR #206, opened 2023-07-19
[CODE WIDE] Operator chaining

ianspektor:
TODO

- [x] try inherited_members or some similar hack
- [x] update notebooks after https://github.com/google/temporian/pull/202 is merged
- [x] migrate window ops
- [x] migrate calendar ops

✅  good:
- loving the syntax already
- autocomplete shows output types correctly
image
image

⚠️ not good:
- docstrings need to be defined in the EventSetOperationsMixin class.
    - `functools.wraps` doesn't work because it makes autocomplete think the function is expecting the evset as first argument, but it isn't, since it's a class method that receives self. also doesn't show up in docs.
    - assigning the function to the class (either during definition, in the constructor, or dynamically after creating it) doesn't work because it gets marked as a variable. also it doesn't show up in docs.


docs/mkdocs.yml
```diff
@@ -96,6 +96,7 @@ plugins:
             group_by_category: true
             show_category_heading: false
             show_root_heading: true
+            inherited_members: true
```
achoum, line 99: Did it worked?

ianspektor: It did :) the option literally got released last week in https://github.com/mkdocstrings/python/releases/tag/1.2.0

ianspektor:
Opened issue to mkdocstrings-python to fix full EventSet source path being shown in docs https://github.com/mkdocstrings/python/issues/90

ianspektor:
Opened another issue for weird behavior when linking to `temporian.EventSet.xxx` methods  https://github.com/mkdocstrings/python/issues/91

docs/mkdocs.yml
achoum: that's what it's called being lucky :)

temporian/core/operators/binary/arithmetic.py
```diff
@@ -141,7 +141,7 @@ def add(
         ValueError: ... corresponding features should have the same dtype. ...
 
         >>> # Cast f1 to float
-        >>> c = tp.cast(a["f1"], float) + a["f2"]
+        >>> c = a["f1"].cast(float) + a["f2"]
```
achoum, line 144: Nit: Not related to this PR; but I just noticed it.

s/float/tp.float64

achoum (approved):
This is amazing. Thanks for pulling this off!!

Nit + optional (another pr): Can you run "python tools/build_cleaner.py" to see if deps need to be updated.

DonBraulio (approved):
Go ahead and merge this one and I can update the notebooks in #202 altogether. That one is ready but not approved yet, so let's resolve this one first if it's ready now.

PR #207, opened 2023-07-20
tp.moving_count ignores features

achoum:
Previously, tp.moving_count was looking at feature values and checking for NaN.
This created a disconnect between the concept of counting and the need for feature values.

The new version of tp.moving_count only look for sampling: tp.moving_count is the number of sampling
in a time-windows.

The old behavior can be emulated with tp.filter and tp.isnan.


temporian/core/operators/window/moving_count.py
```diff
@@ -44,76 +47,65 @@ def moving_count(
     window_length: Duration,
     sampling: Optional[EventSetOrNode] = None,
 ) -> EventSetOrNode:
-    """Computes the number of values in a sliding window over an
+    """Computes the number of samplings in a sliding window over an
```
ianspektor, line 50: s/samplings/events or timestamps

temporian/core/operators/window/moving_count.py
```diff
@@ -44,76 +47,65 @@ def moving_count(
...
     [`EventSet`][temporian.EventSet].
 
     For each t in sampling, and for each index and feature independently,
-    returns at time t the number of non-nan values for the feature in the window
-    (t - window_length, t].
+    returns at time t the number of values in the window (t - window_length, t].
```
ianspektor, line 54: s/values/events here too

ianspektor (approved):

achoum:
Thanks

PR #208, opened 2023-07-24
Notebooks CI test

DonBraulio:
- Notebooks are tested as part of CI, only in GH actions, caching downloads + trained models.
 - Notebooks should still be saved with outputs, since we can't run them on readthedocs because it doesn't have cache (downloading + training models takes a long time).
  -- The only exceptions are `user_guide.ipynb` and `getting_started.ipynb`, which run fast in readthedocs.

### DIFF
--- a/.github/workflows/test_notebooks.yaml
+++ b/.github/workflows/test_notebooks.yaml
@@ -1,6 +1,6 @@
 name: test_notebooks
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   run_notebooks:
--- a/.github/workflows/test_notebooks.yaml
+++ b/.github/workflows/test_notebooks.yaml
@@ -47,7 +47,7 @@ jobs:
         run: poetry install --no-interaction
 
       - name: Install notebook dependencies
-        run: pip install -r ./docs/src/tutorials/requirements.txt
+        run: poetry run pip install -r ./docs/src/tutorials/requirements.txt
 
       - name: Execute all notebooks
-        run: poetry run ./tools/run_notebooks.sh `ls docs/src/*.ipynb`" "`ls docs/src/tutorials/*.ipynb`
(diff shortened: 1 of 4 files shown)


.github/workflows/test_notebooks.yaml
```diff
@@ -0,0 +1,53 @@
+name: test_notebooks
+
+on: [pull_request]
+
+jobs:
+  run_notebooks:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9"]
```
ianspektor, line 10: this can be removed - matrix is useful for when you have several configs you want to test for, a matrix with a single item is the same as just using that item directly

.github/workflows/test_notebooks.yaml
```diff
@@ -0,0 +1,53 @@
...
+          cache-dependency-path: |
+            poetry.lock
+            ./docs/src/tutorials/requirements.txt
+
+
+      - name: Install poetry dependencies
+        run: poetry install --no-interaction
+
+      - name: Install notebook dependencies
+        run: pip install -r ./docs/src/tutorials/requirements.txt
```
ianspektor, line 50: just to confirm, we are sure pip is installing deps within the poetry env, right?

ianspektor (approved):

.github/workflows/test_notebooks.yaml
DonBraulio: Good catch! I know it's working because the notebooks run significantly faster, but reviewing the code, I'm not sure why it works 🤔 . I'm adding `poetry config virtualenvs.create false`, which makes this more clear and is recommended for docker images.

.github/workflows/test_notebooks.yaml
DonBraulio: I know but I think it's better to define this in one single place (it's used multiple times). Also, at some point it could be useful to test more versions.

PR #209, opened 2023-07-27
display evset as table in notebook

ianspektor:
TODOs
- [x] set max number of features, max number of timestamps, and max number of indexes, max number of chars in a cell
- [x] allow configuring those ☝🏼 in some way (like `pd.set_option('display.max_columns', ...)`)
  - (Braulio) Moved options to `config.py`, can be changed through env vars or at runtime: `config.max_display_events = 3` (we should define an object and check types in a following PR).
- [x] test

image


temporian/implementation/numpy/data/event_set.py
```diff
@@ -587,3 +585,76 @@ def creator(self) -> Optional[Operator]:
...
+        """HTML representation, mainly for IPython notebooks."""
+        features = self.schema.features[: config.max_display_features]
+        n_features = len(self.schema.features)
+        cut_features = n_features > config.max_display_features
+        indexes = self.schema.indexes
+        convert_datetime = self.schema.is_unix_timestamp
+        all_index_keys = self.get_index_keys(sort=True)
+        repr = ""
+        for index_key in all_index_keys[: config.max_display_indexes]:
+            repr += "<h3>Index: ("
```
achoum, line 599: It would interesting to find a way to escape html characters, syntax check, and more generally, protect from injections.
There seems to be several python library for that. Alternatively, maybe we can use existing python package (e.g. xml.dom.minidom ) ?

temporian/implementation/numpy/data/event_set.py
```diff
@@ -587,3 +585,76 @@ def creator(self) -> Optional[Operator]:
         created EventSets have a `None` creator.
         """
         return self.node()._creator
+
+    def _repr_html_(self) -> str:
```
achoum, line 589: I see this function are growing significantly, especially when we add interactive elements. Could it be moved in a separate file ?

temporian/implementation/numpy/data/event_set.py
```diff
@@ -587,3 +585,76 @@ def creator(self) -> Optional[Operator]:
...
+        """HTML representation, mainly for IPython notebooks."""
+        features = self.schema.features[: config.max_display_features]
```
achoum, line 591: What is there is less than "max_display_features" ? (same for the other indexes)

temporian/implementation/numpy/data/event_set.py
```diff
@@ -587,3 +585,76 @@ def creator(self) -> Optional[Operator]:
...
+                [
+                    f"{idx.name}={self._repr_value(val, idx.dtype)}"
+                    for val, idx in zip(index_key, indexes)
+                ]
+            )
+            repr += ")</h3>"
+            index_data = self.data[index_key]
+            n_events = len(index_data.timestamps)
+            repr += f"{n_events} events × {n_features} features"
+            repr += "<table><tr><th><b>Timestamp</b></th>"
```
achoum, line 610: Lowercase?

achoum (commented):
This is very cool :)

temporian/utils/config.py
```diff
@@ -1,3 +1,22 @@
...
 
-DEBUG_MODE = os.environ.get("TEMPORIAN_DEBUG_MODE", False)
+debug_mode = bool(os.environ.get("TEMPORIAN_DEBUG_MODE", False))
+
+# Limits for repr(evset), print(evset)
+max_printed_indexes = int(os.environ.get("TEMPORIAN_MAX_PRINTED_INDEXES", 5))
+max_printed_features = int(os.environ.get("TEMPORIAN_MAX_PRINTED_FEATURES", 10))
+max_printed_events = int(os.environ.get("TEMPORIAN_MAX_PRINTED_EVENTS", 20))
+
+# Limits for html display of evsets (notebooks)
```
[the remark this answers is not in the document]
DonBraulio: Agree, changing.

temporian/implementation/numpy/data/event_set.py
DonBraulio: That is not a problem in python, it will stop at the list length.

temporian/utils/config.py
```diff
@@ -1,3 +1,22 @@
...
+# Indexes (number of tables)
+max_display_indexes = int(os.environ.get("TEMPORIAN_MAX_DISPLAY_INDEXES", 10))
+# Features (columns) per table
+max_display_features = int(os.environ.get("TEMPORIAN_MAX_DISPLAY_FEATURES", 20))
+# Events (rows) per table
+max_display_events = int(os.environ.get("TEMPORIAN_MAX_DISPLAY_EVENTS", 100))
```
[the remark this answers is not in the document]
DonBraulio: Ok, I can implement that and print a not-recommended warning, since it's a risky move to hang the notebook.
Do you agree?

temporian/implementation/numpy/data/event_set.py
DonBraulio: Alright, I'll take a look into it.

DonBraulio:
@achoum this is ready to re-review. Changes since your last review:
 - Moved display methods to separate file
 - Using xml.dom.minidom instead of raw html, checked that it prevents html injection
 - Setting limit=0 will disable limit (show all index or feats, events).

temporian/utils/config.py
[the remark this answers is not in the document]
achoum: Sure :)

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        f"features: {evset.schema.features}\n"
+        "events:\n"
+        f"{data_repr}\n"
+        f"memory usage: {string.pretty_num_bytes(evset.memory_usage())}\n"
+    )
+
+
+def display_html(evset: EventSet) -> str:
+    """HTML representation, mainly for IPython notebooks."""
+    from xml.dom import minidom
```
achoum, line 56: I would put this at the top of the file (unless there is reason no to).

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+    max_indexes = config.max_display_indexes or None
+    max_events = config.max_display_events or None
+    max_features = config.max_display_features or None
+
+    # Other configs
+    convert_datetime = evset.schema.is_unix_timestamp
+    visible_features = evset.schema.features[:max_features]
+    index_schemas = evset.schema.indexes
+    all_index_keys = evset.get_index_keys(sort=True)
+    total_indexes = len(all_index_keys)
```
achoum, line 90: num_indexes ?

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+    total_features = len(evset.schema.features)
+    hiding_feats = max_features is not None and total_features > max_features
```
achoum, line 92: has_hidden_feats or  not show_all_feats ?

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Create one table and header per index value
+    for index_key in all_index_keys[:max_indexes]:
+        # Index header text (n events x n features)
+        index_text = ", ".join(
```
achoum, line 97: Is the "," useful? (need to check the print to see if this helps)

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+            [
+                f"{idx.name}={repr_value(val, idx.dtype)}"
+                for val, idx in zip(index_key, index_schemas)
+            ]
+        )
+        index_data = evset.data[index_key]
+        index_n_events = len(index_data.timestamps)
```
achoum, line 104: num_timestamps ?

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        title = dom.createElement("h3")
```
achoum, line 105: Note (for later). In case the style (e.g. h3) is already set (which, it likely the case when rending our documentation), I would not use any h1-h3 style. Instead, we could create our own minimalistic css style.

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        title.appendChild(dom.createTextNode(f"Index: ({index_text})"))
+        descript = dom.createTextNode(
+            f"{index_n_events} events × {total_features} features"
+        )
+        top.appendChild(title)
+        top.appendChild(descript)
+
+        # Table with column names
+        table = dom.createElement("table")
+        col_names = ["timestamp"]
```
achoum, line 115: Nit: Merge in a single line

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        col_names += [feature.name for feature in visible_features]
+        if hiding_feats:
+            col_names += ["..."]
```
achoum, line 118: What about using a h-ellipsis ? (https://www.toptal.com/designers/htmlarrows/punctuation/horizontal-ellipsis/)

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        table.appendChild(create_table_row(col_names, header=True))
+
+        # Rows with events
+        for i, timestamp in enumerate(index_data.timestamps[:max_events]):
```
achoum, line 122: s/i/timestamp_idx

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -0,0 +1,199 @@
...
+            row = []
+
+            # Timestamp column
+            timestamp = (
```
achoum, line 126: Redefining timestamp with another type make reading the code harder.
What about "raw_timestamp" or "timestamp_repr" (or something like that)?

temporian/implementation/numpy/data/test/event_set_test.py
```diff
@@ -69,12 +80,76 @@ def test_repr(self):
...
-        self.assertGreaterEqual(memory_usage, 1200 - 500)
+    def test_html_repr_no_limits(self):
+        config.max_display_indexes = 0
+        config.max_display_features = 0
+        config.max_display_events = 0
+
+        self.assertEqual(
+            self.evset._repr_html_(),
+            "<div><h3>Index: (x=1, y=hello)</h3>"
+            + "3 events × 2 features"
```
achoum, line 120: Since the number of features is the same for all the indexes, maybe we don't need to display it each time.

achoum (approved):
Awesome.
Left a few comments.

temporian/implementation/numpy/data/display_utils.py
DonBraulio: yes, I'd leave it because it's like a tuple:  `(idx_1=val_1, idx_2=val_2)`

temporian/implementation/numpy/data/test/event_set_test.py
DonBraulio: I agree. Though changing this required adding text common to the whole eventset.
Took the chance and included number of indexes and memory usage.
Final result (setting small max_feats and indexes on purpose, only for this demo):
Screen Shot 2023-08-07 at 13 13 24

PR #213, opened 2023-08-02
Combine operator.

DonBraulio:
To fix/discuss @achoum:
To unify timestamps in the docstring example, I wanted to use the smallest window length. Since it's not possible to use a `duration=0` (normalize_duration checks `duration>0`), I defined `duration.eps`. But when I tried to use it in the moving_sum duration, it doesn't work. Apparently C++ interprets this as a 0 in some value range, since there are some empty values in the output.
To reproduce the problem, set `window_length=tp.duration.eps` in the docstring example and check
`bazel test  //temporian/test:docstring_test --test_output=streamed`.

### DIFF
--- a/docs/src/reference/index.md
+++ b/docs/src/reference/index.md
@@ -47,7 +47,8 @@ Check the index on the left for a more detailed description of any symbol.
 | [`EventSet.end()`][temporian.EventSet.end]                             | Generates a single timestamp at the end of the input.                                                        |
 | [`EventSet.enumerate()`][temporian.EventSet.enumerate]                 | Creates an ordinal feature enumerating the events according to their timestamp.                              |
 | [`EventSet.filter()`][temporian.EventSet.filter]                       | Filters out events in an [`EventSet`][temporian.EventSet] for which a condition is false.                    |
-| [`tp.glue()`][temporian.glue]                                          | Concatenates [`EventSets`][temporian.EventSet] with the same sampling.                                       |
+| [`tp.glue()`][temporian.glue]                                          | Concatenates features from [`EventSets`][temporian.EventSet] with the same sampling.                         |
+| [`tp.combine()`][temporian.combine]                                    | Combines events from [`EventSets`][temporian.EventSet] with different samplings.                             |
 | [`EventSet.join()`][temporian.EventSet.join]                           | Join [`EventSets`][temporian.EventSet] with different samplings but the same index together.                 |
 | [`EventSet.lag()`][temporian.EventSet.lag]                             | Adds a delay to an [`EventSet`][temporian.EventSet]'s timestamps.                                            |
 | [`EventSet.leak()`][temporian.EventSet.leak]                           | Subtracts a duration from an [`EventSet`][temporian.EventSet]'s timestamps.                                  |
--- a/temporian/core/operators/combine.py
+++ b/temporian/core/operators/combine.py
@@ -116,7 +116,7 @@ def combine(
         ...                  features={"A": [0, 10, 30], "B": [0, -10, -30]}
         ...                 )
         >>> b = tp.event_set(timestamps=[1, 4],
-        ...                  features={"B": [-10, -40], "A": [10, 40]}
+        ...                  features={"A": [10, 40], "B": [-10, -40]}
         ...                 )
         >>> c = tp.combine(a, b)
         >>> c
@@ -131,7 +131,8 @@ def combine(
 
         >>> # Unify events with the same timestamp (add their values).
         >>> unique_t = c.unique_timestamps()
-        >>> d = c.moving_sum(window_length=0, sampling=unique_t)
+        >>> d = c.moving_sum(window_length=1, sampling=unique_t)
+        >>> d
         indexes: []
         features: [('A', int64), ('B', int64)]
         events:
--- a/temporian/core/operators/combine.py
+++ b/temporian/core/operators/combine.py
@@ -54,6 +54,7 @@ class Combine(Operator):
             first_input.schema.check_compatible_features(
                 input.schema, check_order=False
             )
+            first_input.schema.check_compatible_index(input.schema)
 
             # Output is unix if all inputs are
             all_unix_timestamp &= input.schema.is_unix_timestamp
@@ -149,4 +150,6 @@ def combine(
     if len(inputs) == 1:
         return inputs[0]
 
-    return Combine(inputs).outputs["output"]  # type: ignore
+    # NOTE: input name must match op. definition name
+    inputs_dict = {f"input_{idx}": input for idx, input in enumerate(inputs)}
+    return Combine(**inputs_dict).outputs["output"]  # type: ignore
--- a/temporian/core/operators/combine.py
+++ b/temporian/core/operators/combine.py
@@ -17,42 +17,68 @@
 
 from temporian.core import operator_lib
 from temporian.core.compilation import compile
-from temporian.core.data.node import EventSetNode, create_node_new_features_new_sampling
+from temporian.core.data.node import (
+    EventSetNode,
+    create_node_new_features_new_sampling,
+)
 from temporian.core.operators.base import Operator
+from temporian.core.typing import EventSetOrNode
 from temporian.proto import core_pb2 as pb
+from temporian.utils.rtcheck import rtcheck
+
+MAX_NUM_ARGUMENTS = 30
 
 
 class Combine(Operator):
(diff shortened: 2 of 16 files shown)


temporian/core/operators/combine.py
```diff
@@ -0,0 +1,236 @@
...
+    def index_from(self) -> str:
+        return self._index_from
+
+    @classmethod
+    def build_op_definition(cls) -> pb.OperatorDef:
+        return pb.OperatorDef(
+            key="COMBINE",
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="index_from",
```
achoum, line 92: What about calling it "how" (like we do for the "join" operator), and have the same possible values?

temporian/core/operators/combine.py
```diff
@@ -0,0 +1,236 @@
...
+            outputs=[pb.OperatorDef.Output(key="output")],
+        )
+
+
+operator_lib.register_operator(Combine)
+
+
+@rtcheck
+@compile
+def combine(
```
achoum, line 109: Very cool.

What about something like?

Combines events from multiple EventSets together.

Inputs events must have the same features (i.e. same feature name and dtypes) and indexes (i.e. same index names and dtypes).

Combines is different from glue that appends together different features.

temporian/core/operators/combine.py
```diff
@@ -0,0 +1,236 @@
...
+    *inputs: EventSetOrNode,
+    index_from: str = "union",
+) -> EventSetOrNode:
+    """Combines all events from multiple EventSets with the same features.
```
achoum, line 113: Note: The documentation is now in the EventSet's parent class.

temporian/core/operators/combine.py
```diff
@@ -0,0 +1,236 @@
...
+            idx=b'b' (2 events):
+                timestamps: [1.5 3. ]
+                'A': [15 30]
+            idx=b'c' (2 events):
+                timestamps: [4.5 5.5]
+                'A': [45 55]
+        ...
+
+        >>> # Use "first" to use only index values from the first input a
+        >>> c = tp.combine(a, b, index_from="first")
```
achoum, line 202: left ?

temporian/implementation/numpy/operators/combine.py
```diff
@@ -0,0 +1,112 @@
...
+            combined_keys = set(first_input.data.keys())
+            for evset in input_evsets[1:]:
+                combined_keys &= set(evset.data.keys())
+        else:
+            raise ValueError(f"Unknown parameter for combine: {index_from=}")
+
+        # Concatenate timestamps and features for each index, and sort
+        for index_key in combined_keys:
+            all_timestamps: List[np.ndarray] = []
+            all_feats: Dict[str, List[np.ndarray]] = {
```
achoum, line 69: Should we index features by feature index?

temporian/implementation/numpy/operators/combine.py
```diff
@@ -0,0 +1,112 @@
...
+                index_data = evset.get_index_value(index_key, normalize=False)
+                evset_feats = evset.schema.feature_names()
+                all_timestamps.append(index_data.timestamps)
+                # Access feature data by name, since position in IndexData may vary
+                for feat in first_features:
+                    feat_idx = evset_feats.index(feat)
+                    all_feats[feat].append(index_data.features[feat_idx])
+
+            # Concatenate and sort timestamps
+            timestamps: np.ndarray = np.concatenate(all_timestamps)
```
achoum, line 86: Is there a fonction that combines efficiently concatenate and sort?

achoum (commented):
Awesome

temporian/core/operators/combine.py
DonBraulio: I think this should be like `glue`, intended to be called as `tp.combine(evset_1, evset_2, evset_3)` (as opposed to `evset_1.combine(evset_2, evset_3)`.
Don't you think so?

temporian/implementation/numpy/operators/combine.py
DonBraulio: Couldn't find one. Maybe it's a good one to try some C++ 😄 
But I'd leave it for a future PR.

temporian/implementation/numpy/operators/combine.py
DonBraulio: Yep, you're right. I'll use a list instead of dict with first_features' indexes.

temporian/core/operators/combine.py
ianspektor: Like 👍🏼

temporian/core/operators/glue.py
```diff
@@ -195,10 +195,10 @@ def glue(
         ```
 
     Args:
-        *inputs: EventSets to concatenate.
+        *inputs: EventSets to concatenate their features.
```
ianspektor, line 198: EventSets to concatenate the features of.

temporian/core/operators/glue.py
```diff
@@ -112,7 +112,7 @@ def build_op_definition(cls) -> pb.OperatorDef:
 def glue(
     *inputs: EventSetOrNode,
 ) -> EventSetOrNode:
-    """Concatenates [`EventSets`][temporian.EventSet] with the same sampling.
+    """Concatenates features from EventSets with the same sampling.
```
ianspektor, line 115: Why the removal of the EventSet doc link?

temporian/implementation/numpy/operators/combine.py
```diff
@@ -0,0 +1,112 @@
...
+    IndexData,
+    EventSet,
+    IndexType,
+)
+from temporian.core.operators.combine import Combine
+from temporian.core.operators.combine import (
+    FROM_FIRST,
+    FROM_INTERSECT,
+    FROM_UNION,
+)
```
ianspektor, line 32: could we make this a string enum? is there any reason not to?

temporian/implementation/numpy/operators/combine.py
```diff
@@ -0,0 +1,112 @@
...
+        elif index_from == FROM_UNION:
+            combined_keys = set()
+            for evset in input_evsets:
+                combined_keys |= set(evset.data.keys())
+        elif index_from == FROM_INTERSECT:
+            combined_keys = set(first_input.data.keys())
+            for evset in input_evsets[1:]:
+                combined_keys &= set(evset.data.keys())
+        else:
+            raise ValueError(f"Unknown parameter for combine: {index_from=}")
```
ianspektor, line 64: should we check this in the core operator too?

temporian/implementation/numpy/operators/combine.py
ianspektor: `sortednp` lib seems to have this functionality, see https://stackoverflow.com/questions/27916710/numpy-merge-sorted-array-to-an-new-array, but I'd argue against adding the dependency just for this

ianspektor (approved):

temporian/core/operators/combine.py
DonBraulio: Yep.
Seeing it again from some distance I agree it's better to simplify the naming (although it only refers to the index values and doesn't join events in any sense).

docs/src/reference/index.md
```diff
@@ -47,7 +47,8 @@ Check the index on the left for a more detailed description of any symbol.
 | [`EventSet.end()`][temporian.EventSet.end]                             | Generates a single timestamp at the end of the input.                                                        |
 | [`EventSet.enumerate()`][temporian.EventSet.enumerate]                 | Creates an ordinal feature enumerating the events according to their timestamp.                              |
 | [`EventSet.filter()`][temporian.EventSet.filter]                       | Filters out events in an [`EventSet`][temporian.EventSet] for which a condition is false.                    |
-| [`tp.glue()`][temporian.glue]                                          | Concatenates [`EventSets`][temporian.EventSet] with the same sampling.                                       |
+| [`tp.glue()`][temporian.glue]                                          | Concatenates features from [`EventSets`][temporian.EventSet] with the same sampling.                         |
+| [`tp.combine()`][temporian.combine]                                    | Combines events from [`EventSets`][temporian.EventSet] with different samplings.                             |
```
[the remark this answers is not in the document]
DonBraulio: Good point. I'll put them both at the top.

temporian/implementation/numpy/operators/combine.py
DonBraulio: Nice one, I hadn't found it. Although I agree with not adding the dependency only for this.

PR #216, opened 2023-08-10
Fix bug in glue op

ianspektor:
`sorted()` in the glue implementation gave the wrong order when inputs >= 11, since "input_10" as a string is less than "input_2". remove the sorted since I don't think it's needed, else we'll need to change it for a custom sorted that takes this into account.

### DIFF
--- a/temporian/core/operators/glue.py
+++ b/temporian/core/operators/glue.py
@@ -201,7 +201,7 @@ def glue(
     Returns:
         EventSet with concatenated features.
     """
-    if len(inputs) == 1:
+    if len(inputs) == 1 and isinstance(inputs[0], EventSetNode):
         return inputs[0]
 
     # Note: The node should be called "input_{idx}" with idx in [0, MAX_NUM_ARGUMENTS).
--- a/temporian/implementation/numpy/operators/glue.py
+++ b/temporian/implementation/numpy/operators/glue.py
@@ -37,9 +37,7 @@ class GlueNumpyImplementation(OperatorImplementation):
         output_schema = self.output_schema("output")
 
         # convert input evest dict to list of evsets
-        evsets: List[EventSet] = list(
-            list(zip(*sorted(list(inputs.items()))))[1]
-        )
+        evsets: List[EventSet] = list(list(zip(*list(inputs.items())))[1])
         if len(evsets) < 2:
             raise ValueError(
                 f"Glue operator cannot be called on a {len(evsets)} EventSets."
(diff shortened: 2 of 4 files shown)


temporian/core/operators/glue.py
```diff
@@ -201,7 +201,7 @@ def glue(
     Returns:
         EventSet with concatenated features.
     """
-    if len(inputs) == 1:
+    if len(inputs) == 1 and isinstance(inputs[0], EventSetNode):
```
ianspektor, line 204: This fixes an unrelated bug which caused glue(literally anything) to return that same thing, discovered while passing `glue(evset_list)` instead of `glue(*evset_list)`

temporian/implementation/numpy/operators/glue.py
```diff
@@ -37,9 +37,7 @@ def __call__(
         output_schema = self.output_schema("output")
 
         # convert input evest dict to list of evsets
-        evsets: List[EventSet] = list(
-            list(zip(*sorted(list(inputs.items()))))[1]
-        )
+        evsets: List[EventSet] = list(list(zip(*list(inputs.items())))[1])
```
DonBraulio, line 40: This is confusing. If we don't need to sort, can we just do `list(inputs.values())` ?

temporian/implementation/numpy/operators/test/glue_test.py
```diff
@@ -118,6 +119,51 @@ def test_duplicate_feature(self):
...
+            evset_2,
+            evset_3,
+            evset_4,
+            evset_5,
+            evset_6,
+            evset_7,
+            evset_8,
+            evset_9,
+            evset_10,
+        )
```
DonBraulio, line 165: I agree that the order should be preserved from the input features, anything else seems unexpected.
In this test, I'd explicitly check that a non-alphabetical order is also preserved.

temporian/core/operators/glue.py
ianspektor: Updated combine too

temporian/implementation/numpy/operators/glue.py
ianspektor: lol. thanks. updated.

temporian/implementation/numpy/operators/glue.py
```diff
@@ -36,10 +36,7 @@ def __call__(
         assert isinstance(self.operator, GlueOperator)
         output_schema = self.output_schema("output")
 
-        # convert input evest dict to list of evsets
-        evsets: List[EventSet] = list(
-            list(zip(*sorted(list(inputs.items()))))[1]
-        )
+        evsets = list(inputs.values())
```
achoum, line 39: Can you add a comment saying that dictionary order if guaranteed?

PR #217, opened 2023-08-10
Recipes

DonBraulio:
> :warning: **Everyone please copy the new git hooks** to clear notebooks outputs automatically (see changes in CONTRIBUTING.md).

- [x] Create one example recipe (time aggregation)
- [x] Add recipes/ to documentation and to `test_notebooks.yaml`
- [x] Create script to empty outputs from all notebooks in recipes, add to commit hook
- [x] Create train/test split recipe

See/comment the recipes doc: https://docs.google.com/document/d/1fcs_1snhr6tVkrxqsct-LaVfbfcYBlQcc1Jgyk58ZQo/edit?usp=sharing

We should wait to merge #222 before creating the plot customization recipes, so I removed that task from here.


ianspektor:
I'd change "Proposed solution" in the recipes to "Solution"

ianspektor:
s/Drop the the/Drop the (aggregate_index recipe)

ianspektor:
would s/Unify events with duplicated timestamps/Unify events with identical timestamps (aggregated_duplicated recipe)

ianspektor:
In split_fraction recipe:
- I would make clear that this only works for flat (unindexed) eventsets
- `len(signal_evset.get_index_value(()))` is hard to understand. we need a better way to get an eventset's length (though this is poorly defined when the eventset isn't flat)

ianspektor (approved):
Amazing work 👏🏼  I can see me browsing these myself when needed :) Thanks!

docs/src/recipes/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+For example, to convert events with non-uniform sampling into a time-series with
+a fixed sampling interval, combine events with duplicated timestamps,
+or grouping together events from different indexes.
+
+| Recipe                                                                |
+| --------------------------------------------------------------------- |
+| [Aggregate events at a fixed interval](aggregate_interval.ipynb)      |
+| [Aggregate events from different indexes](aggregate_index.ipynb)      |
+| [Unify events with duplicated timestamps](aggregate_duplicated.ipynb) |
```
achoum, line 17: What about?

"Deduplicate timestamps".

docs/src/recipes/aggregate_index.ipynb
```diff
@@ -0,0 +1,151 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "c74f7111-1b6c-4454-9770-3f67eeadaca6",
+   "metadata": {},
+   "source": [
+    "# Aggregate events from different indexes\n",
+    "\n",
+    "This recipe applies when you have events indexed by one or more features, and you want to drop some of the index levels and unify the events with the same timestamps.\n",
```
achoum, line 10: What about?

This recipe aggregates possibly non-uniformly sampled events into fixed-length intervals (e.g., seconds, hours, days, or weeks). In other words, it converts the event features into time series.

docs/src/recipes/aggregate_index.ipynb
```diff
@@ -0,0 +1,151 @@
...
+    "\n",
+    "For example, assume we have multiple stores with multiple products being sold, so we have daily sale events indexed by store ID and product ID. We want to aggregate sales per store ID, adding up all sales from different product IDs."
```
achoum, line 12: In this example, we aggregate individual sale transactions (such as the timestamp and value of each product purchase) into daily revenue (i.e., the total sales for each day) for each individual store.

docs/src/recipes/split_fraction.ipynb
```diff
@@ -0,0 +1,189 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6d1afd17-ccfa-4bd8-b0bf-406fa098aad6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "train_evset = signal_evset.filter(sample_positions <= train_until)\n",
```
achoum, line 116: Very cool.

While less useful, another cool recipe would be to randomly select events with a random value e.g. .filter(.random_value() < 0.2)

achoum (approved):
Awesome

docs/src/recipes/index.md
DonBraulio: Changed to "Unify events with identical timestamps", I think its clearer; although "deduplicate" is a funnier word :)

docs/src/recipes/split_fraction.ipynb
DonBraulio: Nice one!
I'll add it to the doc since it would require a new op to add the random column, right?

DonBraulio:
> In split_fraction recipe
@ianspektor  I agree with your comments about changing this, I'll get back to it after we change the `begin()` and `end()` ops.

Thanks everyone for the detailed reviews 🙏

PR #220, opened 2023-08-11
Bugfix in event_set: don't allow set() in feature_values

DonBraulio:
Due to an error, I noticed that this code worked but produced a malformed EventSet:

```python
event_set(
    timestamps=[1, 2],
    features={
        # np.array({1, 2}) would produce a single-item value
        "y": {1, 2},
    },
)
```

Similar issues might happen with other iterable types. So I made explicit in `event_set()` the allowed types for features (tuple, list, np.array or pd.Series)

### DIFF
--- a/temporian/implementation/numpy/data/io.py
+++ b/temporian/implementation/numpy/data/io.py
@@ -111,16 +111,16 @@ def event_set(
     if features is None:
         features = {}
 
+    features = {
+        name: normalize_features(value, name)
+        for name, value in features.items()
+    }
+
     # Check timestamps and all features are of same length
     if not all(len(f) == len(timestamps) for f in features.values()):
         raise ValueError(
             "Timestamps and all features must have the same length."
         )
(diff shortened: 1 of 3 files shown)


temporian/implementation/numpy/data/io.py
```diff
@@ -111,17 +111,17 @@ def event_set(
...
+    features = {
+        name: normalize_features(value, name)
+        for name, value in features.items()
+    }
+
     # Check timestamps and all features are of same length
     if not all(len(f) == len(timestamps) for f in features.values()):
         raise ValueError(
             "Timestamps and all features must have the same length."
         )
```
DonBraulio, line 123: @achoum would you agree if I move this code into `IndexData` ?

I mean: `normalize_timestamps`, `normalize_features` and check `len(timestamps)==len(features)`

ianspektor (approved):

PR #221, opened 2023-08-14
fix missing entries in user guide's table of contents

ianspektor:
Using python comments inside a markdown cell messes with the table of contents, since it (buggily) interprets it as an H1 heading (which makes all the headers after this one hidden, since mkdocs expects a single H1 at the top of the file). 

Solution is to either not use those comments (and pass that text to above/below the code cell if possible, or to use same-line comments, i.e. instead of this:
```
# This is a comment that messes up the ToC
<code>
```

do this
```
<code>  # This is a comment that doesn't mess up the ToC
```

note that this only applies to ``` code blocks inside markdown cells, comments inside code cells are fine.

### DIFF
--- a/docs/src/user_guide.ipynb
+++ b/docs/src/user_guide.ipynb
@@ -1643,7 +1643,6 @@
     "lines_to_next_cell": 0
    },
    "source": [
-    "\n",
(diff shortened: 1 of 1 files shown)


DonBraulio (approved):
Ooh nice catch! I would've been ages to figure out this one 😅

PR #225, opened 2023-08-18
M5 tutorial uses new merged plotting (instead of using matplotlib manually)

achoum:


achoum:
> [Braulio](https://github.com/DonBraulio)
The M5 notebook does not contain the cell output. This is the way we want it right?

ianspektor:
> The M5 notebook does not contain the cell output. This is the way we want it right?

If I understand correctly it's only the user guide, getting started, and recipes that we want without outputs

achoum:
Thanks

DonBraulio:
Sorry for the delayed response.
It's as Ian says: any notebook that trains a model or downloads data should be saved **with** outputs, since it would take a long time to execute them in read the docs.

If you install the git hook, you can just save everything with outputs and `git commit` will take care of clearing what needs to be cleared.

PR #226, opened 2023-08-21
Fix notebooks running unreleased code with last-released version

ianspektor:
image

added a `stable` tag that we'll keep up to date when we create a new release (and added to release instructions that were on CONTRIBUTING gude before but were wiped at some point). notebooks can point to that tag to ensure they're running the version of the notebooks that runs with the last-released version of the lib (i.e. the one they install from pypi)


achoum (approved):
Nice

PR #235, opened 2023-08-22
Since last steps

DonBraulio:


temporian/core/operators/since_last.py
```diff
@@ -82,9 +97,12 @@ def has_sampling(self) -> bool:
 def since_last(
     input: EventSetOrNode,
     sampling: Optional[EventSetOrNode] = None,
+    steps: int = 1,
```
achoum, line 100: The default value (1) is defined in several places. What about only defining it in the Eventset::since_last ?

temporian/implementation/numpy_cc/operators/since_last.cc
```diff
@@ -8,46 +8,56 @@
...
-    while (next_event_idx < n_event && v_event[next_event_idx] <= t) {
-      next_event_idx++;
-    }
-    double value;
-    if (next_event_idx == 0) {
-      value = std::numeric_limits<double>::quiet_NaN();
-    } else {
-      value = t - v_event[next_event_idx - 1];
+namespace
+{
```
achoum, line 12: The formatting has changed. It looks like a Microsoft c++ formatting.
Can you change it go Google c++ formatting (setting in VSCode)?

I'll look at the code after.

achoum (commented):
Initial

temporian/implementation/numpy_cc/operators/since_last.cc
DonBraulio: Fixed 👍

temporian/core/operators/since_last.py
DonBraulio: Done.

temporian/core/operators/since_last.py
```diff
@@ -32,10 +32,14 @@ class SinceLast(Operator):
     def __init__(
         self,
         input: EventSetNode,
+        steps: int,
         sampling: Optional[EventSetNode] = None,
     ):
         super().__init__()
 
+        if steps <= 0:
+            raise ValueError("Invalid value received: {steps=}")
```
ianspektor, line 41: would be nice to specify in the error that steps need to be >=1

ianspektor (approved):

temporian/core/operators/since_last.py
DonBraulio: Totally 😄

PR #236, opened 2023-08-23
Update html display of event-set.

achoum:


ianspektor:
- love the new summary of features, indexes, events, etc. at the start
- like the reduced default number of features, events, etc.

on the other hand, I think picking colors can be problematic:
- we can't control what the environment this is shown in looks like (e.g. contrast might not be right, see example below)
- I feel this is a very opinionated look, that not everyone might like
- using same color for two different things (red for both feature names and index names, and blue for both feature types and index values) makes it seem like those are connected, when they aren't

here's what it looks like for me on jupyter lab as an example. white text at the beginning is not readable, and the yellow box has too much brightness for my dark-themed environment.
image

I would propose:
- definitely keeping the new section at the top, but without a background, a separating line or surrounding border could do the trick
- not using any colors
- use bold, italics and headings as much as needed
- remove the border line of each table (this one is only a personal preference, feel free to ignore)

ianspektor (approved):
Looks great! Thanks :)

One more thing, could we remove the binary `b''` from the index keys?

image

achoum:
b'' removed. Note: When a user pass a str in the indexing functions, we need to make sure to convert this string into a bytes (otherwise, the value will not be found).

PR #238, opened 2023-08-23
Test notebooks in parallel

ianspektor:
parallelized notebook runs, time went down from ~35min to ~15min (which is what the M5 one takes to run, we won't be able to push it down further)

downside is we need to add new notebooks to the action's matrix, but I think it's worth the effort, 35min for a ci check generates a lot of friction


DonBraulio (approved):
🙌 
Let's keep an eye on the [caché size](https://docs.github.com/es/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).
If we end up removing-recreating cachés all the time, we should expand the quota (?) or prioritize. These notebook cachés make a huge difference, maybe disabling the bazel cachés is not as impactful.

DonBraulio:
Looking back at the exec times, it looks like there's a significant overhead for running in parallel, due to VMs deployment.
For example, the recipes notebooks only take a couple seconds to run, but the individual action now takes ~6m.

Is it possible to group together the user_guide, getting_started and recipes ? It would also reduce total cache sizes.
Also, consider removing the global `requirements.txt` file, from GH actions and also from RTD (we're no longer running tutorial notebooks there, so we shouldn't need extra deps). This file was intended to decouple the step and measure the installation time, but now it doesn't seem to make sense.

ianspektor:
> the recipes notebooks only take a couple seconds to run, but the individual action now takes ~6m.

Correct, but it isn't an overhead we should care about since its still half what the longest notebook takes. if at any point this is the longest-running job in the CI, I would think about tackling it.

> Is it possible to group together the user_guide, getting_started and recipes ? It would also reduce total cache sizes.

Cache sizes wouldn't be affected, it would be a single cache instead of 3, but with the same size as the other ones combined (also, note that none of those write anything to tmp). Also note that none of those write anything to tmp I think, so they don't actually write anything to cache. 

If you're talking about poetry and bazel caches, those are shared between the jobs, so it isn't a problem either.

It would be possible to group them, but we don't lose anything running them in separate VMs so we'd just be adding extra  complexity with nothing to show for.

> Also, consider removing the global requirements.txt file, from GH actions

I'd like to remove it, but right now its being used for the cache-dependency-path so that the poetry cache's hash is dependent on that file. If we removed it, the cache wouldn't be overwritten when new dependencies are installed for the notebooks, right? I'd like finding a solution to it though

ianspektor:
Merging as-is, opening ticket to discuss you comments next week @DonBraulio

PR #242, opened 2023-08-24
Unsupervised anomaly detection

ianspektor:
managed to get decent results :) and for features to improve results incrementally

TODO
- [x] create config value for max number of (feature_name, feature_dtype) and (index_name, index_dtype) displayed, this notebook displays about 1000 features atm


achoum:
Awesome. Some comments:

1.
What about the following text for the introduction?


This tutorial demonstrates how to use Temporian and Scikit Learn to detect anomalies in a multivariate time series dataset. Anomaly detection in time-series, time-sequence and other temporal formats is critical in a variety of domains. For instance, it is used in manufacturing to detect equipment failure and production lines, in computer systems and by financial institutions to detect fraudulent activities, and in energy management to detect outages.

We will use the [Server Machine Dataset](https://github.com/NetManAIOps/OmniAnomaly/) (SMD) dataset. This dataset contains daily aggregated resource usage metrics (e.g., CPU and RAM utilization, network traffic) from 28 computers in 3 data centers over a period of 5 weeks. The timestamps and feature names, available as a csv file, have been anonymized and normalized. Therefore, calendar feature engineering or the use of expert knowledge is not possible.

The dataset will be loaded, feature engineered and converted into a tabular dataset using Temporian. Then, we will use an [Isolation forest](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.IsolationForest.html#sklearn.ensemble.IsolationForest.predict) model on this tabular data to detect abnormalities. Finally, we will evaluate the quality of our detection on a ground truth of abnormalities.

2.
Loading dataset.

Can we load the raw csv files with Temporian, and with the "combine" operator?

3.
Printing of "evset.schema.features"

I wonder if we can assume that the reader already knows some bit about Temporian. What about adding at the top: We assume you are already familiar ... [link to base tutorial]; and then reduce the amount of beginner details.

This is espetially true since a lot of details are already available in the supervised abnormality detection tutorial.

4.
What about having a link to the other tutorial (supervised abnormality detection) with a quick note on the difference between the two? We could also have a link from the supervised one to this one.

5.
No need to compute "y_train = y.filter(train_mask)". It is confusing (it looks like a supzevized training) :)

6.
What about creating a function "split_event_set" that only split the dataset in two part; and do the separation of the features/labels after. For simplicity, I would remove the one-hot encoding.

7.
In make_datasets, the code structure could be simplified and unified. For example.
- Call the input eventset  "data", "raw_data", "full_data" (or something like that).
- The same variable should not be re-assigned with a different type. For example "X_train" is both an event set and then a numpy array.
- I wonder if we need to keep separate the features and labels in the preparation stage. This is kind of doubling the amount of code.
- I would try to limit the amount of transition from and to pandas if possible.

8.
I would rely on the hyper-parameters of IsolationForest as much as possible (e.g. max_samples). If parameters need to be set, I would explain why.

9.
I think computing the accuracy is a bad idea. I would not compute and report it.
I am also not a fan of the weighting story. I don't see a lot of sense in it.
Instead, for now, report the ROC and ROC AUC. Later (or if you are courageous), report to the AMOC and AMOC AUC.

10.
After a plot or report (e.g. after `evaluate(model, X_train, y_train, X_test, y_test, "lagged features")`), explain what the user needs to notice in the plot.

11.
Coming back on point 3, what about two stages: raw data and many features. We don't need to explain in detail and plot all those features.

12.
The last plot contains two "pred" and two "label" plots. It is not clear what they are. Why not having just one plot.

PR #243, opened 2023-08-25
Improve Beam backend [part 1]

achoum:
- Add Verbose in Beam's run
- Rename read_csv and write_csv to from_csv and to_csv in Beam API
- Replace complex event format enum with bool arg "grouped_by_index"
- Create "help" namespace e.g. tp.help.grouped_by_index()


temporian/beam/io.py
```diff
@@ -346,15 +295,22 @@ def to_event_set(
         schema: Schema of the data. Note: The schema of a Temporian node is
             available with `node.schema`.
         timestamp_key: Key containing the timestamps.
-        format: Format of the input data to be converted into an event-set.
+        grouped_by_index: Are events groupped by index. Run
```
ianspektor, line 298: nit, improve grammar, correct typo: Whether events are grouped by index.

temporian/beam/io.py
```diff
@@ -441,49 +388,48 @@ def to_dict(
...
-    This method is compatible with the output of `read_csv_raw` and the
+    This method is compatible with the output of `from_csv_raw` and the
     Official Beam IO connectors. This method is the inverse of `to_event_set`.
 
     Args:
         pipe: PCollection of Beam EventSet.
         schema: Schema of the data.
         timestamp_key: Key containing the timestamps.
-        format: Format of the output data.
+        grouped_by_index: Are events groupped by index. Run
```
ianspektor, line 402: (same here)

temporian/io/format.py
```diff
@@ -0,0 +1,49 @@
+"""Format of EventSet when importing/exporting them."""
```
ianspektor, line 1: Is this "help" solution standard practice in other libs? I feel like it's a bit convoluted, both for the user (who needs to run a command to get the info they want - I'd rather link them to a page in our docs for example) and for us, who need to maintain this extra code and remember where it lives (to improve this we could name all help files "help.py" stored in corresponding directories). Might be worth it to discuss this in tomorrow's meeting!

ianspektor (approved):

temporian/beam/io.py
achoum: *Done

temporian/beam/io.py
achoum: *Done

PR #244, opened 2023-08-25
Select indexes operator

ianspektor:
Only implemented the `keys` parameter option since I had to add a bunch of stuff to the proto and I want that reviewed before adding the other ones (in another PR).

Operator name TBD, chose a very verbose one for now.

- [X] remove `indexes` parameter from `plot()` and replace usages of it with the new operator
    - will do in a follow up PR


ianspektor:
@achoum doubt that came up while implementing this, why do we allow `str` as a possible value inside an index key? aren't those always normalized to bytes?

ianspektor:
Have some changes in mind pending, will reopen when done to avoid asking for a re-review

achoum:
> why do we allow str as a possible

Because many python users don't understand the difference between bytes and str, and str is easier to feed in small codes / examples.

temporian/core/event_set_ops.py
```diff
@@ -1925,6 +1925,52 @@ def select(
 
         return select(self, feature_names=feature_names)
 
+    def select_index_values(
+        self: EventSetOrNode,
+        keys: Optional[Union[IndexKey, List[IndexKey]]] = None,
+    ) -> EventSetOrNode:
+        """Selects a subset of index values from an
+        [`EventSet`][temporian.EventSet].
```
achoum, line 1933: I would add an explanation for advanced users that when using compiled/graph mode, the selected index values are compiled.

temporian/core/event_set_ops.py
```diff
@@ -1925,6 +1925,52 @@ def select(
 
         return select(self, feature_names=feature_names)
 
+    def select_index_values(
+        self: EventSetOrNode,
+        keys: Optional[Union[IndexKey, List[IndexKey]]] = None,
```
achoum, line 1930: What happens when key is None?
(+Do we need it for now?)

achoum, line 1930: Do we want to support passing indexes that only contain one item without a tuple?
e.g. select_index_values(["A", "B"])

temporian/core/operators/select_index_values.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if isinstance(keys, list) and all(isinstance(k, tuple) for k in keys):
+        pass
+    elif isinstance(keys, tuple):
+        keys = [keys]
+    else:
+        raise TypeError(
+            "Unexpected type for keys. Expect a tuple or list of"
+            f" tuples. Got '{keys}' instead."
+        )
```
achoum, line 92: What about using the same code as "normalize_indexes" in "temporian/implementation/numpy/data/plotter.py"?

(probably moving the code from plotter to typing)

temporian/core/typing.py
```diff
@@ -64,5 +64,18 @@
 is equivalent to `{event_set.node() : event_set}`.
 """
 
-
 TypeOrDType = Union[DType, Type[float], Type[int], Type[str], Type[bool]]
+
+
+IndexKeyItem = Union[int, str, bytes]
```
achoum, line 70: > str
I think this is the difference between user inputs  and internal representation.
We don't allow index on strings. If a user feeds a string (which is allowed), it is converted into a bytes.

temporian/implementation/numpy/data/plotter.py
```diff
@@ -46,10 +45,10 @@
 # "indexes" argument in tp.plot.
 InputIndex = Optional[
```
achoum, line 46: In typing, what about having  both "InputIndex" (possible under a different name) for a user input and a normalized IndexKey (which is a tuple of int and bytes)?

temporian/core/typing.py
```diff
@@ -64,5 +64,18 @@
...
+"""One of the values inside an [IndexKey][temporian.core.typing.IndexKey]."""
+
+IndexKey = Tuple[IndexKeyItem, ...]
```
achoum, line 73: Great name.

temporian/implementation/numpy/operators/select_index_values.py
```diff
@@ -0,0 +1,70 @@
...
+        output_evset = EventSet(data={}, schema=output_schema)
+
+        # TODO: remove this behavior when adding number/fraction params
+        # and force one of the params to not be None
+        if keys is None:
+            keys = []
+
+        # Fill output EventSet's data
+        for key in keys:
+            key = normalize_index_key(key)
```
achoum, line 52: To facilitate the serialization code and catch error early, I would do this conversion when creating the op (e.g., when calling a.set_index_values or when creating a SelectIndexValues object).

temporian/implementation/numpy/operators/select_index_values.py
```diff
@@ -0,0 +1,70 @@
...
+            index_data = input.data[key]
```
achoum, line 53: We probably need a nice error message :)

temporian/core/serialization.py
```diff
@@ -623,6 +622,51 @@ def _unserialize_dtype(dtype: pb.DType):
...
+        )
+    if isinstance(item, bytes):
+        return pb.Operator.Attribute.ListIndexKeys.IndexKey.IndexKeyItem(
+            bytes_=item
+        )
+    if isinstance(item, int):
+        return pb.Operator.Attribute.ListIndexKeys.IndexKey.IndexKeyItem(
+            integer_64=item
+        )
+    else:
```
achoum, line 640: Remove "else"

temporian/proto/core.proto
```diff
@@ -86,6 +87,21 @@ message Operator {
...
     }
+    message ListIndexKeys{
+      repeated IndexKey values = 1;
+
+      message IndexKey {
+        repeated IndexKeyItem values = 1;
+
+        message IndexKeyItem {
+          oneof types {
+            string str = 1;
```
achoum, line 98: If we go with serializing the normalized form only, we don't need the "str" case.

achoum (commented):
Awesome :).
A few comments.

temporian/core/event_set_ops.py
ianspektor: Had it optional since it will be optional when I add the other possible params this week, but agreed that right now its inconsistent, removed

ianspektor: changed `IndexKey = Union[Tuple[IndexValue, ...], IndexValue]` to support receiving single value

temporian/implementation/numpy/data/plotter.py
ianspektor: love it! done

achoum (approved):
I meant to approve it.

PR #245, opened 2023-08-25
Improve Beam backend [part 2]

achoum:
Beam's IO for tf.record files

### DIFF
--- a/temporian/beam/__init__.py
+++ b/temporian/beam/__init__.py
@@ -30,19 +30,12 @@ Solutions:
 """
     )
 
-from temporian.beam.io import dict as _dict
-from temporian.beam.io import csv as _csv
-from temporian.beam.io import tensorflow as _tensorflow
-
-from_csv_raw = _csv.from_csv_raw
-from_csv = _csv.from_csv
-to_csv = _csv.to_csv
-
-to_event_set = _dict.to_event_set
-to_dict = _dict.to_dict
-
-to_tensorflow_record = _tensorflow.to_tensorflow_record
-from_tensorflow_record = _tensorflow.from_tensorflow_record
+from temporian.beam.io.dict import to_event_set, to_dict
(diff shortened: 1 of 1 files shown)


temporian/beam/__init__.py
```diff
@@ -30,14 +30,19 @@
...
+
+from_csv_raw = _csv.from_csv_raw
+from_csv = _csv.from_csv
+to_csv = _csv.to_csv
+
+to_event_set = _dict.to_event_set
+to_dict = _dict.to_dict
+
+to_tensorflow_record = _tensorflow.to_tensorflow_record
+from_tensorflow_record = _tensorflow.from_tensorflow_record
```
ianspektor, line 45: this would be simpler by importing the names directly from each module, since they're not being renamed (`from temporian.beam.io import from_csv_raw, from_csv, to_csv`, etc)

temporian/beam/io/tensorflow.py
```diff
@@ -0,0 +1,273 @@
...
+            `MatchFiles`.
+        schema: Schema of the data. If you have a Temporian node, the schema is
+            available with `node.schema`.
+        timestamp_key: Key containing the timestamps.
+        grouped_by_index: Are events groupped by index. Run
+            `tp.help.grouped_by_index()` for the documentation. Currently, only
+            grouped_by_index=True is implemented.
+
+    Returns:
+        A PCollection of event-set compatible with tpb.run.
```
ianspektor, line 181: s/event-set/EventSet 🙏🏼

ianspektor (approved):

temporian/beam/io/tensorflow.py
achoum: But there is an issue here.
It is not a PCollection of EventSet object (instead it is a collection of IndexValue), which together represent an EventSet. Same as for the index, we need to normalize the name.
*Adding to tomorrow brainstorming session agenda

PR #246, opened 2023-08-29
Beam Tutorial

achoum:
Note: Only the first part of the tutorial is effectively executing some Temporian + Beam code.
I am hopping to have a "Temporian on Beam using Dataflow" tutorial in a near future.


docs/src/tutorials/temporian_with_beam.ipynb
ianspektor, line 1: the installation needs to be `%pip install temporian[beam] -q` right?

ianspektor (approved):

docs/src/tutorials/temporian_with_beam.ipynb
achoum: Yes. Good point.
*Done

PR #249, opened 2023-08-30
Propose new display w/o colors

ianspektor:
This is a proposal/example to discuss over. I'm not sure I prefer it over the colored one:
- colored one is easier to look at/you can find the info you're searching for faster if you are already used to the colors
- non-colored one is more sober, less opinionated, will be less intrusive when displayed in places where design matters (think screenshots of an evset in a blogpost, slide deck, etc.)

Colored (current one)

image

Non-colored (proposed)

image


achoum:
Thanks for the proposition.

If the main issue with color is about style (e.g, sobriety), I think making the library efficient to use (e.g., ease of use) should be considered more important. If someone were to wanting to make a screenshot without colors for a blog post, applying a desaturation or an extension (e.g. Dark Reader).

A possible issue I could see with color is about confusing users e.g. "Why is that red?". I think we could try to address this issue with a better selection of color:

The currently proposed colors are placeholder colors that were randomly selected using a white background. Maybe it would make sense to select proper colors. For instance, what about trying dimmed VSCode palette?

For my work, I would prefer colors as any gain of speed is valuable. If a significant number of people will be negatively impacted by having color (after having tried better colors), I propose we add an option to disable it (e.g. a little javascript button to disable color).

The change of syntax in feature schema and index values looks good.

ianspektor:
In-between proposal:
- keep new syntax for index keys and dtypes (more condensed)
- swapped color palette for a color-blind one, can be refined further
- will add @DonBraulio's `disable_colors` proposal to config 👍🏼 

image

achoum:
Nice.

- What do you think about a desaturated image?
- Nit: In  my opinion, there is a bit too much bold.

desaturated

PR #250, opened 2023-08-30
Change internal representation of event-set in the beam backend.

achoum:
Prior to this change, different features were stored in different items of the same pcollection.
After this change, different features are stored in different items in different pcollection.
This makes the implementation of operator easier (with the exception of IO operations) and potentially more efficient (e.g., tp.select is now free).

Also:
- Update window operator (both beam and in-process) so than a missing feature index is treated as a missing feature value.
- Hint name convention is similar to the in-process hint name convention recently introduced.
- Implement moving_sum with sampling in Beam.


temporian/beam/test/utils.py
```diff
@@ -49,15 +48,20 @@ def check_beam_implementation(
             instead.
     """
 
-    if input_node is None:
-        input_node = input_data.node()
+    if isinstance(input_data, EventSet):
+        input_data = [input_data]
+
+    # input_node = input_data.node()
```
ianspektor, line 54: remove

temporian/beam/typing.py
ianspektor, line 1: Love this.

If any of these are part of the public API (e.g. BeamEventSet) we should add them to the docs + add a docstring to them (note that docstrings for a typevar are defined under it with the """ syntax).

temporian/beam/typing.py
```diff
@@ -0,0 +1,75 @@
...
+# The Beam computation relies eavily on tuples instead of dataclasses or named
+# tuples as it seems tuples are the most efficient solution.
+#
+# Individual features of a same event set are stored separately.
+# An important implication is that timestamps are effectively repeated for
+# each feature.
+
+# Temporian index in Beam.
+BeamIndexKeyItem = Union[int, bytes]
+BeamIndexKey = Tuple[BeamIndexKeyItem, ...]
```
ianspektor, line 32: These two are the same as `NormalizedIndexKeyItem` and `NormalizedIndexKey`, does it make sense to have them separate?

temporian/beam/typing.py
```diff
@@ -0,0 +1,75 @@
...
+# StructuredRow is a single event / row during the conversion from dict of
+# key/value to internal the Temporian beam format for EventSets. In a
+# StructuredRow, index and features are indexed by integer instead of string
+# keys.
+SingleFeatureValue = Any  # np.generic
+StructuredRowValue = Tuple[TimestampsPyType, Tuple[SingleFeatureValue, ...]]
+StructuredRow = Tuple[BeamIndexKey, StructuredRowValue]
+
+# A FeatureItem contains the values of a features + timestamps for
+# an index. This is the internal representation used during the op computation.
```
ianspektor, line 48: nit: move this comment to where FeatureItem is defined

temporian/beam/typing.py
```diff
@@ -0,0 +1,75 @@
...
+# If no feature is available (i.e. event-set without features), the
+# "FeatureValues" is set to None and the FeatureIdx is set to -1.
+FeatureValues = np.ndarray
+FeatureIdx = int
+
+# Index of the timestamps, feature and feature idx in the "FeatureItem" and
+# "FeatureItemWithIdx" tuples.
+PosTimestampValues = 0
+PosFeatureValues = 1
+PosFeatureIdx = 2
```
ianspektor, line 59: all-caps snake case (POS_TIMESTAMP_VALUES, etc.) could make sense for these constants

temporian/beam/typing.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Same as "FeatureItem", but with the idx of the feature. Used during import and
+# export.
+FeatureItemWithIdxValue = Tuple[
+    TimestampValues, Optional[FeatureValues], FeatureIdx
+]
+FeatureItemWithIdx = Tuple[BeamIndexKey, FeatureItemWithIdxValue]
+
+# From the point of view of the user, a BeamEventSet play the same role as
+# and EventSet with Temporian in-process.
```
ianspektor, line 74: s/and/an, s/with/in

temporian/beam/typing.py
```diff
@@ -0,0 +1,75 @@
...
+FeatureValues = np.ndarray
+FeatureIdx = int
+
+# Index of the timestamps, feature and feature idx in the "FeatureItem" and
+# "FeatureItemWithIdx" tuples.
+PosTimestampValues = 0
+PosFeatureValues = 1
+PosFeatureIdx = 2
+
+# A list of timestamps and optionnally feature values.
```
ianspektor, line 61: typo, optionally

ianspektor (approved):
LGTM 👍🏼  Love the centralization of the typing stuff.

(just a heads up, I'm doing very shallow reviews of the beam code so far, will go deeper into it once I start contributing to it)

temporian/beam/test/utils.py
achoum: *Done

temporian/beam/typing.py
achoum: Yes, those are the same.
* Replaced.

temporian/beam/typing.py
achoum: This entire block is about the definition of "FeatureItem". I define sub messages before to make the definition of "FeatureItem" more compact.

temporian/beam/typing.py
achoum: *Fixed

temporian/beam/typing.py
achoum: Good point.
*Done

PR #253, opened 2023-09-04
Refactor grouped_by_index param

ianspektor:
I prefer this over the CLI approach, but don't love it:
1. Option A, we hardcode the full `grouped_by_index` docs link into the docstring (with the absolute URL and all), and it breaks silently when we move it to another place
2. Option B, we use the relative link as I did now, mkdocs fails if it breaks which is great, but users can't click on it to open the docs, they'll probably need to search for it

Since this is a complicated argument used in several places I would argue towards making the format an enum again, called `BeamEventSetFormat`. This would allow us to write docs there, easily display it in docs, and easily link users towards it in the parameter specs. It also future-proofs us in case a third format comes up.

I'm up for making the change if you agree @achoum, lmk!


ianspektor (commented):
discussed in today's meeting:

make 2 enums:
- one for format options from/to tensorflow
- one for dict

create Type = Union[Enum, "value1", "value2", etc] for typing the functions that receive these enums as params so that user can pass string directly instead of tp.beam.format.Format.value

parameter name should be "format"

achoum (approved):
Approved with the solution discussed today.

ianspektor:
DB server of the loan_outcome_prediction tutorial seems to be down, will wait for some hours and try again

PR #254, opened 2023-09-07
Variable window length

ianspektor:
Python
- made a new `WindowlLength` type for the `window_length` param, so that it can be either a `Duration` or `EventSet`.
- if a `Duration` it gets serialized as an attribute, if an `EventSet` it gets serialized  as an input. this is transparent to the user.
- implementation checks whether a `window_length` input `EventSet` was received, else it uses the attribute (one of them will be None)
- could make sense to make these two different attributes (`window_length` and `variable_window_length`), though I like this approach better

CPP
- added 2 new `accumulate()` overloads for variable window length with and without external sampling and extended cpp and python defs for them

### DIFF
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -2205,7 +2205,7 @@ class EventSetOperations:
 
         Raises:
             KeyError: If any of the specified `indexes` are not found in the
-            input.
+                input.
         """
         from temporian.core.operators.add_index import set_index
 
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -1392,7 +1392,10 @@ class EventSetOperations:
         window (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         Example without sampling:
             ```python
@@ -1473,7 +1476,10 @@ class EventSetOperations:
         (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         If the window does not contain any values (e.g., all the values are
         missing, or the window does not contain any sampling), outputs missing
@@ -1524,7 +1530,10 @@ class EventSetOperations:
         (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         If the window does not contain any values (e.g., all the values are
         missing, or the window does not contain any sampling), outputs missing
@@ -1547,8 +1556,8 @@ class EventSetOperations:
 
             ```
 
-        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for examples
-        of moving window operations with external sampling and indices.
+        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for
+        examples of moving window operations with external sampling and indices.
 
         Args:
             window_length: Sliding window's length.
@@ -1575,7 +1584,10 @@ class EventSetOperations:
         (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         Missing values (such as NaNs) are ignored.
 
@@ -1600,8 +1612,8 @@ class EventSetOperations:
 
             ```
 
-        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for examples of moving window
-        operations with external sampling and indices.
+        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for
+        examples of moving window operations with external sampling and indices.
 
         Args:
             window_length: Sliding window's length.
@@ -1632,7 +1644,10 @@ class EventSetOperations:
         time t the sum of the feature in the window (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         Missing values (such as NaNs) are ignored.
 
@@ -1657,8 +1672,8 @@ class EventSetOperations:
 
             ```
 
-        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for examples of moving window
-        operations with external sampling and indices.
+        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for
+        examples of moving window operations with external sampling and indices.
 
         Args:
             window_length: Sliding window's length.
@@ -2182,12 +2197,15 @@ class EventSetOperations:
         (t - window_length, t].
 
         If `sampling` is provided samples the moving window's value at each
-        timestamp in `sampling`, else samples it at each timestamp in the input.
+        timestamp in `sampling`, else samples it at each timestamp in
+        `window_length` if it is an `EventSet`, else in the input. If both
+        `sampling` and `window_length` are provided as `EventSets`, their
+        sampling must be the same.
 
         Missing values (such as NaNs) are ignored.
 
         If the window does not contain any values (e.g., all the values are
-        missing, or the window does not contain any sampling), outputs missing
+        missing, or the window does not contain any timestamp), outputs missing
         values.
 
         Example:
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -2197,14 +2197,15 @@ class EventSetOperations:
 
         Args:
             indexes: List of index / feature names (strings) used as
-                the new indexes. These names should be either indexes or features in
-                the input.
+                the new indexes. These names should be either indexes or
+                features in the input.
 
         Returns:
             EventSet with the updated indexes.
 
         Raises:
-            KeyError: If any of the specified `indexes` are not found in the input.
+            KeyError: If any of the specified `indexes` are not found in the
+            input.
         """
         from temporian.core.operators.add_index import set_index
 
@@ -2252,8 +2253,8 @@ class EventSetOperations:
 
             ```
 
-        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for examples of moving window
-        operations with external sampling and indices.
+        See [`EventSet.moving_count()`][temporian.EventSet.moving_count] for
+        examples of moving window operations with external sampling and indices.
 
         Args:
             window_length: Sliding window's length.
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -14,7 +14,6 @@
 
 # pylint: disable=import-outside-toplevel
 
-
 from __future__ import annotations
 from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
 
@@ -22,7 +21,12 @@ from temporian.core.data.duration import Duration
 
 
 if TYPE_CHECKING:
-    from temporian.core.typing import EventSetOrNode, TypeOrDType, IndexKeyList
+    from temporian.core.typing import (
+        EventSetOrNode,
+        TypeOrDType,
+        IndexKeyList,
+        WindowLength,
+    )
 
 T_SCALAR = (int, float)
 
@@ -1379,7 +1383,7 @@ class EventSetOperations:
 
     def moving_count(
         self: EventSetOrNode,
-        window_length: Duration,
+        window_length: WindowLength,
         sampling: Optional[EventSetOrNode] = None,
     ) -> EventSetOrNode:
         """Gets the number of events in a sliding window.
@@ -1387,9 +1391,8 @@ class EventSetOperations:
         Create a tp.int32 feature containing the number of events in the time
         window (t - window_length, t].
 
-        If the `sampling` argument is not provided, outputs a timestamp for
-        each timestamp in `input`. If the `sampling` argument is provided,
-        outputs a timestamp for each timestamp in `sampling`.
+        If `sampling` is provided samples the moving window's value at each
+        timestamp in `sampling`, else samples it at each timestamp in the input.
 
         Example without sampling:
             ```python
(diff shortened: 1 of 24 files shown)


temporian/implementation/numpy_cc/operators/window.cc
ianspektor, line 1: @achoum the extension reformatted this whole file even though I have Google's format defined, maybe it was misformatted before? If I try to reformat since_last.cc for example it leaves it as-is.

Anyway, I did the reformatting as the first commit in the CPP stuff for easier review, try filtering out the "format window.cc" commit (`1e5bfd0`), or click [here](https://github.com/google/temporian/pull/254/files/1e5bfd00b19b222f88136185a2269ac2637dc70f..7b6aa65f9760cc9399fe15f680feadb9644508a7) (link will be outdated if I push any more commits after this comment)

temporian/implementation/numpy/operators/test/moving_count_test.py
```diff
@@ -54,6 +53,34 @@ def test_cc_wo_sampling(self):
...
         )
 
+    def test_cc_window_0(self):
+        assert_array_equal(
+            operators_cc.moving_count(
+                _f64([1, 2, 3, 5, 20]),
+                0,
+            ),
+            _i32([0, 0, 0, 0, 0]),
+        )
```
ianspektor, line 63: Is this behavior intended? Should a 0-sized window not include the item it ends on? It behaves like this with both fixed and variable window

DonBraulio: I think it's the expected behavior. In fact, this is only possible calling directly to the cpp operator, because null durations are not allowed from the python api.
We have `tp.duration.shortest` for cases where you want to make sure that a window doesn't overlap two different timestamps.

temporian/implementation/numpy/operators/test/simple_moving_average_test.py
```diff
@@ -99,6 +99,27 @@ def test_cc_w_nan_w_sampling(self):
...
+    def test_cc_w_sampling_w_variable_winlength(self):
+        assert_array_equal(
+            cc_sma(
+                _f64([0, 1, 2, 3, 5, 20]),  # timestamps
+                _f64([nan, 10, 11, 12, 13, 14]),  # feature
+                _f64([-1, 1, 4, 19, 20, 20]),  # sampling
+                _f64([10, 0.5, 2.5, 19, 16, np.inf]),  # window length
+            ),
+            _f64([nan, 10, 11.5, 11.5, 13.5, 12]),
+        )
```
DonBraulio, line 121: Can we add a case where we only have 1 or 2 sampling points and many different window lengths for each sampling?
E.g: step at the end timestamp and consider an `inf` window length, `0.1*total_length`, `0.5*total_length`, `0.25*total_length`, etc. Preferably in random order.

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,568 @@
...
+           v_timestamps[first_diff_ts_idx] == curr_ts) {
+      accumulator.Add(v_values[first_diff_ts_idx]);
+      first_diff_ts_idx++;
+    }
+
+    if (end_idx > 0) {
+      // Move begin_idx forward or backwards depending on begin_diff.
+      if (curr_ts - v_timestamps[end_idx - 1] -
+              (curr_window_length - v_window_length[end_idx - 1]) >
+          0) {
```
DonBraulio, line 176: Nice, I think this ensures that it works with maximum float precision.
It would be great if we can add a test using `tp.duration.shortest` (see `test_cc_max_resolution_wo_sampling` in moving_count_test) to ensure that it works here as well.

DonBraulio (commented):
Nice work! LGTM code-wise, left some comments, mainly to test a few edge cases.

temporian/implementation/numpy/operators/test/moving_count_test.py
ianspektor: Gotcha, fixed 👍🏼 Thanks!

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,568 @@
...
-    return output;
+    while (begin_idx < n_event &&
+           // Compare both sides around ~0 to get maximum float resolution
+           v_sampling[sampling_idx] - v_timestamps[begin_idx] >=
+               window_length) {
+      accumulator.Remove(v_values[begin_idx]);
+      begin_idx++;
+    }
+
+    v_output[sampling_idx] = accumulator.Result();
```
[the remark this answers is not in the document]
ianspektor: see comment above about filtering out the "format window.cc" commit

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,566 @@
...
+  // Allocate output array
+  auto output = py::array_t<OUTPUT>(n_sampling);
+
+  auto v_output = output.template mutable_unchecked<1>();
+  auto v_timestamps = evset_timestamps.unchecked<1>();
+  auto v_values = evset_values.template unchecked<1>();
+  auto v_sampling = sampling_timestamps.unchecked<1>();
+  auto v_window_length = window_length.unchecked<1>();
+
+  TAccumulator accumulator;
```
achoum, line 228: What about having a assert than the number of values in" timestamps" == number of values in "values", and number of values in "window_length" == number of values in "sampling_timestamps"?

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,566 @@
...
+
+  // Index of the first value in the window.
+  size_t begin_idx = 0;
+  // Index of the first value outside the window.
+  size_t end_idx = 0;
+
+  double curr_ts;
```
achoum, line 151: Nit: here, there might be only limited benefit to declaring the variable outside of the while.

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,566 @@
...
-      if constexpr (std::numeric_limits<INPUT>::has_quiet_NaN)
-      {
-        if (std::isnan(value))
-        {
-          return;
+    // Set current value of window to all values with the same timestamp.
+    const auto result = accumulator.Result();
+    for (size_t i = end_idx; i < first_diff_ts_idx; i++) {
+      v_output[i] = result;
+    }
```
achoum, line 200: Could this not hold anymore?
For example, if multiple values with the same timestamp do not have the same window length.

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,566 @@
...
+    while (end_idx < n_event && v_timestamps[end_idx] <= right_limit) {
+      accumulator.Add(v_values[end_idx]);
+      end_idx++;
+    }
+
+    if (end_idx > 0) {
+      // Move begin_idx forward or backwards depending on begin_diff.
+      if (right_limit - v_sampling[sampling_idx - 1] -
+              (curr_window_length - v_window_length[sampling_idx - 1]) >
+          0) {
```
achoum, line 246: To improve readability, and factor code, what about creating a function:

bool contract_window(double ts, double last_ts, double window_length, double last_window_length)

returning (ts - last_ts - (window_length - last_window_length)) > 0
?

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,566 @@
...
-            {
-              current_extremum = value;
-            }
+    }
+
+    assert(values.front() == value);
+    if (values.size() == 1) {
+      values.clear();
+    } else {
+      assert(!values.empty());
```
achoum, line 437: Nit: I would add an assert that, before the pop, the front value of "values" is equal to "value".

temporian/core/operators/window/moving_count.py
```diff
@@ -44,7 +43,7 @@ def feature_schema(self, input: EventSetNode):
 @compile
 def moving_count(
     input: EventSetOrNode,
-    window_length: Duration,
+    window_length: WindowLength,
```
achoum, line 46: What happens if the feature window_length is negative? (e.g., error, skip, replace with zero)

temporian/implementation/numpy/operators/test/moving_max_test.py
```diff
@@ -58,14 +59,35 @@ def test_cc_wo_sampling(self):
...
+                _f64([0, 1, 2, 3, 5, 20]),  # timestamps
+                _f32([nan, 10, nan, 12, 13, 14]),  # feature
+                _f64([-1, 3, 40]),  # sampling
                 3.5,
             ),
             _f32([nan, 12, nan]),
         )
 
+    def test_cc_wo_sampling_w_variable_winlength(self):
+        assert_array_equal(
```
achoum, line 71: Some ideas of things to test (here, or in one of the other functions) :)

- Duplicate timestamps with same window length.
- Duplicate timestamps with different window length.
- Duplicate timestamps where some nan window length.
- Empty arrays
- A single value with NaN window length.
- A single value with NaN value and finite window length
- A single NaN value with NaN window length.

temporian/implementation/numpy/operators/window/base.py
```diff
@@ -62,11 +63,24 @@ def __call__(
...
+                if window_length is not None:
+                    window_length_data = window_length.data[index_key].features[
+                        0
+                    ]
+                    # Check all window length values are positive
+                    if not np.all(window_length_data > 0):
+                        raise ValueError(
+                            "All values in `window_length` must be strictly"
+                            " positive."
+                        )
```
achoum, line 76: It could be dangerous for the data to be able to trigger an error.
Instead, I think that it would be better to have a defined non-error behavior in case the window length is negative (possibly with a warning).

achoum (commented):
This is great :).
Left some comments.

temporian/implementation/numpy/operators/test/simple_moving_average_test.py
ianspektor: added

temporian/implementation/numpy_cc/operators/window.cc
ianspektor: added

temporian/core/event_set_ops.py
```diff
@@ -1379,17 +1383,16 @@ def leak(self: EventSetOrNode, duration: Duration) -> EventSetOrNode:
...
         """Gets the number of events in a sliding window.
 
         Create a tp.int32 feature containing the number of events in the time
         window (t - window_length, t].
 
-        If the `sampling` argument is not provided, outputs a timestamp for
-        each timestamp in `input`. If the `sampling` argument is provided,
-        outputs a timestamp for each timestamp in `sampling`.
+        If `sampling` is provided samples the moving window's value at each
+        timestamp in `sampling`, else samples it at each timestamp in the input.
```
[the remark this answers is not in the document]
ianspektor: I would change third option's behavior: If "window_length" is a feature and "sampling" is not provided, "window_length"'s sampling will be used.

temporian/implementation/numpy/operators/test/moving_sum_test.py
```diff
@@ -207,6 +225,62 @@ def test_with_sampling(self):
...
+                    [1, 10.0],
+                    [2, 11.0],
+                    [3, 12.0],
+                    [5, 13.0],
+                    [6, 14.0],
+                ],
+                columns=["timestamp", "a"],
+            )
+        )
+        sampling = from_pandas(
```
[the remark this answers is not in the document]
ianspektor: Agreed, but rest of this file was using the pandas constructor so I sticked to it for consistency :) See next meeting's notes where I added a section about revamping the whole operator test structure/content

temporian/implementation/numpy_cc/operators/window.cc
ianspektor: Mind expanding on that? Although a quick search reveals a possible answer https://stackoverflow.com/questions/7959573/declaring-variables-inside-loops-good-practice-or-bad-practice (TLDR scoping the variable correctly lets compiler do better optimization + the variable isn't re-allocated on each loop, only once). Changed all declarations to `const auto` inside the loop

temporian/implementation/numpy_cc/operators/window.cc
ianspektor: assert is a couple lines above

temporian/implementation/numpy/operators/window/base.py
ianspektor: Agreed, changed to warning and added test

temporian/core/operators/window/moving_count.py
ianspektor: outputs missing values in this case, added docs + tests

temporian/implementation/numpy/operators/test/moving_max_test.py
ianspektor: added the ones not covered in existing ones

temporian/core/typing.py
```diff
@@ -94,6 +95,22 @@
...
+If a [Duration][temporian.duration.Duration], the window length is constant.
+
+If an [EventSet][temporian.EventSet], the window length will vary for each
+timestamp in the effective sampling (i.e., `sampling` if it was provided or the
+input otherwise).
+
+In this case, the `window_length` `EventSet` must have a single positive float64
+feature, and the same sampling as the effective sampling, and the window length
+for each timestamp in the effective sampling will be the feature value at that
+timestamp in `window_length`.
```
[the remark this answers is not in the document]
achoum: Note that in the cc code, "window_length" is the value of the window length without timestamps information.
So the timestamps information should still be provided one way of another.

The concept of "effective sampling" may be unfamiliar to the user. What about?:

A window length can be either constant or variable.

A constant window length is specified with a [Duration][temporian.duration.Duration]. For example, `window_length=5.0` or `window_length=tp.duration.days(4)`.

A variable window length is specified as an [EventSet][temporian.EventSet] object with a single feature of type `tp.float64`. The sampling of this event-set is restricted as follows:

(1) If you call a moving operator without extra `sampling` argument (e.g., `a.moving_sum(window_length=b)`), the window length event-set `b` should have the same sampling as the input `a`.

(2) If you call a moving operator with the sampling argument (e.g., `a.moving_sum(window_length=b, sampling=c)`), the window length event-set `b` should have the same sampling as the sampling `c`.

temporian/core/event_set_ops.py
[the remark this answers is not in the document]
achoum: This is a more powerful and compact version. This is good :)
In this case, would it make sense to remove case 2 (Do not allow both a window_length and a sampling argument) to make sure there is only one way to do things?

As a remark, I think we loose something: So far, if a user does not provide the sampling argument, he/she knows that the output as the same sampling as the input. When a user is faced with a sampling mismatch error, it is easy to look for the ops with a `sampling` argument. However, with this new version of case 3, moving operators can change the output sampling without the `sampling` argument.

temporian/implementation/numpy_cc/operators/window.cc
achoum: Missing?

temporian/implementation/numpy_cc/operators/window.cc
achoum: Nice :).
Note however that it depends on the object (and the cost of construction / destruction). The person asks about string and the first answer is about ints.
For example, with simple primitives (e.g., double), it doesn't matter. All three functions compile to the same code: https://godbolt.org/z/qboWse5v4
However, if you have an expensive object that can be reused, it should be put outside of the loop (see second answer in your link).

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -10,436 +10,571 @@
...
-          current_extremum = values.front();
-          for (const auto value : values)
-          {
-            if (Compare(value, current_extremum))
-            {
-              current_extremum = value;
-            }
+    }
+
+    assert(values.front() == value);
```
achoum, line 437: You cannot ask "front" if the value is empty.

achoum (approved):
Awesome.

temporian/core/event_set_ops.py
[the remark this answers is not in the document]
ianspektor: Agree with prohibiting specifying both sampling and variable window_length, changing it

PR #255, opened 2023-09-07
Operator tick_calendar

DonBraulio:
Implementing pretty much as discussed in pseudo-code.

I'm not using tuples for ranges (e.g: `days=(1,31)`), just separate parameters like `min_day` and `max_day`, because it's easier for attribute types and interfacing with C++ function.

- [x] Operator & C++ implementation
- [x] Basic docstring
- [x] Auto setup and control `None` args.
- [x] Unit tests
- [x] More docstring examples

Important changes since prev. review:
- Using `Literal['*']`, handling that in `typecheck.py`.
- Using attribute type ANY, serializing '*' or number.
- Added tests for auto-parameters and invalid ranges/types.


temporian/implementation/numpy/operators/test/tick_calendar_test.py
```diff
@@ -0,0 +1,177 @@
...
+)
+from temporian.implementation.numpy.operators.test.test_util import (
+    assertEqualEventSet,
+    testOperatorAndImp,
+)
+
+
+class TickCalendarOperatorTest(absltest.TestCase):
+    def setUp(self):
+        pass
```
achoum, line 32: Can be removed.

temporian/core/operators/tick_calendar.py
```diff
@@ -0,0 +1,287 @@
...
+    ):
+        super().__init__()
+        if not input.schema.is_unix_timestamp:
+            raise ValueError(
+                "Can only use tick_calendar on unix timestamp samplings"
+            )
+
+        # Attributes
+        self._min_second = min_second
+        self._max_second = max_second
```
achoum, line 56: This min/max logic is kind of a (great) implementation detail.
However, it means that we cannot later change / improve / extend the implementation (since OP will possibly be serialized).

If the conversion  from user-input to min/max would be done in the constructor of TickCalendar, instead of in tick_calendar, the code would be more extensive.

temporian/implementation/numpy/operators/tick_calendar.py
```diff
@@ -0,0 +1,80 @@
...
+        assert isinstance(self.operator, TickCalendar)
+        output_schema = self.output_schema("output")
+
+        # Create output EventSet
+        output_evset = EventSet(data={}, schema=output_schema)
+
+        # Fill output EventSet's data
+        for index_key, index_data in input.data.items():
+            if len(index_data.timestamps) < 2:
+                dst_timestamps = np.array([], dtype=np.float64)
```
achoum, line 46: Add a note in the documentation telling that, if there is a single event that fall exactly on a calendar event specified by the user, the tick_calendar op will *not* generate a timestamp.

I think it would be simpler if a single event would be begin = end = single event timestamp.

Note: That an event set can contain multiple events with the same timestamp value.

temporian/implementation/numpy/operators/tick_calendar.py
```diff
@@ -0,0 +1,80 @@
...
+            else:
+                begin = int(index_data.timestamps[0])  # discard sub-second info
```
achoum, line 48: s/begin/start_timestamp (since this is the same thing.

temporian/implementation/numpy_cc/operators/tick_calendar.cc
```diff
@@ -0,0 +1,111 @@
...
+#include <string>
+#include <vector>
+
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+py::array_t<double> tick_calendar(
+    const long start_timestamp,                  // min date
```
achoum, line 17: const double

Then, when converting to "std::time_t", apply a floor for begin, and a ceil for end.

Note: "long" is 32 bits (or at least, it can be).

temporian/implementation/numpy_cc/operators/tick_calendar.cc
```diff
@@ -0,0 +1,111 @@
...
+                  break;
+                }
+
+                // Check weekday match
+                if (tm_struct.tm_wday >= min_wday &&
+                    tm_struct.tm_wday <= max_wday) {
+                  ticks.push_back(time);
+                }
+              } else {
+                // Invalid date (end of month)
```
achoum, line 71: Can you add comment of when this is expected?

temporian/core/operators/tick_calendar.py
```diff
@@ -0,0 +1,287 @@
...
+
+    # Always set second=0 by default
+    min_second, max_second = set_arg_range(second, (0, 59), prefer_free=False)
+
+    # prefer_free becomes True when next args should be set to '*' by default
+    # e.g: user sets only hour=1 -> second=0,minute=0, mday='*', month='*'
+    prefer_free = second is not None
+    min_minute, max_minute = set_arg_range(minute, (0, 59), prefer_free)
+
+    prefer_free = prefer_free or minute is not None
```
achoum, line 259: prefer_free |= minute is not None

temporian/implementation/numpy/operators/test/tick_calendar_test.py
```diff
@@ -0,0 +1,177 @@
...
+from temporian.implementation.numpy.operators.tick_calendar import (
+    TickCalendarNumpyImplementation,
+)
+from temporian.implementation.numpy.operators.test.test_util import (
+    assertEqualEventSet,
+    testOperatorAndImp,
+)
+
+
+class TickCalendarOperatorTest(absltest.TestCase):
```
achoum, line 30: It would also be interesting to test invalid situation e.g. seconds=-1, seconds="hello", seconds=60 :)

achoum (commented):
This is great.
Some comments :)

temporian/core/event_set_ops.py
```diff
@@ -2180,6 +2180,142 @@ def tick(
 
         return tick(self, interval=interval, align=align)
 
+    def tick_calendar(
+        self: EventSetOrNode,
+        second: Union[int, str, None] = None,
```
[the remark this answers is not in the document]
DonBraulio: Nice one, didn't know it!

temporian/core/event_set_ops.py
```diff
@@ -2180,6 +2180,142 @@ def tick(
...
+        is equivalent to:
+        `tick_calendar(second=0, minute=0, hour='*', day_of_month='*', month='*')`
+        , resulting in one tick at every exact hour of every day/month/year in
+        the input guide range.
+
+        Examples:
+            ```python
+            >>> # Every day in the period (exactly one year)
+            >>> a = tp.event_set(timestamps=["2021-01-01", "2021-12-31 23:59:59"])
+            >>> b = a.tick_calendar()
```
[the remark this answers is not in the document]
DonBraulio: Yeah, we discussed the empty args case in the meeting, but I don't think there's much to gain allowing this case.

The example could become either `hours=0` or `mdays='*'`.

temporian/core/event_set_ops.py
```diff
@@ -2180,6 +2180,142 @@ def tick(
...
+            >>> b
+            indexes: ...
+            events:
+                (365 events):
+                    timestamps: [...]
```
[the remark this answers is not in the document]
DonBraulio: Totally agree.
May implement in future PR because it might require changing further doctests.

temporian/core/operators/base.py
```diff
@@ -26,7 +26,7 @@
 
 # Valid types for operator attributes
 AttributeType = Union[
-    str, int, float, bool, bytes, List[str], Dict[str, str], List[DType]
+    None, str, int, float, bool, bytes, List[str], Dict[str, str], List[DType]
```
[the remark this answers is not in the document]
DonBraulio: Sorry, this was needed in a previous implementation, but is no longer used.

temporian/implementation/numpy/operators/tick_calendar.py
DonBraulio: You're right, since this means closed interval, it makes sense to allow a single value.

temporian/core/operators/tick_calendar.py
DonBraulio: The serialization code would be more complex if we want to allow both `'*'` and `int` for the attribute types.
We'd need to have an optional flag (e.g. `is_wildcard`) plus an optional integer. Or use `-1` as a special value which may collide with actual values in the future (e.g: if we allow `mday=-1`).
Also, if we want to support ranges in the future, we'd also need to include more optional attributes.

So, I think that saving the range is actually simpler and more generic than  what is currently supported by the api. We can always add optional attributes later for other use cases, right?
Let's discuss it tomorrow.

temporian/core/event_set_ops.py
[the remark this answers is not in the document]
achoum: I still see the old version i.e. `second: Union[int, str, None] = None`
Maybe you forgot to sync?

temporian/core/event_set_ops.py
```diff
@@ -2387,6 +2387,146 @@ def tick(
 
         return tick(self, interval=interval, align=align)
 
+    def tick_calendar(
+        self: EventSetOrNode,
+        second: Union[int, Literal["*"], None] = None,
```
achoum, line 2392: Use "Optional" instead of Union[..., None].

achoum:
Re-request review when needed

ianspektor:
Please update changelog with new op + any other changes

temporian/implementation/numpy_cc/operators/tick_calendar.cc
```diff
@@ -0,0 +1,113 @@
...
+            while (second <= max_second && in_range) {
+              std::tm tm_struct = {};
+              tm_struct.tm_year = year;      // Since 1900
+              tm_struct.tm_mon = month - 1;  // zero-based
+              tm_struct.tm_mday = mday;
+              tm_struct.tm_hour = hour;
+              tm_struct.tm_min = minute;
+              tm_struct.tm_sec = second;
+
+              std::time_t time = std::mktime(&tm_struct);
```
achoum, line 57: s/std::time_t/const std::time_t

Adding "const" to variables that are not expected to change can help readability :)

temporian/core/event_set_ops.py
achoum: Please fix ?

achoum:
Can you "resolve" the conversations that are addressed?

temporian/core/event_set_ops.py
DonBraulio: This is already fixed, for some reason github shows the old version. The new one is here:

https://github.com/google/temporian/pull/255/files#diff-c533efec938d9d3b45600f0d233ca722232b26f02e2e8196f680c49c752733e7R2390

achoum (approved):
Awesome

PR #256, opened 2023-09-10
Implementation of the "until_next" operator

achoum:


temporian/core/event_set_ops.py
```diff
@@ -2420,3 +2420,55 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
...
+            >>> b = tp.event_set(timestamps=[1, 12, 21, 22, 42])
+            >>> c = a.until_next(sampling=b, timeout=5)
+            >>> c
+            indexes: []
+            features: [('until_next', float64)]
+            events:
+                (5 events):
+                    timestamps: [ 1. 12. 12. 21. 35.]
+                    'until_next': [ 1.  2.  1.  1. nan]
+            memory usage: 0.6 kB
```
ianspektor, line 2460: replace with `...`

temporian/core/event_set_ops.py
```diff
@@ -2420,3 +2420,55 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
...
+
+            ```
+
+        Args:
+            sampling: EventSet to use the sampling of.
+            timeout: Maximum amount of time to wait. If not sampling is observed
```
ianspektor, line 2466: typo, If no sampling

temporian/core/event_set_ops.py
```diff
@@ -2420,3 +2420,55 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
         from temporian.core.operators.unique_timestamps import unique_timestamps
 
         return unique_timestamps(self)
+
+    def until_next(
```
ianspektor, line 2424: add new op to changelog

temporian/core/operators/until_next.py
```diff
@@ -0,0 +1,113 @@
...
+    input: EventSetOrNode,
+    sampling: EventSetNode,
+    timeout: Duration,
+) -> EventSetOrNode:
+    assert isinstance(input, EventSetNode)
+
+    return UntilNext(
+        input=input,
+        sampling=sampling,
+        timeout=normalize_duration(timeout),
```
ianspektor, line 112: is there any reason to call `normalize_duration` here instead of in the class's constructor? I moved it to the constructor in the window ops

temporian/core/operators/until_next.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self.add_attribute("timeout", timeout)
+        self._timeout = timeout
+
+        if not math.isfinite(timeout):
+            raise ValueError(
+                f"Timeout should be finite. Instead, got {timeout}"
+            )
+
+        if timeout <= 0:
```
ianspektor, line 58: `normalize_duration` checks this already 👀 I think that's maybe the reason I moved it inside in the window ops (especially since we are sometimes testing the class without the API function)

temporian/core/operators/until_next.py
```diff
@@ -0,0 +1,113 @@
...
+                indexes=input.schema.indexes,
+                is_unix_timestamp=input.schema.is_unix_timestamp,
+                creator=self,
+            ),
+        )
+
+        self.check()
+
+    @property
+    def timeout(self) -> float:
```
ianspektor, line 76: return NormalizedDuration

temporian/implementation/numpy/operators/until_next.py
```diff
@@ -0,0 +1,72 @@
...
+
+        timeout = self.operator.timeout
+
+        # Create output EventSet
+        output_evset = EventSet(data={}, schema=output_schema)
+
+        empty_timestamps = np.array([], dtype=np.float64)
+
+        # Fill output EventSet's data
+        for index_key, index_data in input.data.items():
```
ianspektor, line 49: this behavior differs from the window operators right? window ops return the same index values as `sampling`, this will return the same index values as `input`?

ianspektor (approved):

temporian/core/event_set_ops.py
achoum: *Done

temporian/core/event_set_ops.py
```diff
@@ -2420,3 +2420,55 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
...
+        `until_next` is different from `since_last` in that `since_last` returns
+        one value for each sampling (sampling events are after input events),
+        while `until_next` returns one value for each input value (here again,
+        sampling events are after input events).
+
+        In addition, while  the output [`EventSet`][temporian.EventSet] of
+        `until_next` has the same sampling as the `sampling` argument (i.e.,
+        same number of events, aligned timestamps). In `until_next`, the output
+        [`EventSet`][temporian.EventSet] as the same number of events as the
+        `input` argument but with a different sampling.
```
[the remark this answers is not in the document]
achoum: Good point.
*Applied suggested text

temporian/core/event_set_ops.py
```diff
@@ -2420,3 +2420,55 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
...
+
+        `until_next` is useful to measure the time it takes for an issue
+        (`input`) to be detected by an alert (`sampling`).
```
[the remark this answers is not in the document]
achoum: *Replaced with proposed text in other comment

temporian/core/operators/until_next.py
achoum: Nice.
*Remove unnecesary test

temporian/core/operators/until_next.py
achoum: The "timeout" argument is tagged as FLOAT_64 in the op.
It means the constructor should receive a float i.e., the normalization should be done before.

temporian/implementation/numpy/operators/until_next.py
achoum: Yes, exactly. This op generate one event for each event in "input" => The output has the same index values as the input.

achoum:
Thanks

temporian/core/operators/until_next.py
achoum: Yes.

I think that informally, my concern is that checking should be done after normalization.
Imagine the following bad case:

- User feed some input
- Input is checked and validated.
- Input is normalized to something else.
- Normalized input is serialized.
- Later, normalized input is deserialized.
- Normalized input is checked and **invalid**
- The user is blocked.

Instead, it is better to have:

- User feed some input
- Input is normalized to something else.
- Normalized input is checked and **invalid**

This is especially problematic because while normalized input are very constrained, user inputs can be a lot of things.

Having the operator class being super restrictive (i.e., only allowing normalized inputs) and doing the checks is a way to ensure we are always in the second case.

If you are not convinced, let me know. I am happy to update this PR.

ianspektor: Okay, got the idea, but I'd argue that having all the op's logic centralized in the class would be desirable. Although if we very clearly define what goes in the function and what goes in the class that'd be fine by me too.

> Normalized input is checked and invalid

When/why would this happen? I think this is the main issue - shouldn't normalized inputs always pass any check we're doing on them? On second thought, I guess this could be a problem if the normalization we're doing isn't idempotent (since we'd be applying it twice, once when op is instantiated and again when its deserialized), but it should be.

Another option would be to have the op's main constructor do normalization, and have a secondary one that is used by the deserialization code that skips the normalization and checks, assuming they will pass.

achoum: >> On second thought, I guess this could be a problem if the normalization we're doing isn't idempotent 

That is the issue :). We have really poor test coverage for that (I suspect we actually never test it).

>> Another option would be to have the op's main constructor do normalization, and have a secondary one that is used by the deserialization code that skips the normalization and checks, assuming they will pass.

There can only be one constructor in python.

Another alternative is to unit test operator stability to serialization. For instance, we could unit test that an operator attributes are the same with and without being serialized+deserialized.

ianspektor: Feel free to change the window ops to do the normalization in the function call too to keep it consistent

PR #260, opened 2023-09-13
Show datetimes instead of float on unix timestamps

DonBraulio:


temporian/implementation/numpy/data/display_utils.py
```diff
@@ -340,10 +343,31 @@ def display_html_header(dom: Dom, evset: EventSet) -> Html:
...
+    # Create string format with precision, e.g "{:.6g}"
+    float_template = "{:.%d%s}" % (config.print_precision, "g")
+    return float_template.format(value)
+
+
 def display_text(evset: EventSet) -> str:
     # Configs and defaults
     max_events = config.print_max_events or sys.maxsize  # see np.printoptions
     max_indexes = config.print_max_indexes  # 0 will print all
+    is_datetime = evset.schema.is_unix_timestamp
```
achoum, line 370: s/is_datetime/is_unix_timestamp

However, do we need to create this variable?

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -340,10 +343,31 @@ def display_html_header(dom: Dom, evset: EventSet) -> Html:
     return root
 
 
+def repr_value_html(value: Any, dtype: DType) -> str:
```
achoum, line 346: Any reason this block was moved?

achoum (commented):
This is great :).
Left a few comments

temporian/implementation/numpy/data/display_utils.py
DonBraulio: I moved the text related methods together, below the html ones. They were all over the place.

temporian/implementation/numpy/data/display_utils.py
```diff
@@ -365,9 +389,16 @@ def display_text(evset: EventSet) -> str:
             ):
                 index_key_repr.append(f"{index_name}={index_value}")
             index_key_repr = " ".join(index_key_repr)
+            timestamps = index_data.timestamps
+            if is_datetime:
+                timestamps = np.array(
+                    convert_timestamps_to_datetimes(timestamps),
```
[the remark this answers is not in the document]
DonBraulio: You're right, actually there's no need to normalize these timestamps, they're part of the EventSet and should be float64 (otherwise, we've other greater issues).
Just calling `timestamps.astype('datetime64[s]')` is enough. It converts the whole array but that's not too expensive. Converting only the displayed events is not trivial and would require a new test (overkill, imo).

achoum (approved):
Awesome. Thanks

PR #265, opened 2023-09-20
Operator tp.where()

DonBraulio:
Also, changes in `normalize_features`:
 - Fixed a bug in which array features of type `np.str_` (unicode strings) were not casted to `np.bytes_`
 - Added a message if conversion fails due to invalid unicode characters (also for `np.object_` ).
 - `np.datetime` gets converted to `float64` instead of `int64`, to keep it consistent with timestamps handling.


ianspektor:
Changelog 🔔

temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -114,6 +114,21 @@ def normalize_features(
     `normalize_features` should match `_DTYPE_MAPPING`.
     """
 
+    def _to_bytes_array(
+        feat_array: Union[Tuple, List, np.ndarray]
+    ) -> np.ndarray:
+        """Try to convert string/object array to np.bytes or fail"""
+        try:
+            if isinstance(feat_array, np.ndarray):
+                return feat_array.astype(np.bytes_)
```
achoum, line 123: Will this duplicate the data in memory if the original data is already bytes?

temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -114,6 +114,21 @@ def normalize_features(
...
+            return np.array(feat_array, dtype=np.bytes_)
+        except UnicodeEncodeError as e:
+            raise ValueError(
+                f"Feature {name} contains non-supported unicode characters."
```
achoum, line 127: In your example, the issue was with non-ascii characters (!= non-unicode).
Maybe we can add support for unicode characters easily?

achoum (commented):
This is awesome

temporian/core/event_set_ops.py
```diff
@@ -2475,3 +2475,84 @@ def unique_timestamps(self: EventSetOrNode) -> EventSetOrNode:
         from temporian.core.operators.unique_timestamps import unique_timestamps
 
         return unique_timestamps(self)
+
+    def where(
+        self: EventSetOrNode,
+        on_true: Union[EventSetOrNode, Any],
```
ianspektor, line 2481: Should we create an alias for `Union[int, float, str, bool]`? we have `T_SCALAR` defined in event_set_ops which is only (int, float), but int that same file in `__ne__`'s def we join it with `(bool, str)` too. Maybe this definition could improve the signature's legibility

temporian/core/operators/where.py
```diff
@@ -0,0 +1,149 @@
...
+                pb.OperatorDef.Input(key="on_true", is_optional=True),
+                pb.OperatorDef.Input(key="on_false", is_optional=True),
+            ],
+            outputs=[pb.OperatorDef.Output(key="output")],
+        )
+
+    @property
+    def on_true(self) -> Any:
+        """Returns None if 'on_true' is EventSet instead of a single value"""
+        return self._on_attributes.get("on_true", None)
```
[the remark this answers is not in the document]
ianspektor: I'd argue that accessing a 2-item dict is cheap enough for a property, but I'd remove the dict altogether and store the 2 vars separately (`self._on_true` and `self._on_false`, `Optional[any]`)

temporian/core/operators/window/base.py
```diff
@@ -68,6 +68,7 @@ def __init__(
                     "`window_length` must have exactly one float64 feature."
                 )
             self.add_input("window_length", window_length)
+            self._window_length = None
```
ianspektor, line 71: good catch :) and great example of the problems associated with having tons of tests on the implementation and not so tons of them on the core op

temporian/core/operators/where.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @classmethod
+    def build_op_definition(cls) -> pb.OperatorDef:
+        return pb.OperatorDef(
+            key="WHERE",
+            attributes=[
+                pb.OperatorDef.Attribute(
+                    key="on_true",
+                    type=pb.OperatorDef.Attribute.Type.ANY,
+                    is_optional=True,
```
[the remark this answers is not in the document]
DonBraulio: Yes. I don't think it's obvious what value to use by default if some of them is not provided.

temporian/implementation/numpy/data/dtype_normalization.py
DonBraulio: It would, but this function is not called in that case. 
Good catch tho, because according to the docs, `astype` always copies the input array. I'll add a comment.

temporian/core/event_set_ops.py
DonBraulio: I'm not sure about this. `T_SCALAR` is actually a tuple, we're not using it in typehints.

This would make sense if we define types in `typing.py` like `PythonValue` and `PythonScalarValue` and use it consistently in the event_set_ops types.
But I think we should discuss it previously, I would not add it in this PR.

achoum (approved):
Awesome

PR #274, opened 2023-10-03
Protubuf and pandas versions relaxed

DonBraulio:
This fixes the protobuf warnings on colab notebooks, and allows using Pandas >= 2.
Also, local builds in linux container where not working because poetry was running `build.py` using python 2.7 instead of the configured through `PYTHON_VERSION`, fixed by prepending current python `/bin` into the `PATH` env (also simplified calling `pip` instead of using `$PIP_BIN`).


ianspektor (approved):

pyproject.toml
```diff
@@ -35,11 +35,13 @@ generate-setup-file = true
 [tool.poetry.dependencies]
 python = "^3.8,<3.12"
 absl-py = "^1.3.0"
-protobuf = "^4.21.12"
-pandas = "^1.5.2"
+protobuf = ">=3.12.2"
```
rstz, line 38: Is there a reason to go this low? IIRC, protobuf-generated code changed with the release of 3.20 (or 3.19), so I'd make sure this does not cause any incompatibility. Since Google-internally, we mostly live at head, it's useful to not stray too far from it.

rstz (commented):
LGTM, one question

pyproject.toml
DonBraulio: Very good point. According to poetry.lock, tensorflow and apache-beam depend on `>=3.20.3`, so let's use that.

rstz (approved):

PR #276, opened 2023-10-04
Fix longlong timestamps

DonBraulio:
**TL;DR:** This PR fixes some issues in `normalize_timestamps()`, simplifying the code, avoiding issues with numpy dtypes, and improving the error handling by avoiding unnecessary checks.

For context, this fixes an issue that showed up in a notebook by @achoum, caused by numpy representing numbers using `np.longlong` internally instead of the usual `np.int64`, after loading a CSV from pandas. This can be reproduced by doing:

```
>>> np.array([1, 2], dtype=np.longlong).dtype
dtype('int64')
>>> np.array([1, 2], dtype=np.longlong).dtype.type
<class 'numpy.longlong'>
```

The bug is solved by using `np.issubdtype(np.dtype, np.integer)`, instead of accessing `my_array.dtype.type` which refers to the underlying scalar type, and there are a bunch of those (see [Built in scalar types](https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types)).

### DIFF
--- a/temporian/implementation/numpy/data/dtype_normalization.py
+++ b/temporian/implementation/numpy/data/dtype_normalization.py
@@ -15,17 +15,13 @@
 """Shared data type normalization functions."""
 
 from __future__ import annotations
-import datetime
 import logging
-from typing import TYPE_CHECKING, Any, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
 import numpy as np
 
 from temporian.core.data.dtype import DType
-from temporian.core.data.duration_utils import (
-    MIN_TIMESTAMP_S,
-    datetime64_array_to_float64,
-)
+from temporian.core.data.duration_utils import datetime64_array_to_float64
 
 if TYPE_CHECKING:
     from temporian.core.typing import (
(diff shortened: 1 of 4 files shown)


temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -177,8 +180,10 @@ def normalize_timestamps(
         values = np.array(values)
 
     # values is represented as a number. Cast to float64.
-    if values.dtype.type in [np.float32, np.int64, np.int32]:
-        values = values.astype(np.float64)
+    if np.issubdtype(values.dtype, np.integer) or np.issubdtype(
+        values.dtype, np.floating
+    ):
+        values = values.astype(np.float64, copy=False)
```
achoum, line 186: I would do a copy here.

temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -187,40 +192,20 @@ def normalize_timestamps(
...
-        elif all(
-            str(type(x)) == "<class 'pandas._libs.tslibs.timestamps.Timestamp'>"
-            for x in values
-        ):
-            values = np.array([x.to_numpy() for x in values])
-
     if values.dtype.type == np.datetime64:
         # values is a date. Cast to unix epoch in float64 seconds.
         values = datetime64_array_to_float64(values)
+        if np.any(values < MIN_TIMESTAMP_S):
```
achoum, line 202: This might cause some issues (We have datasets where the unix timestamp is negative).
Could this be checked differently?

achoum (approved):
Awesome.
A few nits

temporian/implementation/numpy/data/dtype_normalization.py
DonBraulio: Actually this `MIN_TIMESTAMP_S` is a negative number. It's kind of arbitrarily set to `1900-01-01`, which seemed like a safe date span 😄

temporian/implementation/numpy/data/dtype_normalization.py
DonBraulio: Do we also want to copy `float64` inputs to decouple from user's data?

PR #277, opened 2023-10-05
Recipe with tick_calendar and since_last(steps=2)

DonBraulio:
* Added a recipe showcasing how to use `tick_calendar` to accumulate monthly sales, and also `since_last` with steps.
* Moved some unary operators (log, isnan, etc) to `evset.isnan()`. Didn't move `invert()` because it's already available as `~evset`.


achoum (approved):
Nice

PR #278, opened 2023-10-06
Map operator

ianspektor:


temporian/core/event_set_ops.py
```diff
@@ -1416,6 +1414,100 @@ def leak(self: EventSetOrNode, duration: Duration) -> EventSetOrNode:
 
         return leak(self, duration=duration)
 
+    def map(
+        self: EventSetOrNode,
+        func: MapFunction,
+        output_dtypes: Optional[TargetDtypes] = None,
+    ) -> EventSetOrNode:
+        """Applies a function on each of an [`EventSet`][temporian.EventSet]'s
+        values.
```
achoum, line 1423: Nit: How to make it clear that a "value" is a feature-value for a single event?

temporian/core/event_set_ops.py
```diff
@@ -1416,6 +1414,100 @@ def leak(self: EventSetOrNode, duration: Duration) -> EventSetOrNode:
...
+            ...     features={"a": [10, 20, 30], "b": ["100", "200", "300"]},
+            ... )
+
+            >>> def f(value):
+            ...     if value.dtype == np.int64:
+            ...         return float(value) + 1
+            ...     else:
+            ...         return int(value) + 2
+
+            >>> b = a.map(f, output_dtypes={"a": float, "b": int})
```
achoum, line 1462: Awesome

temporian/core/event_set_ops.py
```diff
@@ -1416,6 +1414,100 @@ def leak(self: EventSetOrNode, duration: Duration) -> EventSetOrNode:
...
+                single dtype, all features will be expected to have that dtype.
+                If a mapping, the keys can be either feature names or the
+                input dtypes (and not both types mixed), and the values are the
+                target dtypes for them. All dtypes must be Temporian types (see
+                `dtype.py`).
+
+        Returns:
+            EventSet with the function applied on each value.
+        """
+        from temporian.core.operators.map import map
```
achoum, line 1507: In this function, I would use another name to avoid collision with python's map e.g.

from temporian.core.operators.map import map  as tp_map

temporian/core/operators/map.py
```diff
@@ -0,0 +1,153 @@
...
+            outputs=[pb.OperatorDef.Output(key="output")],
+        )
+
+
+operator_lib.register_operator(Map)
+
+
+@typecheck
+@compile
+def map(
```
achoum, line 134: Nit: On one side, it is great this function as the name of the operator / method name, on the other side, "map" is a commonly used python function.

temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -236,3 +238,103 @@ def normalize_index_key_list(
     normalized_indexes = [normalize_index_key(x) for x in normalized_indexes]
 
     return normalized_indexes
+
+
+def normalize_dtype(x: Any) -> DType:
```
achoum, line 243: What about using this map (https://github.com/google/temporian/blob/main/temporian/core/data/dtype.py#L91), and possibly utility function (if we have one)?

temporian/implementation/numpy/data/dtype_normalization.py
```diff
@@ -236,3 +238,103 @@ def normalize_index_key_list(
...
+    if x == float:
+        return DType.FLOAT64
+    if x == str:
+        return DType.STRING
+    if x == bool:
+        return DType.BOOLEAN
+    raise ValueError(f"Cannot normalize {x!r} as a DType.")
+
+
+def normalize_target_dtypes(
```
achoum, line 257: Add documentation

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,97 @@
...
+        assert isinstance(operator, Map)
+        super().__init__(operator)
+
+    def __call__(self, input: EventSet) -> Dict[str, EventSet]:
+        assert isinstance(self.operator, Map)
+
+        output_schema = self.output_schema("output")
+
+        func = self.operator.func
+        receives_extras = len(signature(func).parameters) > 1
```
achoum, line 45: What about computing "receives_extras" being an attribute of the op definition? (instead of being re-computed at execution time)

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,97 @@
...
+                    if receives_extras:
+                        args.append(
+                            MapExtras(
+                                index_key=index_key,
+                                timestamp=index_data.timestamps[i],
+                                feature_name=feature_schema.name,
+                            )
+                        )
+                    res = func(*args)
+                    output_values.append(res)
```
achoum, line 71: This is the expensive loop.
Can we create the "MapExtras" outside of the loop, and avoid creating / destroying any object in it (e.g. "args", "MapExtras")?

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,97 @@
...
+                output_schema.feature_dtypes(),
+            ):
+                output_values = []
+                for i, value in enumerate(orig_feature):
+                    args = [value]
+                    if receives_extras:
+                        args.append(
+                            MapExtras(
+                                index_key=index_key,
+                                timestamp=index_data.timestamps[i],
```
achoum, line 66: Nit: "timestamps" in enumerate?

temporian/core/operators/map.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+        # output_dtypes is an array of output dtype for all the input features,
+        # in the same order
+        output_dtypes = build_dtypes_list_from_target_dtypes(
+            input, dtype, dtype_to_dtype, feature_name_to_dtype
+        )
+        assert len(output_dtypes) == len(input.schema.features)
+
+        if len(signature(func).parameters) > 2:
```
achoum, line 87: What if the function takes 0 args?

temporian/test/utils.py
```diff
@@ -80,12 +80,15 @@ def assertOperatorResult(
...
+    try:
+        serialized_op = serialization._serialize_operator(op)
+        nodes = {}
+        for node in op.inputs.values():
+            nodes[serialization._identifier(node)] = node
+        for node in op.outputs.values():
+            nodes[serialization._identifier(node)] = node
+
+        _ = serialization._unserialize_operator(serialized_op, nodes)
+    except ValueError as e:
```
achoum, line 92: To be more fancy, I would create a boolean field "is_serializable" in the proto definition of the op: https://github.com/google/temporian/blob/main/temporian/proto/core.proto#L187

Then, we can have a utility function (like "has_leak") that check if a graph is serializable without relying on exceptions (since there might be other reasons for an exception).

temporian/core/event_set_ops.py
ianspektor: reworded to "Applies a function on each value of an [`EventSet`][temporian.EventSet]'s features."

temporian/core/operators/map.py
ianspektor: I'd keep it as-is (user can rename the symbol on import if needed as I did in event_set_ops)

temporian/test/utils.py
ianspektor: Would it make sense to make it a class variable instead? Then we could define it as true in the base operator and have the non-serializable ones override it to false. Or is there a reason to store this info in the proto itself?

temporian/implementation/numpy/operators/map.py
ianspektor: removed unnecessary creation of args dict. not sure how we could create the MapExtras outside the loop, since it uses the timestamp of the event that we are looping over

temporian/core/event_set_ops.py
achoum: Nice.

Could you also add that:
- This function will be slow. When possible existing operators should be used.
- A temporian graph with a map cannot be serialized.

temporian/implementation/numpy/operators/map.py
achoum: You could create an MapExtras, and update the timestamp at each loop.

temporian/test/utils.py
achoum: Having the definition in the proto means we can operator on the ops outside of the python language, which is a good option to have. This is why we have an operator def as a proto.

temporian/core/operators/test/test_map.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def test_too_many_args(self):
+        evset = event_set(timestamps=[1], features={"a": [2]})
+
+        with self.assertRaisesRegex(
+            ValueError, "`func` must receive 1 or 2 arguments."
+        ):
+            evset.map(lambda v, e, z: v + e.timestamp)
+
+    def test_too_little_args(self):
```
achoum, line 129: What is the function has optional arguments e.g. the "round" function.

evset.map(round)

Is the user forced to wrap the function (and, is there a risk of mistake)?

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,100 @@
...
+                                value,
+                                MapExtras(
+                                    index_key=index_key,
+                                    timestamp=timestamp,
+                                    feature_name=feature_schema.name,
+                                ),
+                            )
+                        )
+                    else:
+                        assert len(signature(func).parameters) == 1
```
achoum, line 73: Move outside of the loop

achoum (approved):

temporian/implementation/numpy/operators/map.py
achoum: Let's document it then :)
I think the gain is worth it.

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,100 @@
...
+            for feature_schema, orig_feature, output_dtype in zip(
+                input.schema.features,
+                index_data.features,
+                output_schema.feature_dtypes(),
+            ):
+                output_values = []
+                for value, timestamp in zip(
+                    orig_feature, index_data.timestamps
+                ):
+                    if receives_extras:
```
achoum, line 61: This is still a lot of work inside of the for loop: the append, the creation of the MapExtra, and the if would all be move outside of the loop.

temporian/implementation/numpy/operators/map.py
```diff
@@ -0,0 +1,100 @@
...
+                                MapExtras(
+                                    index_key=index_key,
+                                    timestamp=timestamp,
+                                    feature_name=feature_schema.name,
+                                ),
+                            )
+                        )
+                    else:
+                        assert len(signature(func).parameters) == 1
+                        output_values.append(func(value))
```
achoum, line 74: Bonus: Pre-allocate the list (or numpy array).

PR #281, opened 2023-10-10
Improve error messages when feeding wrong type to moving window op

achoum:


temporian/core/operators/window/moving_max.py
```diff
@@ -31,6 +31,12 @@ def operator_def_key(cls) -> str:
         return "MOVING_MAX"
 
     def get_feature_dtype(self, feature: FeatureSchema) -> DType:
+        if not feature.dtype.is_numerical:
+            raise ValueError(
+                "moving_max requires numerical point inputs."
+                " Instead, get feature {feature.name!r} with type"
```
ianspektor, line 37: s/get/got or received (same in the other messages)

ianspektor: actually, would rephrase to "moving_max requires the EventSet to contain numerical features only, but received feature {} with type {}" (same in the other ones)

temporian/core/operators/window/test/test_moving_sum.py
```diff
@@ -197,6 +197,14 @@ def test_with_variable_winlen_same_sampling_uses_correct_cpp_impl(
             window_length=window_length.data[()].features[0],
         )
 
+    def test_error_input_bytes(self):
+        evset = event_set([1, 2], {"f": ["A", "B"]})
+        with self.assertRaisesRegex(
+            ValueError,
+            "moving_sum requires numerical point inputs",
+        ):
+            _ = evset.moving_sum(1)
```
ianspektor, line 206: missing test for moving_std, moving_min, moving_max (we will migrate the old tests soon, but we might forget to add these if they aren't there already)

ianspektor (approved):
Much needed improvement, thanks! 👍🏼

temporian/core/operators/window/test/test_moving_sum.py
achoum: Good point.
*Adding min/max/sd tests

PR #282, opened 2023-10-17
Calendar ops tests

ianspektor:


.vscode/settings.json
```diff
@@ -25,7 +25,15 @@
     "**/.cache/**": true,
     "**/temporian.egg-info/**": true,
   },
-  "python.formatting.provider": "black",
+  "black-formatter.interpreter": [
+    "${workspaceFolder}/.venv/bin/python"
```
achoum, line 29: What about adding the instructions to setup the workspace (virtual-env?) for this to work?

achoum (approved):
Nice

.vscode/settings.json
ianspektor: they were in the CONTRIBUTING guide already but as an optional step, reworded to make it mandatory

PR #284, opened 2023-10-18
fix calendar tests

ianspektor:
- removed calendar tests from event_set_ops_test. added this step in the test restructure notion page.
- added __init__ file to the new calendar ops test package, it wasn't being found by unittest (this is why my previous PR had a -0.01 coverage diff and why I looked into it). added this step to the same notion page.


DonBraulio (approved):
👍

PR #286, opened 2023-10-18
Select_index_values and rename tests

ianspektor:


temporian/core/test/event_set_ops_test.py
```diff
@@ -191,10 +191,6 @@ def test_propagate(self):
             )
```
DonBraulio, line 191: I see that we also have a `test_propagate` here, we should remove it too, right?

DonBraulio (approved):
LGTM, one minor comment.

PR #287, opened 2023-10-18
Update arithmetic_multi_index test to new structure

DonBraulio:
Checking that I'm following correctly the new structure 😄


temporian/core/operators/test/test_arithmetic_multi_index.py
```diff
@@ -0,0 +1,135 @@
...
+
+# store ids
+TRYOLABS_SHOP = 42
+GOOGLE_SHOP = 101
+# product ids
+MATE_ID = 1
+BOOK_ID = 2
+PIXEL_ID = 3
+
+class ArithmeticMultiIndexNumpyImplementationTest(absltest.TestCase):
```
ianspektor, line 30: rename test

temporian/core/operators/test/test_arithmetic_multi_index.py
```diff
@@ -0,0 +1,135 @@
...
+
+from temporian.test.utils import f32, f64, assertOperatorResult
+
+# store ids
+TRYOLABS_SHOP = 42
+GOOGLE_SHOP = 101
+# product ids
+MATE_ID = 1
+BOOK_ID = 2
+PIXEL_ID = 3
```
ianspektor, line 28: i'd substitute these for just `1` and `2` in the tests themselves. no need to add complexity in the tests.

temporian/core/operators/test/test_arithmetic_multi_index.py
```diff
@@ -0,0 +1,135 @@
...
+            timestamps=[0, 1, 2, 1, 1, 2, 0, 1],
+            features={
+                "store_id": [TRYOLABS_SHOP] * 4 + [GOOGLE_SHOP] * 4,
+                "product_id": [MATE_ID] * 3 + [BOOK_ID] * 3 + [PIXEL_ID] * 2,
+                "f3": f64([4.5, -5.5, 16, 0, 3, -8, 3, 5]),
+                "f4": f32([5, 3, 1, 2, -8, 2, 4, 3]),
+            },
+            indexes=["store_id", "product_id"],
+            same_sampling_as=self.evset_1
+        )
```
ianspektor, line 58: same here, I'd try and simplify as much as possible (remove store var names, remove list multiplications, halve the number of events?). Some of the olds tests are unnecessarily complex. See for example how I simplified `rename`'s tests here when compared to the old ones: https://github.com/google/temporian/pull/286/files#diff-a9119b805f6d05d7d9950728b1aae5e51ac5ce1a6cb5a0c1a917de1673588bae

ianspektor (approved):
LGTM after comments are addressed 🙌🏼 thanks!

temporian/core/operators/test/test_arithmetic_multi_index.py
DonBraulio: Agreed. didn't reduce the number of events though, because they are useful to test different scenarios in all arithmetic ops.

PR #288, opened 2023-10-19
Fix tests arithm scalar

DonBraulio:
⚠️ **Important:**
 - Moved `tp.equal()` to `evset.equal(other_evset / scalar_value)`.

 - Also: **Removed unary operators from `__init__`** (`tp.log()`, `tp.abs()`, `tp.isnan()`), since they were already in `event_set_ops`.
 
Tests restructured in this PR:
- [x] Arithmetic scalar tests
- [x] Arithmetic tests (evsets)
- [x] Unary tests
- [x] Logical tests
- [x] Relational / relational scalar tests 

For a future PR: we should probably remove ops. that are already overloaded from `__init__`, such as `tp.greater_equal()`. Unless there a good reason to have 2 ways to do that.


docs/src/reference/index.md
```diff
@@ -78,11 +83,6 @@ Check the index on the left for a more detailed description of any symbol.
 | [`tp.equal()`][temporian.equal] [`tp.not_equal()`][temporian.not_equal] [`tp.greater()`][temporian.greater] [`tp.greater_equal()`][temporian.greater_equal] [`tp.less()`][temporian.less] [`tp.less_equal()`][temporian.less_equal]               | Compute a relational binary operator between two [`EventSets`][temporian.EventSet].                                         |
```
ianspektor, line 83: `tp.equal()` no longer exists right? I'd move all of the binary ones to the EventSet if we're moving `.equal()`. @achoum is there a reason we haven't already done this?

temporian/core/operators/test/test_unary.py
```diff
@@ -0,0 +1,140 @@
...
+                    True,
+                    False,
+                    True,
+                    False,
+                    False,
+                    True,
+                ],
+            },
+            indexes=["store_id", "product_id"],
+        )
```
ianspektor, line 54: remove feature names, make much shorter?

ianspektor (approved):
Great work 👍🏼

docs/src/reference/index.md
DonBraulio: I think the main reason is that we already have the magic methods for those. `evset.equal()` is needed because we can't overload the `==` operator.
IMO we should remove all these  from `__init__`, without adding them to `event_set_ops`. I don't see any use case where using these methods is a much better option, and we should try to not have two ways to do things.

(this was the last point I wanted to discuss today)

temporian/core/operators/test/test_unary.py
DonBraulio: You're right, this is unnecessarily long 😬

docs/src/reference/index.md
ianspektor: LGTM from my side to remove them 👍🏼

achoum (approved):
Great!

PR #290, opened 2023-10-26
upgrade mkdocs deps

ianspektor:
upgraded mkdocstrings-python and griffe as per instructions in these two issues I opened some months ago:
- https://github.com/mkdocstrings/python/issues/90
- https://github.com/mkdocstrings/python/issues/91

issue 91 does look fixed, issue 90 doesn't (full tree path is shown instead of identifier, e.g. `temporian.implementation.numpy.data.event_set.EventSet.begin` is shown as the header on the begin operator's class, even though we're using the `::: temporian.EventSet.begin` identifier for it). will open a new issue to check I'm not missing anything.


ianspektor:
Opened https://github.com/mkdocstrings/python/issues/111 to follow up on https://github.com/mkdocstrings/python/issues/90

achoum (approved):

PR #291, opened 2023-10-26
Restructure tests: add and drop index

ianspektor:


.vscode/settings.json
```diff
@@ -40,9 +40,6 @@
   },
   "editor.tabSize": 4,
   "editor.formatOnSave": true,
-  "python.linting.flake8Enabled": false,
-  "python.linting.pylintEnabled": true,
-  "python.linting.enabled": true,
```
ianspektor, line 45: removed these because they were deprecated

achoum (approved):
Nice

PR #293, opened 2023-10-26
Docs updates

DonBraulio:
⚠️ This PR changes the path of getting_started guide so git-hooks must be updated:
```
cp .git-hooks/* .git/hooks
```

#288 needs to be merged before this (user_guide uses `evset.equal()`).

Main changes:
- Many updates to the user_guide to use eager mode.
- Set getting_started as starting point in the docs, combined the relevant parts of 3_minutes guide (deleted).


docs/src/getting_started.ipynb
```diff
@@ -8,7 +8,17 @@
    "source": [
     "# Getting Started with Temporian\n",
```
ianspektor, line 9: Change to `Getting Started` since the H1 is whats shown in the navbar and its too long as-is
image

ianspektor (approved):

.git-hooks/pre-commit
```diff
@@ -2,7 +2,7 @@
 #
 files_to_check="docs/src/recipes/*.ipynb"
 files_to_check+=" docs/src/user_guide.ipynb"
-files_to_check+=" docs/src/tutorials/getting_started.ipynb"
```
achoum, line 5: Do we risk breaking existing url that we shared?

Maybe we can add a redirection from tutorials/getting_started.ipynb to getting_started.ipynb?

docs/src/getting_started.ipynb
```diff
@@ -8,7 +8,17 @@
...
-    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/temporian/blob/last-release/docs/src/tutorials/getting_started.ipynb)"
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/temporian/blob/last-release/docs/src/tutorials/getting_started.ipynb)\n",
+    "\n",
+    "This guide will introduce you to the basics of Temporian, including:\n",
+    "- What is an **EventSet** and how to create one from scratch.\n",
+    "- Visualizing input/output data using **EventSet.plot()** and interactive plots.\n",
+    "- Converting back and forth between EventSets and pandas **DataFrames**.\n",
+    "- Transforming the EventSets by using **operators**.\n",
+    "- How operators work when using **indexes**.\n",
+    "- Commonly used operations like **glue**, **resample**, **lag**, moving windows and arithmetics.\n",
```
achoum, line 19: Nit: It is easier to read a list like this one when all the items have the same sentence construction e.g. All sentence start with an infinitive verb;

docs/src/getting_started.ipynb
```diff
@@ -8,7 +8,17 @@
    "source": [
     "# Getting Started with Temporian\n",
     "\n",
-    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/temporian/blob/last-release/docs/src/tutorials/getting_started.ipynb)"
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/temporian/blob/last-release/docs/src/tutorials/getting_started.ipynb)\n",
+    "\n",
+    "This guide will introduce you to the basics of Temporian, including:\n",
```
achoum, line 13: Some of the reader will come directly to this page, skipping the readme. I would add a paragraph telling what Temporian is, and where is the main doc.

docs/src/getting_started.ipynb
```diff
@@ -8,7 +8,17 @@
...
+    "- What is an **EventSet** and how to create one from scratch.\n",
+    "- Visualizing input/output data using **EventSet.plot()** and interactive plots.\n",
+    "- Converting back and forth between EventSets and pandas **DataFrames**.\n",
+    "- Transforming the EventSets by using **operators**.\n",
+    "- How operators work when using **indexes**.\n",
+    "- Commonly used operations like **glue**, **resample**, **lag**, moving windows and arithmetics.\n",
+    "\n",
+    "If you're interested in a topic that is not included here, we provide links to other parts of the documentation on the final section, to continue learning."
```
achoum, line 21: I would conclude the introduction with something desirable (e.g. a carrot) about what the reader will get by reading this document. For example, something along the lines of: you will generate good features easily and without having to worry about errors. You models will perform better (maybe less cheesy).

docs/src/getting_started.ipynb
```diff
@@ -51,13 +61,66 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "5e9bf8e3-07fd-4e4d-bdb8-d32a48a366c7",
+   "metadata": {},
+   "source": [
+    "## Part 1: Events and EventSets\n",
+    "\n",
+    "The most basic unit of data in Temporian is an **event**. An event consists of a timestamp and a set of feature values.\n",
```
achoum, line 71: **Events** are the basic unit of data in Temporian. They consist of a timestamp and a set of feature values. Events are not handled individually, but are instead grouped together into **EventSets**.

docs/src/getting_started.ipynb
```diff
@@ -51,13 +61,66 @@
...
+   "outputs": [],
+   "source": [
+    "evset = tp.event_set(\n",
+    "    timestamps=[\"2023-02-04\", \"2023-02-06\", \"2023-02-07\", \"2023-02-07\"],\n",
+    "    features={\n",
+    "        \"feature_1\": [0.5, 0.6, np.nan, 0.9],\n",
+    "        \"feature_2\": [\"red\", \"blue\", \"red\", \"blue\"],\n",
+    "        \"feature_3\":  [10.0, -1.0, 5.0, 5.0],\n",
+    "    },\n",
+    "    indexes=[\"feature_2\"],\n",
```
achoum, line 98: Should we first create an eventset without an index?

docs/src/getting_started.ipynb
```diff
@@ -51,13 +61,66 @@
...
+   "cell_type": "markdown",
+   "id": "effc4483-9a1a-4e21-b376-3ed188ced821",
+   "metadata": {},
+   "source": [
+    "An `EventSet` can hold one or several time sequences, depending on what its `index` is.\n",
+    "\n",
+    "If it has no index, it will hold a single multivariate time sequence, which means that all events will be considered part of the same group and will interact with each other when operators are applied.\n",
+    "\n",
+    "If it has one (or many) indexes, its events will be grouped by their `indexes` values, so it will hold one multivariate time sequence for each unique value (or unique combination of values) of its indexes, and operators will be applied to each time sequence independently.\n",
+    "\n",
```
achoum, line 113: What about:

An EventSet can hold one or several time sequences, depending on its index.
- If it has no index, an EventSet holds a single multivariate time sequence.
- If it has one (or more) indexes, the events are grouped by their index values. This means that the EventSet will hold one multivariate time sequence for each unique value (or unique combination of values) of its indexes. 
Many operators are applied on each time sequence of an EventSet independently. Indexing is the primary way to handle rich and complex databases. For instance, in a retail database, you can index on customers.

docs/src/getting_started.ipynb
```diff
@@ -51,13 +61,66 @@
...
   {
    "attachments": {},
    "cell_type": "markdown",
    "id": "18cc96f7",
    "metadata": {},
    "source": [
-    "## Example Data\n",
+    "### Example Data\n",
     "\n",
     "This minimal data consists of just one `signal` with a `timestamp` for each sample.\n",
```
achoum, line 125: Three features?

achoum (approved):
Awesome

docs/src/user_guide.ipynb
```diff
@@ -46,7 +46,7 @@
     "feature_3: 10\n",
     "```\n",
     "\n",
-    "Events are not handled individually. Instead, events are grouped together into [EventSet][temporian.EventSet]s. When representing an `EventSet`, it is convenient to group similar features together and to sort them according to the timestamps in increasing order.\n",
+    "Events are not handled individually. Instead, events are grouped together into `EventSets`. When creating one, it is convenient to put similar features together and to sort them according to the timestamps in increasing order.\n",
```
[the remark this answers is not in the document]
DonBraulio: Right, the cross link doesn't work but using the url seems ok for the first time.

.git-hooks/pre-commit
DonBraulio: Good idea, I'll create a symlink and see if it works.

DonBraulio: FYI: symlink didn't work, but I could configure a redirect through the rtd admin panel.

PR #297, opened 2023-10-31
Cleanup init

DonBraulio:
Deprecated docstrings on operators, the user guide explains the usage of all these operators in a much more concise way.


ianspektor (approved):
LGTM! Will add the dosctrings to the magics in a follow up PR, so that we keep them + magics show up in the docs

PR #298, opened 2023-11-01
fix notebooks display in docs

ianspektor:
notebook display was broken after https://github.com/google/temporian/pull/290 (great catch by @DonBraulio), this fixes it


DonBraulio:
It's better now but the code cells are behaving weird, right? The code text swaps color according to theme, but the background is always dark (which makes it hard to read with light theme text)

DonBraulio (approved):
Awesome 🙌🙌🙌

PR #299, opened 2023-11-01
Magic methods docs

ianspektor:
Add docstrings to magic methods to make them show up in docs (adapted docs that existed in arithmetic, arithmetic_scalar, logical, etc.)


DonBraulio (approved):
Great work! Thanks for taking the time to update all of these so thoroughly 👏

PR #300, opened 2023-11-01
Tests restructure

DonBraulio:
This should be the last round 😄


temporian/core/operators/test/test_filter_moving_count.py
```diff
@@ -0,0 +1,66 @@
...
+
+import numpy as np
+
+from temporian.core.data import duration
+from temporian.implementation.numpy.data.io import event_set
+from temporian.test.utils import assertOperatorResult
+
+
+class FilterMovingCountTest(parameterized.TestCase):
+    @parameterized.parameters(
```
ianspektor, line 26: nice!

ianspektor (approved):
LGTM after removing methods from event_set_ops 💪🏼

PR #301, opened 2023-11-02
Support for timezone argument in calendar operators

DonBraulio:
Adds support to specify the timezone in calendar operators, using the timezone name (e.g: `America/Montevideo`) or an offset in hours (e.g: `-3`) in the argument `tz`.

To ease serialization, the core operator only receives the offset in hours.


temporian/core/operators/calendar/test/test_timezone_argument.py
```diff
@@ -0,0 +1,126 @@
...
+
+from datetime import datetime, timedelta, timezone
+from absl.testing import absltest
+from absl.testing.parameterized import TestCase
+
+from temporian.implementation.numpy.data.io import event_set
+from temporian.test.utils import assertOperatorResult, i32
+
+
+class CalendarTimezoneTest(TestCase):
```
ianspektor, line 23: I feel these tests might be easier to miss here than if they were in each operator's test file - is there a reason to have them here together?

temporian/core/operators/calendar/base.py
```diff
@@ -56,9 +66,32 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+                    type=pb.OperatorDef.Attribute.Type.FLOAT_64,
+                ),
+            ],
             outputs=[pb.OperatorDef.Output(key="output")],
         )
 
+    @staticmethod
+    def convert_to_utc_offset(timezone: Union[str, float, int]) -> float:
+        if isinstance(timezone, str):
+            tz_time = datetime.datetime.now(tz=pytz.timezone(timezone))
```
ianspektor, line 81: would add test for the case where timezone isn't one of the existing options (and check the error message is decent, or catch it and throw our own if not)

temporian/core/operators/calendar/base.py
```diff
@@ -56,9 +66,32 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+            offset_timedelta = datetime.datetime.utcoffset(tz_time)
+            return (
+                0.0
+                if offset_timedelta is None
+                else offset_timedelta.total_seconds() / 3600
+            )
+        return float(timezone)
```
ianspektor, line 88: add case for timezone not being one of the accepted types (since runtime type checking can be disabled by the user I don't think we want to rely on it heavily)

temporian/core/operators/calendar/base.py
```diff
@@ -56,9 +66,32 @@ def build_op_definition(cls) -> pb.OperatorDef:
...
+                pb.OperatorDef.Attribute(
+                    key="utc_offset",
+                    type=pb.OperatorDef.Attribute.Type.FLOAT_64,
+                ),
+            ],
             outputs=[pb.OperatorDef.Output(key="output")],
         )
 
+    @staticmethod
+    def convert_to_utc_offset(timezone: Union[str, float, int]) -> float:
```
ianspektor, line 79: would de-attach this method from this class since it doesn't win anything by being in it (could be here outside the class, or in dtype_normalization, or somewhere else)

temporian/core/operators/calendar/test/test_timezone_argument.py
DonBraulio: The reason is I don't want to duplicate the very specific setUp() code, and also it's easier to check that I'm testing that the argument is connected in all operators.
I agree it's easy to miss the file though, so I renamed it to `test_base_timezone` and now it appears on top together with `test_base.py` (which already existed and tests other generic stuff).

PR #303, opened 2023-11-02
Cleanup tests restructure

DonBraulio:
- Remove some folders that don't make sense after tests restucture completed.
- Update create_operator.py to the new tests format.


tools/create_operator.py
```diff
@@ -274,64 +274,48 @@ def __call__(
...
+        result = evset.{lower_op}()
+        expected = event_set(timestamps=[], same_sampling_as=evset)
 
-    def test_base(self):
+        assertOperatorResult(self, result, expected, check_sampling=True)
+
+    @parameterized.parameters(
+        {{"in_timestamps": [0.0, 1.0], "out_timestamps": [0.0, 1.0]}},
+        {{"in_timestmaps": [1.0, 2.0], "out_timestamps": [1.0, 2.0]}},
+    )
```
ianspektor, line 302: I think the most probable scenario is not using parameterized rather than using it? I'd leave it out of the template (and maybe add a commented line with the import or a sentence nudging the developer towards using it if needed?)

ianspektor (approved):
Thanks!

tools/create_operator.py
DonBraulio: I think it's a positive thing to induce ourselves to use parametrized tests, since you tend to add more test cases and edge cases. Also, the effort to remove this is negligible. Don't you agree?

PR #304, opened 2023-11-02
Fix tick_calendar wday

DonBraulio:
This PR makes coherent the meaning of `wday` argument in `tick_calendar()`, with the numbering in `calendar_day_of_week()`. The former was using `wday=0` for Sundays (C++ convention) instead of `wday=0` for Mondays (python's `datetime.weekday` convention).

(Apparently, when I implemented `tick_calendar` I just assumed that the C++ and python conventions were the same 🤷‍♂️)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


achoum (approved):
Nice

ianspektor (approved):

PR #307, opened 2023-11-13
Fixed getting started links

javiber:
A couple of links in the readme were broken. They were pointing to (what seems like) an old location of the "Getting started" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Check the [Getting Started tutorial](https://temporian.readthedocs.io/en/stable/
 
 ## Next steps
 
-New users should refer to the [Getting Started](https://temporian.readthedocs.io/en/stable/getting_started/) guide, which provides a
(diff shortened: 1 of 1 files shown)


ianspektor (changes requested):
Getting started's path was updated in https://github.com/google/temporian/pull/293, its broken in the README atm because we haven't released yet (so the changes aren't in /stable in the docs). We should either release more often (though the manual part makes that not too desirable), or maybe point the README's links to /latest?

DonBraulio:
Agree with @ianspektor, it's time to make a new release because the changes in the docs are significant.
I'd keep this PR open until the new version is released, to not forget about it 😄

About changing to `/latest`... I'm not sure, most people will just do `pip install temporian` so I think that pointing to stable is the right thing. Although it's a bit annoying that changing the docs in any way requires a new release for the users to actually see it, imo it's a good thing to release more often at this stage of the project.

javiber:
Agree with you guys, I think it makes sense to just close this one so I will do that.

PR #308, opened 2023-11-13
Fix doc headers

ianspektor:
before
image

now
image


ianspektor:
Ignore failed CI, the loan prediction notebook's server is down

DonBraulio (approved):
Much better!

PR #313, opened 2023-11-16
Added drop operator

javiber:
Behind the scenes, it just uses the select implementation.

I am not sure about the dependencies in the build files, it's working but it was the result of many trial-and-error


temporian/core/operators/drop.py
```diff
@@ -0,0 +1,54 @@
...
+from temporian.core.typing import EventSetOrNode
+from temporian.proto import core_pb2 as pb
+from temporian.utils.typecheck import typecheck
+from temporian.core.operators.select import select
+
+
+@typecheck
+@compile
+def drop(
+    input: EventSetOrNode, feature_names: str | List[str]
```
ianspektor, line 35: The `|` syntax for type union was only introduced in 3.10, and temporian supports 3.8 and up (which is why the [tests GH action is failing](https://github.com/google/temporian/actions/runs/6895848158/job/18760722564?pr=313) on py3.8) - we need to use `Union`

temporian/core/operators/drop.py
```diff
@@ -0,0 +1,54 @@
...
+    if isinstance(feature_names, str):
+        feature_names = [feature_names]
+
+    input_features = input.schema.feature_names()
+
+    if not all([fn in input_features for fn in feature_names]):
+        raise TypeError(
+            "Features"
+            f" {[fn for fn in feature_names if fn not in input_features]} are"
+            " not present in the input features"
```
ianspektor, line 48: nit: not present in the input.

temporian/core/operators/BUILD
```diff
@@ -449,3 +449,16 @@ py_library(
...
+    name = "drop",
+    srcs = ["drop.py"],
+    srcs_version = "PY3",
+    deps = [
+        ":base",
+        "//temporian/core:operator_lib",
+        "//temporian/core/data:node",
+        "//temporian/core/data:schema",
+        "//temporian/proto:core_py_proto",
+    ],
```
ianspektor, line 463: these two don't seem necessary: 
```
"//temporian/core/data:schema",
"//temporian/proto:core_py_proto",
```

and these seem missing:
```
"//temporian/core:compilation",
"//temporian/core:typing",
"//temporian/core/operators:select",
```

try running the `python tools/build_cleaner.py` (though its weird these tests passed with missing deps)

temporian/core/operators/drop.py
```diff
@@ -0,0 +1,54 @@
...
+from temporian.core.compilation import compile
+from temporian.core.data.node import (
+    EventSetNode,
+    create_node_new_features_new_sampling,
+)
+from temporian.core.operators.base import Operator
+from temporian.core.typing import EventSetOrNode
+from temporian.proto import core_pb2 as pb
+from temporian.utils.typecheck import typecheck
+from temporian.core.operators.select import select
```
ianspektor, line 29: several unused imports - please install pylint and address the unused import warnings (maybe add it to the workspace's recommended extensions too while you're at it)

temporian/core/operators/test/BUILD
```diff
@@ -379,3 +379,14 @@ py_test(
...
 )
+
+py_test(
+    name = "test_drop",
+    srcs = ["test_drop.py"],
+    srcs_version = "PY3",
+    deps = [
+        "//temporian/implementation/numpy/data:io",
+        "//temporian/test:utils",
+        "//temporian/core/operators:drop",
```
ianspektor, line 390: unneeded (you're calling drop from the EventSet, not importing it from its module)

ianspektor (approved):
Great work! LGTM after comments are addressed and tests fixed :)

PR #316, opened 2023-11-23
Added before and after operators

javiber:


temporian/core/event_set_ops.py
```diff
@@ -4436,3 +4437,115 @@ def assign(
         from temporian.core.operators.glue import assign
 
         return assign(self, **others)
+
+    def before(
+        self: EventSetOrNode,
+        timestamp: Union[int, float, datetime],
+    ) -> EventSetOrNode:
+        """Filter events [`EventSet`][temporian.EventSet] that happened before a
```
achoum, line 4445: Nit: Some of the functions use the 3rd person e.g. "Filters events ...".

temporian/core/event_set_ops.py
```diff
@@ -4436,3 +4437,115 @@ def assign(
...
+        particual timestamp.
+
+        The timestamp can be a datetime if the EventSet's timestamps are unix
+        timestamps.
+
+        The comparison is strict, meaning that the obtained timestamps would be
+        less than (`<`) the provided timestamp.
```
achoum, line 4452: Would it make sense to say something like?:
This operation is equivalent to "input.filter(input.timestamps() < timestamp)"

temporian/core/event_set_ops.py
achoum, line 4445: I am not a native speaker, but I think "filter events before A" might be understood as removing the ones before A (i.e. the opposite of this operator). I wonder if this could be made less ambiguous.

temporian/core/event_set_ops.py
```diff
@@ -4436,3 +4437,115 @@ def assign(
...
+            events:
+                 (1 events):
+                    timestamps: ['2022-01-01T00:00:00']
+                    'f1': [1]
+            ...
+
+            ```
+
+        Args:
+            timestamp: EventSet with a single boolean feature.
```
achoum, line 4488: What about something like: Strict timestamp upper boundary (or something like that).

temporian/core/operators/filter.py
```diff
@@ -93,3 +95,37 @@ def filter(
...
+
+@compile
+def before(
+    input: EventSetOrNode, timestamp: Union[float, datetime]
+) -> EventSetOrNode:
+    assert isinstance(input, EventSetNode)
+
+    if isinstance(timestamp, datetime):
+        if not input.schema.is_unix_timestamp:
+            raise ValueError(
```
achoum, line 108: Do we want the opposite to be allowed? e.g. timestamp is a float and the input is a unix timestamp.

achoum (approved):
Awesome

temporian/core/event_set_ops.py
javiber: I can add it, do you think I should also add a link to filter?

temporian/core/event_set_ops.py
javiber: I would understand "filter out..." as the opposite, since I'm not a native speaker either, run it through chatGPT the question:
```
Given the following documentation of a function called f:

"Filters events that happened before a particular timestamp."

What do you understand that the function should do?
```
And it understood the correct thing:
```
def f(events, timestamp):
    # Filter events that happened before the specified timestamp
    filtered_events = [event for event in events if event['timestamp'] < timestamp]
    return filtered_events
```

PR #317, opened 2023-11-23
Export timestamps as datetimes if they are unix timestamps

javiber:


ianspektor (approved):
Awesome 💯 Glad to support this!

PR #324, opened 2023-12-06
Fixed TZ with DST on calendar operations

javiber:


temporian/core/operators/calendar/base.py
```diff
@@ -87,17 +82,17 @@ def build_op_definition(cls) -> pb.OperatorDef:
             inputs=[pb.OperatorDef.Input(key="sampling")],
             attributes=[
                 pb.OperatorDef.Attribute(
-                    key="utc_offset",
-                    type=pb.OperatorDef.Attribute.Type.FLOAT_64,
+                    key="tz",
+                    type=pb.OperatorDef.Attribute.Type.ANY,
```
achoum, line 86: It would be interesting to see if a tzinfo is considered a string. If not, to simplify the serialization, would it be export / imported from a string?

achoum (approved):
Awesome

temporian/core/operators/calendar/base.py
javiber: I'm not sure how it can be serialized but it should be possible to use a string. In the new optimized version, this value is no longer a tzinfo, instead it can be a float, int or string

achoum: int, float and string can be well serialized.

javiber:
MAde obsolete by https://github.com/google/temporian/pull/325

PR #325, opened 2023-12-07
Optimize calendar ops

javiber:


temporian/implementation/numpy_cc/operators/calendar/common.h
```diff
@@ -0,0 +1,51 @@
+#include "absl/time/time.h"
+
+static absl::TimeZone handle_tz(const pybind11::object tz) {
+  if (pybind11::isinstance<pybind11::int_>(tz)) {
+    // handle tz as an int
+    int int_tz = tz.cast<int>();
```
achoum, line 6: const int (here and in other places)

temporian/implementation/numpy_cc/operators/calendar/common.h
```diff
@@ -0,0 +1,51 @@
...
+    int int_tz = static_cast<int>(std::round(float_tz));
+    // tz is expresed in hours, needs to be converted to seconds
+    return absl::FixedTimeZone(int_tz);
+
+  } else if (pybind11::isinstance<pybind11::str>(tz)) {
+    // handle tz as an string
+    std::string str_tz = tz.cast<std::string>();
+    absl::TimeZone parsed_tz;
+    if (!absl::LoadTimeZone(str_tz, &parsed_tz)) {
+      throw std::invalid_argument("Invalid timezone '" + str_tz +
```
achoum, line 22: Currently, we don't use exceptions. Would there be another way to pass the error?

temporian/implementation/numpy_cc/operators/calendar/day_of_month.cc
```diff
@@ -0,0 +1,49 @@
...
+#include "absl/time/time.h"
+#include "temporian/implementation/numpy_cc/operators/calendar/common.h"
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+py::array_t<int> calendar_day_of_month(const py::array_t<double> &timestamps,
+                                       const py::object tz) {
+  auto parsed_tz = handle_tz(tz);
```
achoum, line 20: ... = prase_tz(tz); ?

temporian/implementation/numpy_cc/operators/calendar/day_of_year.cc
```diff
@@ -0,0 +1,49 @@
...
+#include <vector>
+
+#include "absl/time/time.h"
+#include "temporian/implementation/numpy_cc/operators/calendar/common.h"
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+py::array_t<int> calendar_day_of_year(const py::array_t<double> &timestamps,
```
achoum, line 18: A large part of this code is similar in the different operators. Would it be possible to factorize it?

temporian/implementation/numpy_cc/operators/calendar/isoweek.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <vector>
+
+#include "absl/time/time.h"
+#include "temporian/implementation/numpy_cc/operators/calendar/common.h"
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+// directly translated to cpp from panda's ccalendar.pyx implementation
```
achoum, line 18: Can you add a link?

temporian/implementation/numpy_cc/operators/calendar/common.h
```diff
@@ -0,0 +1,51 @@
+#include "absl/time/time.h"
+
+static absl::TimeZone handle_tz(const pybind11::object tz) {
```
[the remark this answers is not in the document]
javiber: you mean `tz`?
I'm pretty rusty in c++ so I'd trust your judgement in this, it would prevent the value from being copied so it sounds like a good idea

temporian/implementation/numpy_cc/operators/calendar/common.h
javiber: I like having exceptions in these cases and I think we have some in python, is there any reason to avoid the exceptions in c++?

I see a couple of options:

1. change it to a warning and ignore the timezone (use UTC)
2. return some kind of error code and handle it on python

I think the exception is better than the warning but that's a design decision and if you think the warning fits better with our current design I'm not against it.

number 2 I don't see it as an improvement, it would just move the problem from c++ to python and complicate the return types

javiber:
@achoum @ianspektor I just pushed a major refactor to reuse more code. It made each operator so simple that I ended up consolidating everything under a single calendar.cc.

I'm not sure why the tests are failing on 3.8 and 3.11, I haven't been able to reproduce that locally

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+#include <pybind11/pybind11.h>
+
+#include "absl/time/time.h"
+
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+static absl::TimeZone parse_tz(const py::object tz) {
```
achoum, line 11: Can you comment on the pro/con of using a reference?
const py::object tz
vs
const py::object& tz

temporian/implementation/numpy_cc/operators/calendar/common.h
achoum: Our c++ code is not written with exceptions in mind. Before starting to use them, we would have to figure out the implications for all the current and future code. In addition, Google engineers don't use exceptions (https://google.github.io/styleguide/cppguide.html#Exceptions), and since this code is used in Google, it would be a pain to enable them.

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+    const int int_tz = static_cast<int>(std::round(float_tz));
+    // tz is expresed in hours, needs to be converted to seconds
+    return absl::FixedTimeZone(int_tz);
+
+  } else if (py::isinstance<py::str>(tz)) {
+    // handle tz as an string
+    const std::string str_tz = tz.cast<std::string>();
+    absl::TimeZone parsed_tz;
+    if (!absl::LoadTimeZone(str_tz, &parsed_tz)) {
+      throw std::invalid_argument("Invalid timezone '" + str_tz +
```
achoum, line 30: Blocker

temporian/implementation/numpy_cc/operators/calendar.cc
achoum, line 11: Why "static" ? (here and in other places)

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+    }
+    return parsed_tz;
+
+  } else {
+    throw std::invalid_argument("Unsupported argument type for \"tz\" argument");
+  }
+}
+
+// TODO: support week starting on Sunday
+static int map_week_day(const absl::Weekday &wd) {
```
achoum, line 42: Nit: Could this function be a constexpr ?

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+  // Allocate output array
+  auto output = py::array_t<int>(n_events);
+
+  // Access raw input / output data
+  auto v_output = output.mutable_unchecked<1>();
+  auto v_timestamps = timestamps.unchecked<1>();
+
+  for (Idx i = 0; i < n_events; i++) {
+    // Create absolute time in UTC
+    auto nanos = static_cast<int64_t>(v_timestamps[i] * 1e9);
```
achoum, line 83: const

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+    // Convert to civil time and get the isoweek
+    auto local_time = absl::ToCivilSecond(utc_time, parsed_tz);
+    v_output[i] = calendar_op.Get(local_time);
+  }
+
+  return output;
+}
+
+// Calendar operations interface
+template <typename OUTPUT> struct CALENDAR_OP {
```
achoum, line 95: Since all the implementations return an "int", and none of them have any internal state, what about having the template parameter be simply a non templated function (instead of a templated class)?

Bonus: I don't think the speed gain of using a template is significant here (actually, it is likely to output the same code with the proper compilation flags). Maybe it would make the code simpler to use an argument such a: std::function<int(absl::Time)> function).

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,258 @@
...
+  }
+};
+
+py::array_t<int> calendar_day_of_week(const py::array_t<double> &timestamps,
+                               const py::object tz) {
+  return apply_calendar_op<int, DayOfWeek>(timestamps, tz);
+}
+
+
+// isoweek
```
achoum, line 196: rm "isoweek" or line return

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: That was a workaround when I had `calendar.h` I will remove it now

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: I read a bit about `constexpr` and if I understood correctly I don't think it'll have much of an impact because the weekday is not known at compile time.
With that said, it doesn't look like it would have any downside so if you think we should add it I'll do so

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: I changed this implementation and now it will receive a reference because of a change in design however this discussion still applies to `map_week_day`.

I left the weekday as a const reference to indicate that I don't want the function to modify the weekday but I don't want to make create unnecessary copies of the value. Is this reasoning correct? is there another way to indicate this? does it even make sense to consider this given that the weekday is just an enum so copying the value should not be a problem?

Sorry to respond your question with 3 new questions :)

temporian/implementation/numpy_cc/operators/calendar/common.h
javiber: I changed the implementation to return an error message

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: Changed it to avoid raising exceptions. the new implementation returns an error message

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: done

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: You were right! I made the change and now there is much less boilerplate in the implementation

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: done

temporian/core/operators/calendar/test/test_calendar_day_of_year.py
```diff
@@ -42,6 +41,26 @@ def test_basic(self):
         result = evset.calendar_day_of_year()
         assertOperatorResult(self, result, expected)
 
+    @parameterized.parameters({"tz": 1}, {"tz": 1.0}, {"tz": "Europe/Brussels"})
```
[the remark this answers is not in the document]
javiber: missed this comment before, will add this cases

temporian/implementation/numpy_cc/operators/calendar/common.h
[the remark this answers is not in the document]
achoum: As a rule of thumb, as soon as an object is larger than a two cpu registers, it is better to use a ref or a pointer.

temporian/implementation/numpy_cc/operators/calendar/common.h
```diff
@@ -0,0 +1,51 @@
...
+    absl::TimeZone parsed_tz;
+    if (!absl::LoadTimeZone(str_tz, &parsed_tz)) {
+      throw std::invalid_argument("Invalid timezone '" + str_tz +
+                                  "'. Only names defined in the IANA timezone" +
+                                  "database are valid");
+    }
+    return parsed_tz;
+
+  } else {
+    throw std::invalid_argument("Unsupported argument type");
```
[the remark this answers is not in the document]
achoum: Please comment on action.

temporian/implementation/numpy_cc/operators/calendar/common.h
achoum: Please comment on action.

temporian/implementation/numpy_cc/operators/calendar/day_of_month.cc
achoum: Please comment on action.

temporian/implementation/numpy_cc/operators/calendar/day_of_year.cc
achoum: Please comment on action.

temporian/implementation/numpy_cc/operators/calendar/isoweek.cc
achoum: Please comment on action.

temporian/implementation/numpy_cc/operators/calendar/common.h
achoum: What about creating a typdef std::optional<std::string> Error ?
This way, it is easier to understand that if empty, there is not error.
As discussed, the alternative is to use absl::Status: https://abseil.io/docs/cpp/guides/status

temporian/implementation/numpy_cc/operators/calendar.cc
achoum: >> I changed this implementation and now it will receive a reference because of a change in design however this discussion still applies to map_week_day.

I still see no reference. Are you sure you pushed your change?

What about the following conversion:
- const& => Input argument that cannot be modified.
- const => A small input argument that cannot be modified.
- * (pointer) => An argument that is expected to be modified.

temporian/implementation/numpy_cc/operators/calendar/common.h
javiber: My first option was using the Status but I ran into a couple of problems:
1. Our abseil version is quite old, had to update it
2. Had to add [pybind11_abseil](https://github.com/pybind/pybind11_abseil) in order to convert the `absl::Status `to a python object
3. `pybind11_abseil` needed a newer version of `rules_python` which triggers many errors in other dependencies which should also be updated (protobuf for instance)

Given this I went with your other option and use `std::optional<sdt::string>` will push the changes soon

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: Done

temporian/core/operators/calendar/test/test_calendar_day_of_year.py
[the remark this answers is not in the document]
javiber: Added these tests and one for checking leap years.

I don't love the fact that the tests are spread between all operators, for instance, I tested the leap year for `day_of_the_month` because it's more convenient.
Given that the calendar operators share most code if the leap year is correct for one of the operators then is correct for all of them, so I think this is OK.

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,243 @@
...
+
+#include <iostream>
+
+#include "absl/time/time.h"
+#include "temporian/implementation/numpy_cc/operators/common.h"
+
+namespace {
+namespace py = pybind11;
+
+std::optional<std::string> parse_tz(const py::object &tz, absl::TimeZone &parsed_tz) {
```
achoum, line 14: Please create a typedef

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,243 @@
...
+  } else {
+    return "Unsupported argument type for \"tz\" argument. Expecting int, "
+           "float, "
+           "or str";
+  }
+
+  return {};
+}
+
+int is_leapyear(int64_t year) {
```
achoum, line 46: Use temporian/implementation/numpy_cc/operators/tick_calendar_utils.cc

Also, add new functions in tick_calendar_utils.cc and add unit tests

temporian/implementation/numpy_cc/operators/BUILD
```diff
@@ -78,6 +89,7 @@ pybind_extension(
         ":tick_calendar",
         ":until_next",
         ":window",
+        "//temporian/implementation/numpy_cc/operators:calendar",
```
achoum, line 92: No need for full path.

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,243 @@
...
+
+namespace {
+namespace py = pybind11;
+
+std::optional<std::string> parse_tz(const py::object &tz, absl::TimeZone &parsed_tz) {
+  if (py::isinstance<py::int_>(tz)) {
+    // handle tz as an int
+    const int int_tz = tz.cast<int>();
+    // tz is expresed in hours, needs to be converted to seconds
+    parsed_tz = absl::FixedTimeZone(int_tz * 60 * 60);
```
achoum, line 19: What about creating a variable SECONDS_PER_HOURS ?

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,243 @@
...
+  const auto error = parse_tz(tz, parsed_tz);
+  if (error.has_value()) {
+    return error;
+  }
+
+  // Input size
+  const Idx n_events = timestamps.shape(0);
+
+  // Allocate output array
+  // auto output = py::array_t<int>(n_events);
```
achoum, line 87: Clean?

temporian/implementation/numpy_cc/operators/calendar.cc
```diff
@@ -0,0 +1,243 @@
...
+    // Convert to civil time and call calendar_op
+    const auto local_time = absl::ToCivilSecond(utc_time, parsed_tz);
+    v_output[i] = calendar_op(local_time);
+  }
+
+  return {};
+}
+
+// year calendar op
+int year(absl::CivilSecond value) { return value.year(); }
```
achoum, line 107: Nit: define "year" in "calendar_year" as a lambda

achoum (approved):

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: done! new type is called `ErrorString`

temporian/implementation/numpy_cc/operators/calendar.cc
javiber: Makes sense, will switch to `IsLeapYear` and move `map_week_day` and `parse_tz` to the utils file. I have a couple of notes though:

1. I think it'd be a good idea to rename `tick_calendar_utils` to `calendar_utils` since now they are not just for tick operations.
2. I notice that you are using PascalCase for the functions but I was using snake_case, is there a preference in cpp for it? I tend to default to snake_case automatically because of Python

javiber:
@achoum I ran into some issues when moving my utility functions to `tick_calentar_utils`. I merge this PR anyway as it was long overdue, we can discuss about the utility functions and debug when you return

PR #327, opened 2023-12-11
Added timestamp param to to_pandas

javiber:


temporian/io/pandas.py
```diff
@@ -132,6 +133,7 @@ def to_pandas(
             np.string_ or np.bytes) to Pandas strings (equivalent to np.str_).
         timestamp_to_datetime: IF true, cast Temporian timestamps to datetime64
             when is_unix_timestamp is set to True.
+        timestamps: IF true, the timestamps are included as a column.
```
ianspektor, line 136: super nit, change lowercase f in IF (same in the previous param)

ianspektor (approved):
💯

PR #328, opened 2023-12-13
Added to and from parquet opperations

javiber:


achoum (approved):
Nice.
Could you add those to the API and the documentation also?

Issue #332, opened 2024-01-09
Building wheel for temporian (pyproject.toml) ... error


kopes18:
```
Building wheels for collected packages: temporian
  Building wheel for temporian (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for temporian (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [57 lines of output]
      fatal: not a git repository (or any of the parent directories): .git
      A setup.py file already exists. Using it.
      running build
      running bazel_build
      Traceback (most recent call last):
        File "C:\Users\xxzz\AppData\Local\Temp\pip-install-mfsrqwoh\temporian_7928305007e042308ed74670d834944b\setup.py", line 80, in <module>
          setup(**setup_kwargs)
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\__init__.py", line 103, in setup
          return distutils.core.setup(**attrs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\_distutils\core.py", line 185, in setup
          return run_commands(dist)
                 ^^^^^^^^^^^^^^^^^^
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\_distutils\core.py", line 201, in run_commands
          dist.run_commands()
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\_distutils\dist.py", line 969, in run_commands
          self.run_command(cmd)
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\dist.py", line 963, in run_command
          super().run_command(command)
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\_distutils\dist.py", line 988, in run_command
          cmd_obj.run()
        File "C:\Users\xxzz\AppData\Local\Temp\pip-build-env-odk8ya52\overlay\Lib\site-packages\setuptools\_distutils\command\build.py", line 131, in run
          self.run_command(cmd_name)
[...]

PR #334, opened 2024-01-09
Add support for pip build on Windows

achoum:


ianspektor:
Addresses https://github.com/google/temporian/issues/332

ianspektor (approved):
I have no windows PC to test this on - maybe @rstz could review?

We should also add it to `.github/worfklows/publish.yaml` so that its built, tested, and published to PyPI on merges to main - I could take care of that if needed, + it would also ensure it does build on a clean GH-hosted Windows runner

achoum:
Thanks.

>> We should also add it to .github/worfklows/publish.yaml
Good point. Let me try this in a following PR, and see if I need you help :).

PR #337, opened 2024-01-12
A few fixes

achoum:


ianspektor (approved):
would be good to add a test to ensure the tz conversion is working as expected (in a follow up PR). thanks!

PR #340, opened 2024-01-17
Windows release 0.7.0

achoum:


config/setup.py
```diff
@@ -1,7 +1,7 @@
 import setuptools
 from setuptools.dist import Distribution
 
-_VERSION = "0.1.6"
+_VERSION = "0.7.0"
```
ianspektor, line 4: add modifying the version here in tools/RELEASE.md, step 2?

tools/build.bat
```diff
@@ -12,13 +12,18 @@
 cls
 setlocal
 
-set TEMPORIAN_VERSION=0.1.6
+set TEMPORIAN_VERSION=0.7.0
```
ianspektor, line 15: same for here

ianspektor (approved):

config/setup.py
achoum: *Done

tools/build.bat
achoum: *Done

PR #360, opened 2024-01-30
Add Mac M1 publish runner

rstz:
IIUC, we only need to specify the correct runners to get the MacOS runner. Reference: https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

@ianspektor Can you help me test this?

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -70,10 +70,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      # - name: Set up Python ${{ matrix.python-version }}
-      #   uses: actions/setup-python@v4
-      #   with:
-      #     python-version: ${{ matrix.python-version }}
+      # actions/setup-python doesn't support py < 3.10 on macos arm
+      # https://github.com/actions/setup-python/issues/696#issuecomment-1637587760
 
       - name: Install Python
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yaml
```diff
@@ -54,11 +54,37 @@ jobs:
       - name: Save package for following jobs
         uses: actions/upload-artifact@v3
         with:
-          name: dist-macos-${{ matrix.python-version }}
+          name: dist-macos-intel-${{ matrix.python-version }}
+          path: dist
+
+  build-macos-arm:
+    runs-on: macos-latest-xlarge
```
ianspektor, line 61: is there a reason for using xlarge for arm and large for intel?

ianspektor, line 61: according to this https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories `latest` points to `macos-12`, but the one using M1 is `macos-14`?

rstz: As I understand Githubs documentation: large --> Intel. xlarge --> Arm https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#:~:text=The%20following%20machines%20sizes%20are%20available%20for%20macOS%20larger%20runners.

rstz: I think we can use the large machines, where IIUC correctly large and xlarge are how they are distinguished https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#:~:text=The%20following%20machines%20sizes%20are%20available%20for%20macOS%20larger%20runners.

ianspektor: Hm, agree, in the link I sent below I understood otherwise

ianspektor:
@rstz ready on my side, review and merge away!

rstz:
LGTM! I think you have to formally approve this

ianspektor (approved):

Issue #354, opened 2024-01-18
Scalene supports Python 3.11 and Python 3.12


emeryberger:
[Scalene author here.] Scalene has supported Python 3.11 and 3.12 for some time, so the below line can now be uncommented, though I'd use a more recent version (the current version today is `1.5.33.1`).

https://github.com/google/temporian/blob/9977a866a6090ade06a7575ff13fc14e8bb9b00d/pyproject.toml#L52

ianspektor:
That's great news, thanks for the heads up!

We disabled it some months ago so I'll need to check if what we wrote still works - else I'll work on updating it, we got some useful insights from it.

emeryberger:
Would love to hear more!

ianspektor:
Actually - if someone wants to work on this, by all means do, should be an easy first issue :)

The command to run time profiling with scalene is `bazel run -c opt --config=linux //benchmark:profile_time` (documented [here](https://github.com/google/temporian/blob/main/CONTRIBUTING.md#time-and-memory-profiling)), and it runs [this script](https://github.com/google/temporian/blob/main/benchmark/profile_time.sh).

tanaysd:
@ianspektor interested but want to know approximate timeline to push the fix for this?

ianspektor:
Thank you! Let me know if you need any help or additional info.

tanaysd:
@ianspektor ack!

tanaysd:
gonna need additional time to get this PR in because of a packed week on my side, planning to get it done by end of this week (1.29 - 2.3)

tanaysd:
@ianspektor when you have moment, could you take a look at my privileges? I create a feature branch and was going to push it to create a PR but I received the following message 
<img width="847" alt="image" src="https://github.com/google/temporian/assets/3865925/baa7167c-a0c1-46bc-833e-d6834b592434">

javiber:
Hi @tanaysd!

You can't push to the repo directly but you can create a fork and create a PR from it. More info [here](https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-a-project)

Let me know if you need anything else. I'll be helping Ian with the support of some of these tickets, feel free to tag me.
You can also find me on our [discord](https://discord.gg/QdCkyFF6)

tanaysd:
@javiber that makes sense, I think we should update the [`cont](https://github.com/google/temporian/blob/main/CONTRIBUTING.md) to reflect this pattern.

[...]

PR #365, opened 2024-02-09
activate scalene and bump to latest version

tanaysd:
@ianspektor @javiber ptal, all bazel tests passed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ Benchmarking and profiling of pre-configured scripts is available as follow:
 
 ```shell
 bazel run -c opt --config=linux //benchmark:profile_time -- [name]
-bazel run -c opt --config=linux //benchmark:profile_memory -- [name] [-p]
+bazel run -c opt --config=linux //benchmark:profile_memory -- [name]
 ```
 
 where `[name]` is the name of one of the python scripts in
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ Benchmarking and profiling of pre-configured scripts is available as follow:
 
 ```shell
 bazel run -c opt --config=linux //benchmark:profile_time -- [name]
-bazel run -c opt --config=linux //benchmark:profile_memory -- [name]
+bazel run -c opt --config=linux //benchmark:profile_memory -- [name] [-p]
 ```
 
 where `[name]` is the name of one of the python scripts in
--- a/benchmark/scripts/add_index.py
+++ b/benchmark/scripts/add_index.py
@@ -22,7 +22,6 @@ import temporian as tp
 from temporian.implementation.numpy.data.event_set import EventSet
 
 
-@profile
 def run(input_node, input_data, output_node):
     tp.run(output_node, input={input_node: input_data}, check_execution=False)
 
--- a/benchmark/scripts/basic.py
+++ b/benchmark/scripts/basic.py
@@ -36,7 +36,6 @@ STORE, PRODUCT, TIMESTAMP, SALES, COSTS = (
 )
 
 
-@profile
 def main():
     print(f"Running basic benchmark with N={N}...")
 
@@ -89,7 +88,7 @@ def main():
     )
 
     # Print output's first row, useful to check reproducibility
-    print(res.first_index_data())
+    print(res)
 
 
 if __name__ == "__main__":
--- a/benchmark/scripts/sma.py
+++ b/benchmark/scripts/sma.py
@@ -36,7 +36,6 @@ STORE, PRODUCT, TIMESTAMP, SALES, COSTS = (
 )
 
 
-@profile
 def main():
     print(f"Running lag benchmark with N={N}...")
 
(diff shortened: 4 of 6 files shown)


javiber:
Hi @tanaysd thanks for open this PR so soon!

The tests are complaining that `poetry.lock` is out of sync with the `pyproject.toml` can you include the lock file?

ianspektor:
Thanks for taking on this @tanaysd! Note also that the benchmarks aren't ran as part of the test suite but rather need to be ran manually locally to ensure they work, see [this comment] in the original issue.

tanaysd:
@ianspektor are you referring to [this comment](https://github.com/google/temporian/issues/354#issuecomment-1899159416)?

@javiber will update `.lock` file as well

tanaysd:
@javiber PTAL, update `lock` file, added python restrictions for Scalene as indicated by Scalene's author

@ianspektor executred `bazel` tests for three scripts in the script folder, what should I be looking for?

ianspektor:
> @ianspektor are you referring to https://github.com/google/temporian/issues/354#issuecomment-1899159416?

yep, sorry :)

> @ianspektor executed bazel tests for three scripts in the script folder, what should I be looking for?

if `bazel run -c opt --config=linux //benchmark:profile_time -- basic` (or `-- sma` or `-- add_index`) runs and shows scalene's output you're good to go! let us know so we can test it locally before merging. thanks!

tanaysd:
> > @ianspektor are you referring to [#354 (comment)](https://github.com/google/temporian/issues/354#issuecomment-1899159416)?
> 
> yep, sorry :)
> 
> > @ianspektor executed bazel tests for three scripts in the script folder, what should I be looking for?
> 
> if `bazel run -c opt --config=linux //benchmark:profile_time -- basic` (or `-- sma` or `-- add_index`) runs and shows scalene's output you're good to go! let us know so we can test it locally before merging. thanks!

Yes, for all three tests, it opened a local webpage displaying `scalene` metrics -- you are good to test locally.

ianspektor:
Confirm that scalene installs and runs 👍🏼 Thanks!

Some stuff seems to need to be fixed/updated for Scalene's new version (maybe @emeryberger can further comment on these):

- Running `bazel run -c opt --config=linux benchmark:profile_time -- sma` now shows no line output, only the total result for the `main()` function in `benchmark/scripts/sma.py`. I *think* it showed line-by-line results for the function annotated with @profile before. I'd remove the `@profile` annotation now, which shows the full output (we were using it before because some imports were heavy enough to contaminate the output, which is still the case, but that's better than not seeing the line-by-line).
image

- I'd remove the @profile annotation from the other scripts in `benchmark/scripts` too.
- The `-p` option doesn't seem to exist anymore, it needs to be removed from the `### Benchmarking and profiling` section in `CONTRIBUTING.md`
- The `benchmark/scripts/basic.py` script fails when ran, it's using `EventSet.first_index_data()` that was deprecated. Printing the whole EventSet as a fix is fine.

tanaysd:
> Confirm that scalene installs and runs 👍🏼 Thanks!
> 
> Some stuff seems to need to be fixed/updated for Scalene's new version (maybe @emeryberger can further comment on these):
> 
> * Running `bazel run -c opt --config=linux benchmark:profile_time -- sma` now shows no line output, only the total result for the `main()` function in `benchmark/scripts/sma.py`. I _think_ it showed line-by-line results for the function annotated with @Profile before. I'd remove the `@profile` annotation now, which shows the full output (we were using it before because some imports were heavy enough to contaminate the output, which is still the case, but that's better than not seeing the line-by-line).
> 
> image
> * I'd remove the @Profile annotation from the other scripts in `benchmark/scripts` too. -- `DONE`
> * The `-p` option doesn't seem to exist anymore, it needs to be removed from the `### Benchmarking and profiling` section in `CONTRIBUTING.md` -- `DONE`
> * The `benchmark/scripts/basic.py` script fails when ran, it's using `EventSet.first_index_data()` that was deprecated. Printing the whole EventSet as a fix is fine. -- `DONE`

@ianspektor @javiber PTAL

CONTRIBUTING.md
```diff
@@ -95,7 +95,7 @@ Benchmarking and profiling of pre-configured scripts is available as follow:
 
 ```shell
 bazel run -c opt --config=linux //benchmark:profile_time -- [name]
-bazel run -c opt --config=linux //benchmark:profile_memory -- [name] [-p]
+bazel run -c opt --config=linux //benchmark:profile_memory -- [name]
```
ianspektor, line 98: @tanaysd - big mistake on my side, `profile_memory` uses `mprof`, not scalene (see `benchmark/profile_memory.sh`) 😬 please revert this change. LGTM from my side after that! 👍🏼

tanaysd: done.

ianspektor (approved):

ianspektor:
LGTM - thanks @tanaysd!

emeryberger:
The only reason I can think that there wouldn't be line level information is that no individual line accounted for more than 1% of execution time (or memory, if you are profiling that). That logic is not affected by the presence or absence of the `@profile` decorator. It would make sense to look at the `profile.json` file to verify that individual lines are being tracked.

PR #374, opened 2024-02-20
Add option to force garbage collection in long pipeline runs

achoum:


temporian/implementation/numpy/evaluation.py
```diff
@@ -43,9 +46,13 @@ def run_schedule(
         check_execution: If `True`, data of the intermediate results of the
             operators is checked against its expected structure and raises if
             it differs.
+        force_garbage_collector_interval: If set, triggers the garbage
+            collection every "force_garbage_collector_interval" seconds.
     """
     data = {**inputs}
 
+    gc_being_time = time.time()
```
ianspektor, line 54: typo in var name begin

ianspektor (approved):

temporian/implementation/numpy/evaluation.py
achoum: Good catch.
*Fiex

PR #376, opened 2024-02-21
Tick and tick calendar include left and right

javiber:
- Added option to include the right to both operators and made it `True` by default
- Added option to include the left to both operators and made it `False` by default
- Both options have no effect if the start/end is equal to the first/last tick
- Fixed a bug in tick_calendar where it would miss the first tick under certain conditions
- ~Changed the behavior of the tick operator when input has 1 item, now it always returns an empty event set (used to return one element if `align=False` and empty otherwise)~

[Calendar ops](https://github.com/google/temporian/pull/325) needs to be merged first

### DIFF
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -4136,6 +4136,8 @@ class EventSetOperations:
         mday: Optional[Union[int, Literal["*"]]] = None,
         month: Optional[Union[int, Literal["*"]]] = None,
         wday: Optional[Union[int, Literal["*"]]] = None,
+        include_right: bool = True,
+        include_left: bool = False,
     ) -> EventSetOrNode:
         """Generates events periodically at fixed times or dates e.g. each month.
 
@@ -4252,6 +4254,8 @@ class EventSetOperations:
             wday: '*' (any day), None (auto) or number in range `[0-6]`
                     (Sun-Sat) to tick at particular day of week. Can only be
                     specified if `day_of_month` is `None`.
+            include_right: If True, a tick after the last timestamp is included.
+            include_left: If True, a tick before the first timestamp is included.
 
         Returns:
             A feature-less EventSet with timestamps at specified interval.
@@ -4266,6 +4270,8 @@ class EventSetOperations:
             mday=mday,
             month=month,
             wday=wday,
+            include_right=include_right,
+            include_left=include_left,
         )
 
     def timestamps(self: EventSetOrNode) -> EventSetOrNode:
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -4090,8 +4090,8 @@ class EventSetOperations:
         self: EventSetOrNode,
         interval: Duration,
         align: bool = True,
-        include_right: bool = True,
-        include_left: bool = False,
+        after_last: bool = True,
+        before_first: bool = False,
     ) -> EventSetOrNode:
         """Generates timestamps at regular intervals in the range of a guide
         [`EventSet`][temporian.EventSet].
@@ -4102,7 +4102,7 @@ class EventSetOperations:
             >>> b = a.tick(interval=tp.duration.seconds(3), align=True)
             >>> b
             indexes: ...
-                    timestamps: [ 6. 9. 12. 15.]
+                    timestamps: [ 6. 9. 12. 15. 18.]
             ...
 
             ```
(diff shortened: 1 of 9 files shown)


temporian/core/event_set_ops.py
```diff
@@ -3916,7 +3916,11 @@ def since_last(
         return since_last(self, steps=steps, sampling=sampling)
 
     def tick(
-        self: EventSetOrNode, interval: Duration, align: bool = True
+        self: EventSetOrNode,
+        interval: Duration,
+        align: bool = True,
+        include_right: bool = True,
```
achoum, line 3922: Having "right" = "future", and "left" = "paste" is something we have in the current code.
Can the parameters be about past/future (or something equivalent)?

temporian/core/event_set_ops.py
```diff
@@ -3949,13 +3953,21 @@ def tick(
                 (similar to [`EventSet.begin()`][temporian.EventSet.begin]).
```
achoum, line 3953: Can you add an example with those new parameters?

temporian/core/event_set_ops.py
```diff
@@ -4081,6 +4095,8 @@ def tick_calendar(
             wday: '*' (any day), None (auto) or number in range `[0-6]`
                     (Sun-Sat) to tick at particular day of week. Can only be
                     specified if `day_of_month` is `None`.
+            include_right: If True, a tick after the last timestamp is included.
```
achoum, line 4098: Same remark as before.

temporian/core/operators/test/test_tick.py
```diff
@@ -72,6 +72,94 @@ def test_align_single(self):
             self, result, expected_output, check_sampling=False
         )
 
+    def test_not_inclusive_right(self):
```
achoum, line 75: What about a parametrized test?

temporian/core/operators/test/test_tick_calendar.py
```diff
@@ -25,60 +25,114 @@ class TickCalendarOperatorTest(parameterized.TestCase):
...
+            "right": None,
         },
     )
-    def test_base(self, span, expected, calendar_args):
-        evset = event_set(timestamps=span)
-        expected_evset = event_set(timestamps=expected)
-        result = evset.tick_calendar(**calendar_args)
+    def test_base(self, span, left, ticks, right, calendar_args):
+        for include_right in [None, True, False]:
+            for include_left in [None, True, False]:
```
achoum, line 104: What about a parameterized test?

temporian/implementation/numpy/operators/tick.py
```diff
@@ -41,14 +41,15 @@ def __call__(self, input: EventSet) -> Dict[str, EventSet]:
 
         # fill output EventSet data
         for index_key, index_data in input.data.items():
-            if len(index_data.timestamps) == 0:
+            if len(index_data.timestamps) <= 1:
```
achoum, line 44: For brainstorming: Should there be cases where at least one event is returned?

temporian/core/operators/test/test_tick.py
```diff
@@ -72,6 +72,94 @@ def test_align_single(self):
...
+        )
+
+        # Aligned, no include left, start doesn't match first tick
+        result = evset.tick(4.0, align=True, include_left=False)
+        expected_output = event_set([4, 8, 12])
+        assertOperatorResult(
+            self, result, expected_output, check_sampling=False
+        )
+
+    def test_inclusive_left(self):
```
achoum, line 147: Can you also try with negative timestamps?

temporian/implementation/numpy_cc/operators/tick_calendar.cc
```diff
@@ -28,36 +32,61 @@ py::array_t<double> tick_calendar(
...
   std::vector<double> ticks;
 
-  // Date range
-  const auto start_t = static_cast<std::time_t>(std::floor(start_timestamp));
-  const auto end_t = static_cast<std::time_t>(std::floor(end_timestamp));
+  // Direction in wich to search
+  int step = forward ? 1 : -1;
+
+  // End time
+  const bool end_time_set = end_timestamp >= 0;
```
achoum, line 39: It look like a negative end_timestamp is a magic value. What about using std::optional instead?
Also, should this code work with negative timestamps?

temporian/core/operators/test/test_tick.py
javiber: that's a good catch! I totally forgot about negative timestamps

temporian/implementation/numpy_cc/operators/tick_calendar.cc
javiber: yes! that's much better

temporian/core/operators/test/test_tick.py
javiber: That makes more sense, will change it

temporian/core/operators/test/test_tick_calendar.py
javiber: this one is a bit tricky, I have the test cases defined and want to try all the combinations of left/right inclusion. I could define the test cases outside and programmatically expand them to all the combinations before calling `@parameterized.parameters`

temporian/core/event_set_ops.py
javiber: Renamed them to `past_inclusive` and `future_inclusive`

temporian/implementation/numpy/operators/tick.py
javiber: Changed this, the behavior now is similar to what we have currently on main.

Check the new test: https://github.com/google/temporian/pull/376/files#diff-22438ae7b5327b043eb72f32e896dc62eb5c44d523d3535a0f3b65c1339894a1R57

temporian/core/event_set_ops.py
ianspektor: I'd change to `include_past` and `include_future`, though left/right sound more natural to me. @achoum could you clarify what you mean by it being something we have in the current code/why that's a problem?

ianspektor: resolved: `before_first` and `after_last`

PR #377, opened 2024-02-22
feat(operators): Add trigonometric function operators to EventSet

### LINKED ISSUE #348: New operators: trigonometric functions
New trigonometric function operators: `EventSet.sin()`, `.cos()`, .`tan()`, `.arcsin()`, .`arccos()`, `.arctan()`.

Implementation should be factored with base classes so that each operation only needs to define the numpy method it needs to call.

Can possibly be implemented by extending `temporian/core/operators/unary.py`.

See numpy's definitions of these functions for reference. [str_len](https://numpy.org/doc/stable/reference/generated/numpy.char.str_len.html) for example.

See https://github.com/google/temporian/blob/main/CONTRIBUTING.md#developing-a-new-operator for guidance. 

Questions or requests for additional guidance from possible contributors more than welcome!

akshatvishu:
Closes #348 
- Implemented `sin`, `cos`, `tan`, `arcsin`, `arccos`, and `arctan` operators in `EventSet`.

- Only test for `tan` is failing and due a unusual `AssertionError` .

```py

AssertionError: index[64 chars]   timestamps: [1.]
        'f': [1.7321]
memory usage: 0.5 kB
 != index[64 chars]   timestamps: [1.]
        'f': [1.7321]
memory usage: 0.5 kB
 : 
==========
RESULT:
==========
indexes: []
features: [('f', float64)]
events:
     (1 events):
        timestamps: [1.]
        'f': [1.7321]
memory usage: 0.5 kB

==========
EXPECTED:
==========
indexes: []
features: [('f', float64)]
events:
     (1 events):
        timestamps: [1.]
        'f': [1.7321]
memory usage: 0.5 kB
```


javiber:
We already discussed this on Discord but I'll add it here for reference. 
The expected value for `tan` is incorrect, the error message is rounding the `RESULT` and `EXPECTED` so they look the same but they are not.

Besides the test, the code looks good to me!

akshatvishu:
> We already discussed this on Discord but I'll add it here for reference. The expected value for `tan` is incorrect, the error message is rounding the `RESULT` and `EXPECTED` so they look the same but they are not.


Thank you @javiber for the correction! I have pushed the changes as you suggested!

javiber (approved):

Issue #352, opened 2024-01-17
Feature: tp.to_numpy()


ianspektor:
Export an EventSet to numpy arrays with `tp.to_numpy(evset)`.

See `tp.to_pandas()` for reference.

The EventSet is flattened (its index is dropped and kept as features). Returns a dictionary mapping feature names to values, + an extra "timestamps" key with the timestamps.

Musa-Sina-Ertugrul:
Hi, I want to give a try this. Can you assign me?

![penguins-of-madagascar-waving-hello-pb0tkzpn2x5dz3ho](https://github.com/google/temporian/assets/102359522/0b14695d-efeb-4755-91e0-ab52eecfeed2)

ianspektor:
Hey @Musa-Sina-Ertugrul! Have someone on Discord that wanted to take on it but hasn't started, just pinged them to see if they are OK with freeing it up, I'll let you know when they get back to me. Thanks!

Musa-Sina-Ertugrul:
Not problem If he/she wants I can work on other good first issue labels.

ianspektor:
Hey @Musa-Sina-Ertugrul! No answer from them on Discord so it's all yours if you want it.

Musa-Sina-Ertugrul:
Thanks 😄 Yes, I want

Musa-Sina-Ertugrul:
I could not managed preparing environment sorry :(  Can you give this task to someone

nagavenkateshgavini:
Hello @ianspektor 
I would like to contribute to this feature, please let me know If I can start working on this. Thank you.

ianspektor:
@Musa-Sina-Ertugrul sorry to hear that - could you open an issue with what went wrong? Both so that we can help you set it up and improve the documentation for everyone else :)

PR #378, opened 2024-02-23
352: Enhancement: Support tp.to_numpy() for temporian

nagavenkateshgavini:
PR to https://github.com/google/temporian/issues/352
- [x] add support for tp.to_numpy()
- [x] Provide `timestamps: bool = True` argument
- [x] Provide `timestamp_to_datetime: bool = True` argument
- [x] Add tests in `temporian/io/test/numpy_test.py`
- [x] test_correct
- [x] test_no_index
- [x] test_no_timestamps
- [x] test_timestamp_to_datetime_param
- [x] test_empty_event_set
- [x] Add doc string


temporian/io/test/numpy_test.py
```diff
@@ -0,0 +1,81 @@
...
+            },
+            indexes=["my_index"],
+        )
+
+        result = to_numpy(evset)
+
+        expected_timestamps = np.array(['2023-11-08T17:14:38', '2023-11-29T21:44:46'],
+                                       dtype='datetime64[s]')
+        expected_feature_1 = np.array([0.5, 0.6])
+        expected_my_index = np.array(["red", "blue"])
```
javiber, line 25: Nit: Could we define this as a dictionary?

something like:
``` python
expected = {
    "timestamps": np.array(['2023-11-08T17:14:38', '2023-11-29T21:44:46'],
                                dtype='datetime64[s]'),
    "feature_1": np.array([0.5, 0.6]),
    "my_index": np.array(["red", "blue"]),
}
```

It would make the code a bit more compact and allow the asserts below to be on a for-loop

temporian/io/test/numpy_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Convert byte strings in x to Unicode strings for a fair comparison
+        result['my_index'] = np.array([item.decode('utf-8') for item in result['my_index']])
+
+        np.testing.assert_array_equal(np.sort(result["feature_1"]), np.sort(expected_feature_1))
+        np.testing.assert_array_equal(np.sort(result["timestamp"]), np.sort(expected_timestamps))
+        np.testing.assert_array_equal(np.sort(result["my_index"]), np.sort(expected_my_index))
```
javiber, line 32: I don't like the `np.sort` here, if the order is not correct we would like this to fail but sorting both arrays before the assert would hide that problem

temporian/io/test/numpy_test.py
```diff
@@ -0,0 +1,81 @@
...
+        expected_timestamps = np.array(['2023-11-08T17:14:38', '2023-11-29T21:44:46'], dtype='datetime64[s]')
+        expected_feature_1 = np.array([0.5, 0.6])
+        expected_my_index = np.array(["red", "blue"])
+
+        # Convert byte strings in x to Unicode strings for a fair comparison
+        result['my_index'] = np.array([item.decode('utf-8') for item in result['my_index']])
+
+        np.testing.assert_array_equal(np.sort(result["feature_1"]), np.sort(expected_feature_1))
+        np.testing.assert_array_equal(np.sort(result["my_index"]), np.sort(expected_my_index))
+
```
javiber, line 78: A few more test cases to add:

- `timestamp=False`
- `timestamp_to_datetime=False`
- empty event_set

javiber (changes requested):
Thanks for the contribution! Code looks good!

I left some ideas to improve the tests

temporian/io/test/numpy_test.py
nagavenkateshgavini: @javiber 
When I create `event_set`, the original input is modified. That is why I am sorting yo make fair comparison. See the below example.


```python
evset = event_set(
            timestamps=['2023-11-08T17:14:38', '2023-11-29T21:44:46'],
            features={
                "feature_1": [0.5, 0.6],
                "my_index": ["red", "blue"],
            },
            indexes=["my_index"],
        )
result = to_numpy(evset)

print(result)


{'my_index': array([b'blue', b'red'], dtype='|S4'), 'feature_1': array([0.6, 0.5]), 
'timestamp': array(['2023-11-29T21:44:46', '2023-11-08T17:14:38'],
 dtype='datetime64[s]')}
```

Please suggest, If I can go ahead like this or make changes.

temporian/io/test/numpy_test.py
nagavenkateshgavini: added more tests

temporian/io/test/numpy_test.py
nagavenkateshgavini: modified to dict and added a loop to check each key

nagavenkateshgavini:
@javiber 
I've addressed all your feedback, with the only remaining question being whether to sort the array before comparing it or not.

temporian/io/test/numpy_test.py
javiber: I see, the same thing happens to `to_pandas`. Let's leave it as it is but add a comment saying that we had to sort because to_numpy doesn't follow a particular order (I think it's sorting by index) and I'll open another issue to discuss if we should have a specific order for `to_pandas` and `to_numpy`

javiber (changes requested):
We will tackle the sorting in another issue, I left you a small comment but there is something more important that I missed in the first review, we need to format the test file using `black`.

1. run `black temporian/io/test/numpy_test.py`. Black should already be installed, you might need to activate the virtualenv (`poetry shell`) or pre-pend `poetry run ` to the command
2. add, commit and push the changes to the file

temporian/io/test/numpy_test.py
```diff
@@ -0,0 +1,98 @@
...
+        expected = {
+            "timestamp": np.array(['2023-11-08T17:14:38', '2023-11-29T21:44:46'],
+                                   dtype='datetime64[s]'),
+            "feature_1": np.array([0.5, 0.6]),
+            "my_index": np.array(["red", "blue"]),
+        }
+
+        # Convert byte strings in result to Unicode strings for a fair comparison
+        result['my_index'] = np.array([item.decode('utf-8') for item in result['my_index']])
```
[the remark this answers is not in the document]
nagavenkateshgavini: Thanks, this is done!

javiber:
@nagavenkateshgavini There is still a formatting error, you can run `black --check .` locally to get more info about the problem

nagavenkateshgavini:
@javiber 
just committed by reformatting

```
$ black --check .                        
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install black[jupyter]``
All done! ✨ 🍰 ✨
307 files would be left unchanged.
```

nagavenkateshgavini:
@javiber 
Could you please provide your approval for the pending workflow checks?

ianspektor:
Hey @nagavenkateshgavini! Docs build was fixed in https://github.com/google/temporian/pull/386, please merge main to your PR to fix the `jupyter: command not found` error. Approving the workflows now. Thanks!

temporian/io/numpy.py
```diff
@@ -0,0 +1,95 @@
...
+        >>> evset = tp.event_set(
+        ...     timestamps=['2023-11-08T17:14:38', '2023-11-29T21:44:46'],
+        ...     features={
+        ...         "store": ['STORE_1', 'STORE_2'],
+        ...         "revenue": [1571, 6101]
+        ...     },
+        ...     indexes=["store"],
+        ... )
+
+        # Indices are not set as dataframe's indices. Timestamps are exported as
```
javiber, line 45: the first sentence is not relevant for numpy, can you remove it?

nagavenkateshgavini:
@ianspektor 
I merged my branch with the `main`, and the docs test was successful.
Could you please approve the workflows? Hopefully, this will be the last workflow test for this PR.

cc: @javiber

javiber (approved):

ianspektor:
Congrats and thank you for your contribution @nagavenkateshgavini! 🥳

nagavenkateshgavini:
Thank you, @ianspektor, for the amazing insights; this is truly fascinating. And a huge thank you to @javiber for your constant support and guidance throughout this process.

PR #383, opened 2024-02-28
380: Initial implementation of moving_product operation

### LINKED ISSUE #380: New operator: moving_product
New `EventSet.moving_product()` operator.

See `EventSet.moving_sum()` for reference. Note that this issue will require C++ code.

See https://github.com/google/temporian/blob/main/CONTRIBUTING.md#developing-a-new-operator for guidance.

Questions or requests for additional guidance from possible contributors more than welcome!

akshatvishu:
Closes #380 

Implemented initial setup for `moving_product` in Temporian. Added method stubs and began C++ integration.

ToDo:
- [x] Add test 
- [x] Add docs 
- [x] Run Code Formatter (Black)
- [  ]~~Maybe add it under `temporian/temporian/beam`~~


temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -516,6 +516,53 @@ struct MovingMaxAccumulator : MovingExtremumAccumulator<INPUT, OUTPUT> {
...
+
+    OUTPUT Result() override {
+        double product = 1.0;
+        bool has_zero = false;
+
+        // Calculate the product of all values inside the window
+        for (int idx = start_idx; idx <= end_idx; ++idx) {
+            const INPUT value = Accumulator<INPUT, OUTPUT>::values[idx];
+            if (value == 0) {
+                has_zero = true;
```
javiber, line 547: I think you can simply `return 0` right here and remove the flag `has_zero`

javiber:
Hi @akshatvishu I just tested this locally and found a couple of small things:

- [ ] The example in `event_set_ops.py` line 3200 fails because the result is expressed in exponential notation due to the size of the results. Adjust the notation or use smaller values in the input.
- [ ] Add the `cumprod` operation as a moving_prod with an infinite window length ([cumsum](https://github.com/google/temporian/blob/19a908d9f90353845711b7d802ecf45d5fd95c97/temporian/core/operators/window/moving_sum.py#L66) for reference). As you can guess from the implementation, this is going to be pretty slow so add some warnings on the documentation and we will add a todo to optimize it in the future.
- [ ] In the do you mention that "If the window does not contain any non-missing, non-zero values, the output for that window is missing (NaN)." I'd like a test checking that this is the case.

Other than these small things the PR is looking great, I think we can merge this after these points are addressed.

temporian/core/event_set_ops.py
```diff
@@ -2133,6 +2133,85 @@ def cast(
...
+        self: EventSetOrNode,
+        sampling: Optional[EventSetOrNode] = None,
+    ) -> EventSetOrNode:
+        """Computes the cumulative product of values over each feature in an
+        [`EventSet`][temporian.EventSet].
+
+        This operation only supports floating-point features.
+
+        Foreach timestamp, calculate the product of the feature from the beginning.
+        Shorthand for `moving_product(event, window_length=np.inf)`.
```
javiber, line 2146: no need to explain the underlying implementation

temporian/core/event_set_ops.py
```diff
@@ -2133,6 +2133,85 @@ def cast(
...
+
+        Missing (NaN) values are not accounted for. The output will be NaN until
+        the input contains at least one numeric value.
+
+        If `sampling` is specified or `window_length` is an EventSet, the moving
```
javiber, line 2151: `window_length` doesn't exist for this operator

temporian/core/event_set_ops.py
```diff
@@ -2133,6 +2133,85 @@ def cast(
...
+        window is sampled at each timestamp in them, else it is sampled on the
+        input's.
+
+        Warning: The `cumprod` function leverages an infinite window length for
```
javiber, line 2155: Again, no need to explain the underlying implementation but **do** keep the warning

javiber (approved):

ianspektor:
Congrats and thanks on the huge contribution @akshatvishu! 💪🏼

akshatvishu:
> Congrats and thanks on the huge contribution @akshatvishu! 💪🏼

Big thanks to @javiber  for his  blazing fast reviews and suggestions. Couldn't have done this without his guidance 🙏

PR #385, opened 2024-02-29
Fix coverage workflow

jsoref:
This fixes:
1. https://github.com/google/temporian/actions/runs/8083954295
    [comment](https://github.com/google/temporian/actions/runs/8083954295/job/22088416732#step:2:71)
Unhandled error: HttpError: Not Found
2. Anyone using this workflow in a repository with paranoid default permissions (which my organizations do) see https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-the-default-github_token-permissions -- note: you should apply the restriction to your repositories too (probably at the org level, but at least at the repository level) -- but, obviously that's a bigger change.
3. Pull Requests will now get comments (this fixes https://github.com/google/temporian/pull/381)

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -55,10 +55,12 @@ jobs:
             const pr = ${{fromJson(needs.coverage.outputs.cov-pr).totals.percent_covered}};
             const main = ${{fromJson(needs.coverage.outputs.cov-main).totals.percent_covered}};
             const diff = pr - main;
+            const report = `### Coverage report
+            **Main**: ${main.toFixed(2)}% | **PR**: ${pr.toFixed(2)}% | **Diff: ${diff.toFixed(2)} ${diff >= 0 ? '✅' : '⚠️'}**`;
+            await core.summary.addRaw(report).write()
             github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: `### Coverage report
-              **Main**: ${main.toFixed(2)}% | **PR**: ${pr.toFixed(2)}% | **Diff: ${diff.toFixed(2)} ${diff >= 0 ? '✅' : '⚠️'}**`
+              body: report
             })
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -2,6 +2,10 @@ name: coverage
 
 on: [push]
 
+# limit default permissions to just read-only checkouts
+permissions:
+  contents: read
+
 jobs:
   coverage:
     runs-on: ubuntu-latest
@@ -48,6 +52,10 @@ jobs:
   comment:
     runs-on: ubuntu-latest
     needs: [coverage]
+    permissions:
+      # needed to add coverage comment to the pull request
+      pull-requests: write
+
     steps:
       - uses: actions/github-script@v6
         with:
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -66,9 +66,13 @@ jobs:
             const report = `### Coverage report
             **Main**: ${main.toFixed(2)}% | **PR**: ${pr.toFixed(2)}% | **Diff: ${diff.toFixed(2)} ${diff >= 0 ? '✅' : '⚠️'}**`;
             await core.summary.addRaw(report).write()
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: report
-            })
+            try {
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: report
+              })
+            } catch (e) {
+              console.log("Could not post comment to pull request", e)
+            }
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,6 +1,8 @@
 name: coverage
 
-on: [push]
+on:
+  # this allows for `contents: write` and `pull-requests: write` from forks
+  pull_request_target:
 
 # limit default permissions to just read-only checkouts
 permissions:
@@ -19,7 +21,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          ref: ${{ matrix.branch == 'main' && 'main' || '' }}
+          # for safety, `pull_request_target` changes the default checkout to be the target branch, so we have to request the merge (we're limited to `permissions: {contents: read}`, so this is ok)
+          ref: ${{ matrix.branch == 'main' && 'main' || (github.event.pull_request && format('refs/pull/{0}/merge', github.event.pull_request.number)) || '' }}
 
       - name: Bazel cache
         id: bazel-cache


.github/workflows/coverage.yaml
```diff
@@ -50,6 +54,9 @@ jobs:
     needs: [coverage]
     steps:
       - uses: actions/github-script@v6
+        env:
+          issue_number: ${{ github.event.issue.number }}
+        if: ${{ env.issue_number }}
```
jsoref, line 59: This prevents the github-script from running on pushes that don't have pull requests and thus don't have a place to leave a comment.

It's possible to write this logic in javascript (I did this for testing), but I don't see the point as it's much more expensive to have the poor runner actually use the runtime than it is to get the host to evaluate this eagerly. -- It'd make more sense to do it in the code if you could properly produce a step summary from the script (see below)

.github/workflows/coverage.yaml
```diff
@@ -62,3 +69,18 @@ jobs:
...
+          main_fixed=$(echo "scale=2; $main/1" | bc)
+          pr_fixed=$(echo "scale=2; $pr/1" | bc)
+          diff_fixed=$(echo "scale=2; $diff/1" | bc)
+          if [ $(echo "scale=0; $diff/1" | bc) -gt 0 ]; then
+            icon='✅'
+          else
+            icon='⚠️'
+          fi
+          echo "### Coverage report
+            **Main**: ${main_fixed}% | **PR**: ${pr_fixed}% | **Diff: ${diff_fixed} ${icon}**" >> "$GITHUB_STEP_SUMMARY"
```
jsoref, line 86: It's theoretically possible to write this in github-script, but I tried a couple of times and I gave up. I don't speak github-script (I do sometimes speak javascript/typescript, and I contribute to their various repositories, but I generally end up using shell/perl to talk to github apis instead).

For the curious, you can review [my attempts at making this workflow behave](https://github.com/jsoref/temporian/actions/workflows/coverage.yaml?query=branch%3Afix-coverage-workflow).

At posting, this is the [working one].(https://github.com/jsoref/temporian/actions/runs/8097376074#summary-22128559880)

Think of this more as a proposal than a "definitely merge this approach". If you have someone who's familiar w/ github-script, then they should be able to make a simple fix to https://github.com/jsoref/temporian/commit/64a0753ec2a65f963f1c528f6813b7bb3d09d322 (which didn't work -- see the lack of a summary in https://github.com/jsoref/temporian/actions/runs/8097265388).

ianspektor:
Does this fix https://github.com/google/temporian/pull/381? If so that's extremely helpful :)

jsoref:
It would, yes :)

jsoref:
Or, rather. kinda.

That thing is broken in fun ways.

It depends what you're trying to really achieve. 

Personally I'd get rid of writing a comment on the PR and only do step summary.

My workflow/action (check-spelling) supports both, but the effort required to maintain commenting is scary to most people.

To actually get comments on PRs from forks requires changing from `pull_request` to `pull_request_target` and changing the checkout logic to carefully checkout the attacker's branch.

Actually, let me revise this so it's less scary.

.github/workflows/coverage.yaml
```diff
@@ -2,6 +2,9 @@ name: coverage
 
 on: [push]
 
+permissions:
+  contents: read
```
jsoref, line 6: ## GITHUB_TOKEN Permissions
### Original permissions in non paranoid repositories
```yaml
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

### Original permissions in paranoid repositories
```yaml
  Contents: read
  Metadata: read
  Packages: read
```

### Reduced Permissions
```yaml
  Contents: read
  Metadata: read
```

ianspektor:
> Personally I'd get rid of writing a comment on the PR and only do step summary.

Didn't know about step summaries ([ref](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) for anyone else reading this). 

Is there a way to make them visible in the PR itself instead of in the Action summary? Having to go check the summary pretty much makes it useless, what we want is an obviously visible warning if your PR is decreasing coverage.

.github/workflows/coverage.yaml
```diff
@@ -48,8 +51,14 @@ jobs:
   comment:
     runs-on: ubuntu-latest
     needs: [coverage]
+    permissions:
+      pull-requests: write
```
jsoref, line 55: ## GITHUB_TOKEN Permissions
### Original permissions in non paranoid repositories (when not dealing with forks)
```yaml
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

### Original permissions in non paranoid repositories (when dealing with forks)
```yaml
  Actions: read
  Checks: read
  Contents: read
  Deployments: read
  Discussions: read
  Issues: read
  Metadata: read
  Packages: read
  Pages: read
  PullRequests: read
  RepositoryProjects: read
  SecurityEvents: read
  Statuses: read
```

### Original permissions in paranoid repositories
```yaml
  Contents: read
  Metadata: read
  Packages: read
```

### Reduced Permissions (when not dealing with forks)
```yaml
  Pull-Requests: write
  Metadata: read
```

### Reduced Permissions (when dealing with forks)
```yaml
  Pull-Requests: read
  Metadata: read
```

Note that this isn't good enough to result in things working... to make it work requires that additional bit of stuff...

jsoref:
I'm hoping github will eventually improve access to step summaries -- at the very least, i'm hoping github will replace `details` links with links to step summaries. -- I'm also hoping that larger entities (hi!) will lean on github to make that change.

Sadly, no, step summaries don't have great visibility, but the permissions required for commenting are generally sufficiently scary that most of the time projects I work with decide to use step summaries instead of allowing my workflow to comment.

Granting `pull-requests: write` allows a job to try to do all of the things listed here:
https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests (some of them are guarded by requiring secondary permissions, but there's still enough damage that it really isn't advisable). (Minus the caveat of whether github actually gives out the `write` -- as you've seen, it doesn't for PRs from a fork, but it would for `pull_request_target` -- see below.)

An alternative is having a bot or some purpose built (and carefully audited/maintained) code that has a token (with `pull-requests: write`) and have it be responsible for consuming/posting.

Anyway, if you trust workflows and that people can review/audit them, then, you can spin a wheel.

The other thing you can do is include some `try` blocks to make sure that when you don't have `write` permissions, the step doesn't actually fail.

Note that changing from `push` / `pull_request` to `pull_request_target` also changes how testing / reviewing workflows works. If a change is made to a workflow that uses `pull_request_target`, that workflow change isn't tested as part of the pull request, the new changes would only take effect after merging (otherwise an attacker could put whatever they wanted in their PR, which would obviously be awful).

Basically this stuff is a bit of a headache -- sorry :)

jsoref:
@ianspektor: anyway, if you tell me what you'd like, I'm happy to adjust the workflow to do it (within reason)

ianspektor:
Hey 👋🏼 Don't have time to dive into this atm but will try to get to it early next week - thanks so much for the very detailed explanations!

ianspektor:
Alright, here's what I propose:
- We need the comment. We can additionally do the step summary too, if the comment on PRs from a fork isn't possible. But having the step summary only beats the purpose of this action, which is alerting us when a PR decreases coverage.
- Changing to `pull_request_target` sounds like it makes sense. I don't mind the change in behavior (workflow changes not affecting their own PR). I don't completely understand what you mean by "changing the checkout logic to carefully checkout the attacker's branch" here - please do ellaborate if necessary.
- If changing to `pull_request_target` isn't the way to go, the `try` blocks as a simple solution for not failing on forks sound OK too.

Thoughts?

.github/workflows/coverage.yaml
```diff
@@ -1,6 +1,11 @@
 name: coverage
 
-on: [push]
+on:
+  push:
+  pull_request_target:
```
jsoref, line 5: This is the beginning of restoring `pull_request` support

.github/workflows/coverage.yaml
```diff
@@ -15,7 +20,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          ref: ${{ matrix.branch == 'main' && 'main' || '' }}
+          ref: ${{ matrix.branch == 'main' && 'main' || (github.event.pull_request && format('refs/pull/{0}/merge', github.event.pull_request.number)) || '' }}
```
jsoref, line 23: There's a choice to be made here.

There are 3 cases:
- this is a reference build so we check out `main`
- this a pull request and we need to check out _something_ -- here we're checking out the merge head between `main` and the pull request that's what `/merge` does (this is what actions/checkout would do for the `pull_request` target -- but we have to do it manually because `pull_request_target` is treated differently)  -- there's an alternative of checking out the pull request `/head`, but I think `/merge` is the right choice
- this is a push and we're using the default behavior of letting actions/checkout get the commit for us

jsoref:
> "changing the checkout logic to carefully checkout the attacker's branch" here - please do ellaborate if necessary.

So, this requires essays, and there are some, e.g.: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.

TL;DR:

If someone is triggering an event in their own repository, they already had the ability to write to the repository, and thus it isn't interesting if what they push can make additional changes to their repository. -- This is conceptually, e.g. _me_ pushing to my own repository and a workflow in my repository being given `permissions: { contents: "write" }`.

If someone is triggering an event in someone else's repository, then it'd be bad if they could cause a change to the repository to be made without consent of the owners of that repository. This is the case you have here. I'm making a significant change to a workflow. -- If my change were directly run _before_ you merged it, that'd mean my PR could trigger additional commits to be added and that's really bad. GitHub has a number of protections for this: first by default workflows running from PRs across forks are limited to `permissions: *: read` -- even if a workflow says it wants `permissions: _something_: write`. This is the standard `pull_request` event -- and actions/checkout will, by default, check out the merge of the pull request head and the pull request target branch.

Sometimes, you need to be able to allow a PR to do things with `write` permissions (the comment here is the classic example), so github has a `pull_request_target` event -- but it runs in the context of the target. So, to prevent repository owners and workflow authors from making their repositories too easily take over, the default for actions/checkout is to check out the pull request target -- where the code is code that the target repository has already accepted.

In the proposed workflow, for a future PR w/ this enabled, there are three jobs and they'll look like this:


## [coverage (pr)](https://github.com/check-spelling-sandbox/temporian/actions/runs/8203240330/job/22435645309?pr=3#step:1:1)
```yml
GITHUB_TOKEN Permissions
  Contents: read
  Metadata: read
```
### [Checkout](https://github.com/check-spelling-sandbox/temporian/actions/runs/8203240330/job/22435645309?pr=3#step:2:1)
```yml
Run actions/checkout@v3
  with:
    ref: refs/pull/3/merge
```

-- this part is where things could be dangerous (as the code being run is from a fork and is thus untrustworthy) -- if not for the careful permissions established by the workflow changes here --

## [coverage (main)](https://github.com/check-spelling-sandbox/temporian/actions/runs/8203240330/job/22435645498?pr=3#step:1:1)
```yml
GITHUB_TOKEN Permissions
  Contents: read
  Metadata: read
```
### [Checkout](https://github.com/check-spelling-sandbox/temporian/actions/runs/8203240330/job/22435645498?pr=3#step:2:1)
```yml
Run actions/checkout@v3
  with:
    ref: main
```

## [comment](https://github.com/check-spelling-sandbox/temporian/actions/runs/8203240330/job/22435890473?pr=3#step:1:1)
```yml
GITHUB_TOKEN Permissions
  Metadata: read
  PullRequests: write
```
### no checkout

If there was a checkout, it'd default to using the pull request target's branch. Here the code (and in fact, all of the workflow steps themselves) came from the pull request target.

---

In general, changes to workflows that have pull_request_target events are not seen until after they're merged, so testing them involves setting up a repository where the commits are already present (I'm not certain about how `pull request approvals` interacts with this restriction, but my guess is that it doesn't change anything).

.github/workflows/coverage.yaml
```diff
@@ -1,6 +1,11 @@
 name: coverage
 
-on: [push]
+on:
+  push:
```
ianspektor, line 4: do we need `push` in addition to `pull_request_target`?

.github/workflows/coverage.yaml
ianspektor: merge sounds right 👍🏼

jsoref:
> We could also just change it to `pull_request` temporarily while testing, if it's us modifying it.

Yeah. I'd personally just create a PR into a branch that has the changes I'm testing (as I did here to show things). But, that's probably because I spend 99% of my time dealing with my workflows that use `pull_request_target`.

.github/workflows/coverage.yaml
```diff
@@ -48,17 +53,28 @@ jobs:
...
+              try {
+                github.rest.issues.createComment({
+                  issue_number: context.issue.number,
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  body: report
+                })
+              } catch (e) {
+                console.log("This happens when a PR from a fork triggers a workflow", e)
+              }
```
ianspektor, line 78: is this outdated?

.github/workflows/coverage.yaml
jsoref: This is a choice.

If you want someone who is developing to get feedback on this before they create a PR, then you could leave `push`. If you only want them to get feedback when they create a PR, then you can remove it.

.github/workflows/coverage.yaml
jsoref: Probably. But I'd sooner leave it as belt-and-suspenders -- maybe changing the message to say it isn't particularly expected.

.github/workflows/coverage.yaml
```diff
@@ -48,17 +53,28 @@ jobs:
...
+            }
+            await core.summary.addRaw(report).write()
```
jsoref, line 80: I should move this to _before_ the `createComment` call -- that'd make the system less brittle.

ianspektor:
One more thing: could you add some very concrete (~1-line) explanations in the tricky/not-so-standard parts?
- Permissions
- why pull_request_target and not pull_request
- checkout ref logic
Add links if needed

.github/workflows/coverage.yaml
ianspektor: Let's remove it then, we can check coverage locally if needed. Would removing it also allow removing the `if (context.issue.number) {` in the comment step?

.github/workflows/coverage.yaml
ianspektor: 👍🏼 to changing the message

.github/workflows/coverage.yaml
ianspektor: Almost commented exactly this, go ahead

jsoref:
Sadly the various doc/blog links are kinda lousy, here's the "basic" text:
> [Restricting permissions for tokens](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#restricting-permissions-for-tokens)
To help mitigate the risk of an exposed token, consider restricting the assigned permissions. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)."

It isn't wrong, but it isn't remotely helpful.

[Modifying the permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)

> You can modify the permissions for the `GITHUB_TOKEN` in individual workflow files. If the default permissions for the `GITHUB_TOKEN` are restrictive, you may have to elevate the permissions to allow some actions and commands to run successfully. If the default permissions are permissive, you can edit the workflow file to remove some permissions from the `GITHUB_TOKEN`. As a good security practice, you should grant the `GITHUB_TOKEN` the least required access.

Again, the information is _kinda_ here, but it's more or less buried in the last sentence of the paragraph.

There's a three part series: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ but none of them suggest chaining jobs to handle untrusted actions even though that's really the way to go 🙍‍♂️.

jsoref:
Fwiw, once this is merged, someone should go in and bump `actions/checkout` to `@v4` and `actions/github-script` to `@v7` to get rid of the node16 warnings.

I'm leaving that for a separate PR as I don't really want to extend this PR -- it's already sufficiently complicated.

ianspektor (approved):

ianspektor:
Thanks a ton for both the contribution and the extensive info @jsoref 💪🏼 lmk if you'd like to take #388 too

PR #397, opened 2024-03-18
A few small changes.

achoum:
- Allow for "begin" and "after" operators to work on mix of unix timestamps and non unix timestamps.
- Add tp.duration.to_string to the API (alias for tp.core.data.duration_utils.duration_abbreviation).
- Run black .


docs/public_api_test.py
```diff
@@ -70,7 +70,7 @@ class PublicAPITest(absltest.TestCase):
     def test_public_symbols(self):
         """Asserts that the symbols exposed under tp.<> are exactly the
         ones we expect."""
-        symbols = {s for s in dir(tp) if not s.startswith("__")}
+        symbols = {s for s in dir(tp) if not s.startswith("__") and s != "api"}
```
ianspektor, line 73: `del api` at the end of the root `__init__` is the correct fix for this I believe

ianspektor (approved):

docs/public_api_test.py
achoum: Good point.
*Removed condition and added "del api".

achoum:
Thanks

javiber:
@achoum I think the docs are failing because you're missing the `__init__.py` inside `temporian/api`

temporian/core/operators/filter.py
```diff
@@ -105,20 +105,7 @@ def before(
     assert isinstance(input, EventSetNode)
 
     if isinstance(timestamp, datetime):
-        if not input.schema.is_unix_timestamp:
```
javiber, line 108: Should we have a warning here?

I'm in favor of removing the exception to make the operation more flexible but I'm wondering if we should warn the user about this.

achoum:
> @achoum I think the docs are failing because you're missing the `__init__.py` inside `temporian/api`

Awesome catch. Thanks.

temporian/core/operators/filter.py
achoum: Good point.

*Print warning instead of error.
*Add `is_unix_timestamp` setter.

PR #402, opened 2024-03-20
Fixed build-linux and build-macos-arm

javiber:
There seems to be a problem with actions/checkout@v4 on some containers
There is a problem with `std::optional.value()` function on some macos compilers

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -17,7 +17,7 @@ jobs:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v3  # v4 doesn't work on the container
 
       - name: Build package
         run: |
@@ -110,7 +110,7 @@ jobs:
       SYSTEM_VERSION_COMPAT: 0
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -169,7 +169,7 @@ jobs:
       POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v3
```
ianspektor, line 172: we should be able to keep v4 in the ones that don't use a custom image right?

ianspektor (approved):

jsoref:
* GitHub's first party actions have switched to node20 (GitHub is dropping support for node16 because it's long past EOL -- so any normal runner that uses them gets a whole bunch of warnings that node16 support will go away soon).
* node20 requires modern glibc symbols which is why you'll get an error like:
https://github.com/google/temporian/actions/runs/8360467099/job/22886201078#step:3:17
> /__e/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /__e/node20/bin/node)

Do you need the container for a whole bunch of things, or is it just to get bazel? Because if it's just for bazel, perhaps it would make sense to use:
https://github.com/marketplace/actions/setup-bazel-environment

Note: I'm not remotely attached to any of this, but I am happy to help projects with migrations if they're open to it.

javiber:
Hi @jsoref, I think it's just for bazel but I'm not sure, my plan is to return to this once I better understand why we are using that container and see if we can get an updated image or switch to another way to install it like the action you shared.

For now, I just need to get the release ready

PR #407, opened 2024-03-25
Upgrade to tensorflow 2.13

### LINKED ISSUE #387: Not able to test temporian on Mac OS M2 chip
When I tried to install temporian packages with poetry, there was a problem with TensorFlow installation.

Error:

```shell
  RuntimeError

  Unable to find installation candidates for tensorflow (2.12.0)

  at /opt/homebrew/Cellar/poetry/1.7.1/libexec/lib/python3.12/site-packages/poetry/installation/chooser.py:73 in choose_for
       69│ 
       70│             links.append(link)
       71│ 
       72│         if not links:
    →  73│             raise RuntimeError(f"Unable to find installation candidates for {package}")
       74│ 
       75│         # Get the best link
       76│         chosen = max(links, key=lambda link: self._sort_key(package, link))
       77│ 

Cannot install tensorflow.
```

**Env:**
MAC OS - Ventura 13.1
Python 3.11

Did anyone faced this issue before?

javiber:
Version 2.12 fails on arm-based macOS.

Changed `tool.poetry.dependencies` so that it still accepts 2.12 as well as any newer version (before 3.0). This makes our package more forgiving when our users try to install it with different versions of tensorflow. 

Notice that accepting 2.12 in these dependencies is not a problem because pip will solve correctly for a version that's compatible with the user's environment.

On `tool.poetry.group.dev.dependencies` I had to be more strict because when poetry locks the dependencies, it'd choose 2.12 (even on an arm environment where the package will fail)

Closes #387


pyproject.toml
```diff
@@ -41,7 +41,7 @@ matplotlib = "^3.7.1"
 
 # Extras (keep versions in sync with dev deps)
 apache-beam = { version = "^2.48.0", optional = true }
-tensorflow = { version = "^2.12.0", optional = true }
+tensorflow = {version = ">=2.12.0,<3.0.0", optional = true}
```
achoum, line 44: tensorflow 2.16 drops support for Keras 2 which might cause some issues.
It might be interesting to keep <2.16 for now.

PR #413, opened 2024-04-03
Added bokeh_wegbl backend

javiber:
While testing locally the interactive plot using bokeh stops rendering when the eventset reaches 3 million timestamps (single feature, single index) that's around 45.8MB in memory.

The new plotter configures bokeh to use WebGL and allows me to render plots up to 30 million timestamps (457.8MB in memory).

I experimented with [bokeh server](https://docs.bokeh.org/en/latest/docs/user_guide/server.html) but it didn't perform better.


achoum (approved):
Nice

PR #419, opened 2024-04-09
Renamed from_indexed_dicts to from_struct

javiber:


temporian/__init__.py
```diff
@@ -37,72 +37,47 @@
...
-from temporian.core.data.dtype import int32
-from temporian.core.data.dtype import int64
-from temporian.core.data.dtype import bool_
-from temporian.core.data.dtype import str_
-from temporian.core.data.dtype import bytes_
-
-# Schema
-from temporian.core.data.schema import Schema, FeatureSchema, IndexSchema
-
 # Durations
```
achoum, line 40: Any reason for the change in formating?

javiber: That would be isort, I had it disabled but it came back when I recreated my local setup. Let me amend this

achoum (approved):
Awesome.

PR #426, opened 2024-05-06
Moving quantile

javiber:


docs/src/reference/temporian/operators/window/moving_quantile.md
```diff
@@ -0,0 +1 @@
+::: temporian.EventSet.moving_quantile
```
achoum, line 1: Awesome.
Can you also edit: docs/src/reference/index.md

achoum, line 1: Also, can you create an entry in benchmark/benchmark_time.py to facilitate the benchmarking.

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
+#include <iostream>
+#include <list>
+template <typename T>
+class CustomHeap {
+ private:
+  std::function<bool(T, T)> compare;
```
achoum, line 6: std::functions can be surprisingly slow. Can the compactor be a template argument?

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+      val_to_node.erase(it);
+    } else {
+      // TODO: exception meant for debugging, remove it
+      throw std::invalid_argument("removing a value that doesn't exists");
+    }
+  }
+  int size() { return heap.size(); }
+  int empty() { return heap.empty(); }
+
+  void print() {
```
achoum, line 64: Add comment "DO NOT SUBMIT"

temporian/implementation/numpy_cc/operators/window.cc
```diff
@@ -21,17 +23,41 @@ using ArrayRef = py::detail::unchecked_reference<T, 1>;
...
 // TODO: refactor to avoid code duplication where possible.
 
 // No external sampling, constant window length
-template <typename INPUT, typename OUTPUT, typename TAccumulator>
+template <typename INPUT, typename OUTPUT>
 py::array_t<OUTPUT> accumulate(const ArrayD &evset_timestamps,
                                const py::array_t<INPUT> &evset_values,
-                               const double window_length) {
+                               const double window_length,
+                               Accumulator<INPUT, OUTPUT> &accumulator) {
```
achoum, line 60: Can we keep this a template arg?

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
+#include <iostream>
+#include <list>
+template <typename T>
+class CustomHeap {
+ private:
+  std::function<bool(T, T)> compare;
+  std::list<T> heap;
```
achoum, line 7: Please comment on the exact semantic of "val_to_node", notably (but not only) regarding the validity of the iterators.

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+  std::unordered_map<T, typename std::list<T>::iterator> val_to_node;
+
+ public:
+  CustomHeap(std::function<bool(T, T)> compare) : compare(compare) {}
+
+  void push(T value) {
+    heap.push_back(value);
```
achoum, line 14: This is invalidating all the iterators in "val_to_node" right? Can you add a comment.

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+    auto it = std::prev(heap.end());
+    val_to_node[value] = it;
+    // TODO: better sorting?
+    while (it != heap.begin()) {
```
achoum, line 18: This looks expensive, can you add an explanation.

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  std::optional<T> pop() {
+    if (heap.size() == 0) {
+      return {};
+    } else {
+      auto value = heap.back();
+      heap.pop_back();
+      auto it = val_to_node.find(value);
```
achoum, line 37: What if there are multiple time the same value?

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  void remove(T value) {
+    auto it = val_to_node.find(value);
+    if (it != val_to_node.end()) {
+      heap.erase(it->second);
+      val_to_node.erase(it);
+    } else {
+      // TODO: exception meant for debugging, remove it
+      throw std::invalid_argument("removing a value that doesn't exists");
```
achoum, line 58: Please no exceptions

temporian/implementation/numpy_cc/operators/custom_heap.h
```diff
@@ -0,0 +1,76 @@
...
+      return {};
+    } else {
+      return heap.back();
+    }
+  }
+
+  void remove(T value) {
+    auto it = val_to_node.find(value);
+    if (it != val_to_node.end()) {
+      heap.erase(it->second);
```
achoum, line 54: At this point, the heap is not valid anymore, right?
Can you add a comment.

temporian/core/event_set_ops.py
```diff
@@ -3218,6 +3218,75 @@ def moving_min(
...
+        input's.
+
+        Missing values (such as NaNs) are ignored.
+
+        If the window does not contain any values (e.g., all the values are
+        missing, or the window does not contain any sampling), outputs missing
+        values.
+
+        The quantile calculated in each window is equivalent to numpy's
+        `"averaged_inverted_cdf"` method.
```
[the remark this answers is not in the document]
javiber: currently, I also support int as a valid input, however the output is converted to float

`float -> float`
`double -> double`
`int -> float`

temporian/core/operators/window/moving_quantile.py
```diff
@@ -0,0 +1,91 @@
...
+        quantile: float,
+        sampling: Optional[EventSetNode] = None,
+    ):
+        if quantile < 0 or quantile > 1:
+            raise ValueError(
+                "`quantile` must be a float between 0 and 1. "
+                f"Received {quantile}"
+            )
+        self.quantile = quantile
+        super().__init__(input, window_length, sampling)
```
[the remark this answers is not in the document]
javiber: there was a reason but I had so many backs and forth with this one that I forgot why. Let me try to put at the top as well as change quantile to _quantile and see if anything breaks

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: It shouldn't because heap is a double-linked list so all the existing nodes should remain in the same place. That's the reason why I'm using lists instead of vectors

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: It is, and I suspect this is where we spend the most time.
I need the heap to have the biggest (according to `compare`) item at the top and this was the easiest way to ensure that, I'm open to suggestions for a better sorting algorithm or a better data structure for the heap (trees maybe?)

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: That would be a problem I forgot to add a comment addressing that. It's not a problem in this case because I'm storing the indices as the heap "values" so they are never going to repeat

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: Similar to the comment above, it shouldn't because this is a double-linked list so this node is gone, then the next of the previous node and the previous of the next node are updated but all the nodes should remain in the same place.

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: good catch forgot to remove this one, it was only there for debugging

temporian/implementation/numpy_cc/operators/custom_heap.h
javiber: this was only for debugging so my intention was to remove it, do you think we should keep it?

temporian/core/operators/window/moving_quantile.py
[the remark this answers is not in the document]
javiber: It doesn't work, the reason is that the init of the base class runs a `self.check()` at the end and this needs all the attributes to be defined including the quantile

docs/src/reference/temporian/operators/window/moving_quantile.md
javiber: *added to the list of windows operations.

javiber: *added it

temporian/core/event_set_ops.py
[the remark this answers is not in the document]
javiber: *Added a comment on the doc explaining the point above

PR #440, opened 2026-01-11
Fix silent data loss in add_index for unsupported dtypes

### LINKED ISSUE #437: EventSet.add_index clears events
In the bank fraud detection tutorial the add_index method clears the events as shown by the following code and the plot in the tutorial is also empty

```python
dataset_tp.add_index('CUSTOMER_ID')
```

```
features [3]: TERMINAL_ID (str_), TX_AMOUNT (float64), TX_FRAUD (int64)
indexes [1]: CUSTOMER_ID (str_)
events: 0
index values: 4959
memory usage: 2.6 MB
```

https://github.com/google/temporian/blob/1e33b75b9fadfaf1c30bc20725cba9753105e4f9/docs/src/tutorials/bank_fraud_detection_with_tfdf.ipynb#L704-L709

Tanmay-24:
## Summary
  Fixes #437

  The `add_index()` operation was silently returning 0 events when the index column had an unsupported numpy dtype (e.g., unicode strings with dtype kind `'U'`).

  ## Root Cause
  In `add_index.cc` lines 221-228, when a numpy array's dtype kind was not `'S'` (bytes), the function would return without processing the feature OR raising an error, leaving the group accumulator empty.

  ## Fix
  Moved the `return` statement inside the `if (dtype().kind() == 'S')` block so that unsupported array dtypes fall through to the error handler that throws `std::invalid_argument("Non supported feature type")`.

### DIFF
--- a/temporian/core/operators/test/test_add_index.py
+++ b/temporian/core/operators/test/test_add_index.py
@@ -198,6 +198,20 @@ class AddIndexTest(TestCase):
         ):
             self.evset.set_index([])
 
+    def test_float_index_raises_error(self):
+        """Test that add_index on float column raises error instead of silent data loss.
+
+        Regression test for https://github.com/google/temporian/issues/437
+        """
+        evset = event_set(
+            [1, 2, 3],
+            features={"a": [1.0, 2.0, 3.0], "b": [1, 2, 3]},
+        )
+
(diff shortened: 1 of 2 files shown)


Tanmay-24:
@rstz can you please check? Thanks.
cc @achoum @javiber

PR #442, opened 2026-07-29
fix(ops): extend >, <, >=, <= to accept str and bytes scalars

### LINKED ISSUE #369: Extend EventSet's >, <, >=, <= to work with string literals
- `EventSet < EventSet` works with two numerical EventSets
- `EventSet < int` works with a numerical EventSet
- `EventSet < EventSet` works with two string EventSets
- `EventSet < str` doesn't work with a string EventSet, and it should:
<details>

```python
In [1]: import temporian as tp

In [2]: evset = tp.event_set([1], features={'a': ['a']})

In [3]: evsett = tp.event_set([1], features={'a': ['b']}, same_sampling_as=evset)

In [4]: evset > 'a'
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 evset > 'a'

File ~/temporian/temporian/core/event_set_ops.py:1081, in EventSetOperations.__gt__(self, other)
   1077     from temporian.core.operators.scalar import greater_scalar
   1079     return greater_scalar(input=self, value=other)
-> 1081 self._raise_error("compare", other, "(int,float)")
   1082 assert False

File ~/temporian/temporian/core/event_set_ops.py:88, in EventSetOperations._raise_error(self, op_name, other, allowed_types)
     80 def _raise_error(
     81     self, op_name: str, other: Any, allowed_types: str
     82 ) -> None:
     83     """Raises an error message.
     84
     85     This utility method is used in operator implementations, e.g., +, - *.
     86     """
---> 88     raise ValueError(
     89         f"Cannot use operator '{op_name}' on {self._clsname} and"
     90         f" {type(other)} objects. Only {self._clsname} or values of type"
     91         f" ({allowed_types}) are supported."
     92     )

ValueError: Cannot use operator 'compare' on EventSet and <class 'str'> objects. Only EventSet or values of type ((int,float)) are supported.
```

</details>

Same goes for all other comparison ops.

The `T_SCALAR` tuple defined in `temporian/core/event_set_ops.py` only accounts for `(int, float)`, but in some ops it should include `str` too. A second tuple needs to be defined for it and used where appropiate.

Note that `equal` and `__ne__` in the same file use `T_SCALAR + (bool, str)`, which could be defined as a third option.

Thorough tests needed to ensure we are accepting and rejecting the correct type on each op.

AloysJehwin:
Fixes #369

The four ordering operators (`__gt__`, `__ge__`, `__lt__`, `__le__`) in `event_set_ops.py` were checking `isinstance(other, T_SCALAR)` where `T_SCALAR = (int, float)`. So `evset < 'b'` raised a `ValueError` even though the underlying `RelationalScalarOperator` already supports `DType.STRING`.

Extended the check to `T_SCALAR + (str, bytes)` on all four operators and updated the error message. Four regression tests added, all passing.

### DIFF
--- a/temporian/core/event_set_ops.py
+++ b/temporian/core/event_set_ops.py
@@ -1073,12 +1073,12 @@ class EventSetOperations:
 
             return greater(input_left=self, input_right=other)
 
-        if isinstance(other, T_SCALAR):
+        if isinstance(other, T_SCALAR + (str, bytes)):
             from temporian.core.operators.scalar import greater_scalar
 
             return greater_scalar(input=self, value=other)
 
-        self._raise_error("compare", other, "(int,float)")
+        self._raise_error("compare", other, "(int,float,str,bytes)")
         assert False
 
     def __ge__(self: EventSetOrNode, other: Any) -> EventSetOrNode:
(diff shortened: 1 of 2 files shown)


AloysJehwin:
I have signed the CLA. Please re-check.

AloysJehwin:
This is ready for review. Happy to make any changes the maintainers suggest.

